PR #1220, opened 2022-10-28
[DM | Schema node card] Use filled chevron icon onHover

NicL9923:
Default:
image

onHover the chevron:
image


rllyy97 (approved):

rickixie:
ah - can we make sure the color stays the same? not changing

PR #1233, opened 2022-10-31
Hide loading state for branching subgraph nodes

### LINKED ISSUE #1227: Switch case detail panel loading forever, never shows params

1. Add a Switch node
2. Click the (+) button to add a case
3. Click the newly created Case
4. The panel should open and show a loading indicator

### Expected behavior

User should click the case node and see a panel where they can change the case settings

### Screenshots or Videos

image

### Platform And Browser

- Windows
- Chrome

AB#16065027

rllyy97:
## Main Changes
- Hiding loading state for branching subgraph nodes
- Switch case parameter is now shown correctly
- Fixes #1227

image

### DIFF
--- a/libs/designer/src/lib/ui/panel/panelroot.tsx
+++ b/libs/designer/src/lib/ui/panel/panelroot.tsx
@@ -211,7 +211,7 @@ export const PanelRoot = (): JSX.Element => {
       panelLocation={PanelLocation.Right}
       isCollapsed={collapsed}
       noNodeSelected={!selectedNode}
-      isLoading={!iconUri}
+      // isLoading={!iconUri}
       panelScope={PanelScope.CardLevel}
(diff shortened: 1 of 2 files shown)


libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -211,7 +211,7 @@ export const PanelRoot = (): JSX.Element => {
       panelLocation={PanelLocation.Right}
       isCollapsed={collapsed}
       noNodeSelected={!selectedNode}
-      isLoading={!iconUri}
+      // isLoading={!iconUri}
```
preetriti1, line 214: What is this removing?

preetriti1 (approved):

libs/designer/src/lib/ui/panel/panelroot.tsx
rllyy97: It was for the loading icon in the operation panel header, I just pushed another commit to also check if it's a branching node so this is also fixed now

PR #1238, opened 2022-11-01
Array Editor Label

Eric-B-Wu:
This is a more temporary fix, and will be revisited in my next PR as we refine Item Schema for complex array items

AB#16102587


hartra344 (approved):

Eric-B-Wu:
label gif

PR #1243, opened 2022-11-02
[DM | ReactFlow] Smart edges!

NicL9923:
Thanks to this lib, and a slight modification of the end path result before it's drawn, ReactFlow edges will now no longer overlap nodes (which was a symptom of doing layouting using elkjs primarily for Function nodes)

image


hartra344 (approved):

libs/data-mapper/src/lib/components/edge/ConnectionEdge.tsx
```diff
@@ -50,19 +51,78 @@ export const ConnectionEdge = (props: EdgeProps) => {
...
+    [sourcePosition, targetPosition, sourceX, sourceY, targetX, targetY]
+  );
+
+  const smartEdge = useMemo(
+    () =>
+      getSmartEdge({
+        ...baseEdgeProperties,
+        nodes: reactFlowNodes,
+        options: {
+          // nodePadding
```
refortie, line 81: Are these comments still important?

refortie (approved):

libs/data-mapper/src/lib/components/edge/ConnectionEdge.tsx
NicL9923: Yeah, intentionally left them there for quick/easy awareness of their existence as they may be critical and/or volatile settings in the future if any advanced tweaking is done to layouting

PR #1278, opened 2022-11-08
[Data Mapper] Style minimap, and fix one or two other small style things

NicL9923:
image

image

AB#16192064


refortie:
@rickixie Just something to note, I personally find the light theme, white background and light grey node pretty hard to see. Especially at the small size

refortie (approved):

rickixie:
> @rickixie Just something to note, I personally find the light theme, white background and light grey node pretty hard to see. Especially at the small size

Should be okay because it is less a priority on the page, should be sufficient to represent the outline of the map. But definitely something we will capture when getting accessibility review

PR #1284, opened 2022-11-09
[DM | ReactFlow] Better fitView + remove edge updating

NicL9923:
-Remove edge updating behavior (deprecated/removed from requirements)

-Use a hidden dummy placeholder node for ReactFlow so the initial fitView centers the entire diagram, not just the target schema (we've already been doing this with elk.js layouting with no issues, but many benefits!)

HiddenDummyNodeFitView


DanielleCogs (approved):

rickixie:
We are still waiting on adding the badge to the editing view right? I am hoping that will help the view change more consistent (i.e when changing view or navigating between levels, the margin on the top and and left is consistent so that user doesn't feel like there are a lot of moving pieces) @NicL9923

PR #1302, opened 2022-11-14
Added toast when attempting to delete target node

JoaquimMalcampo:
image


libs/data-mapper/src/lib/components/notification/Notification.tsx
```diff
@@ -106,6 +108,10 @@ export const Notification = (props: NotificationProps) => {
           nodeName: msgParam ?? '',
         }
       ),
+      [NotificationTypes.TargetNodeCannotDelete]: intl.formatMessage({
+        defaultMessage: `Target schema element cannot be deleted`,
```
NicL9923, line 112: Add a period to the end of the message so I don't get yelled at later please :P

libs/data-mapper/src/lib/core/state/DataMapSlice.ts
```diff
@@ -605,7 +605,13 @@ export const deleteConnectionFromConnections = (connections: ConnectionDictionar
 };
 
 export const deleteNodeWithKey = (curDataMapState: DataMapState, reactFlowKey: string) => {
-  if (curDataMapState.curDataMapOperation.flattenedTargetSchema[reactFlowKey]) {
+  const targetNode = curDataMapState.curDataMapOperation.flattenedTargetSchema[reactFlowKey];
+  if (targetNode) {
+    curDataMapState.notificationData = {
+      type: NotificationTypes.TargetNodeCannotDelete,
+      msgParam: targetNode.name,
```
NicL9923, line 612: You should be able to safely leave msgParam out as it doesn't seem to get used

NicL9923 (approved):

PR #1303, opened 2022-11-14
[DM | Telemetry] React app telemetry w/ AppInsights

NicL9923:
Implemented the wrapper for AppInsights telemetry/logging, and some exception tracking around various critical areas we already try/catch

-Also tested a trackEvent call to confirmed things were properly hooked up, and my AppInsights did indeed receive the events - so all good there

-To follow is discussion about specific events we want to track, which will be a very quick implementation

+one or two unrelated fixes (see commits)


libs/data-mapper/src/lib/core/services/appInsights/AppInsights.ts
```diff
@@ -0,0 +1,17 @@
+import { ReactPlugin } from '@microsoft/applicationinsights-react-js';
+import { ApplicationInsights } from '@microsoft/applicationinsights-web';
+
+export const reactPlugin = new ReactPlugin();
+
+const appInsights = new ApplicationInsights({
+  config: {
+    connectionString:
```
hartra344, line 8: Can we insert this during build time and just disable tracking if it's missing. Generally open source projects built from source should not track by default.

hartra344: Lets also make sure there is a vscode setting or something that can turn it off.

hartra344 (approved):

libs/data-mapper/src/lib/core/services/appInsights/AppInsights.ts
ccastrotrejo: There is this [disabling option](https://code.visualstudio.com/docs/getstarted/telemetry#_disable-telemetry-reporting) in the vscode configuration. Deep down these libraries take a look at it and decide whether log it or not. @NicL9923 just to double check try it out.

libs/data-mapper/src/lib/utils/ReactFlow.Util.ts
```diff
@@ -81,6 +82,7 @@ export const useLayout = (
         })
         .catch((error) => {
           console.error(`Elk Layout Error: ${error}`);
```
ccastrotrejo, line 84: NIT: console.log over here

NicL9923: Yeah that one's intentional - any console.errors should be intentional and formatted to help when dev-ing

libs/data-mapper/src/lib/core/services/appInsights/AppInsights.ts
NicL9923: @ccastrotrejo So then that setting should be enough huh? If the package we use for that in the VSIX automatically utilizes it under the hood?

ccastrotrejo: Yep, that should be fine. Thats the way turn it on/off.

NicL9923: @hartra344 @ccastrotrejo It looks like the React AI package plays nice (no thrown exceptions or console errors) if you just don't make the loadAppInsights call but still make the trackX calls as normal - so that should be good - gonna follow up about plugging in the keys in Teams

hartra344 (approved):

PR #1322, opened 2022-11-17
[DM | Tree] Maintain all trees' expanded states + auto-expand target schema tree as its traversed on canvas

NicL9923:
+Unrelated: Allow Code View to show when in Overview (BUG #16296426)

TreeStateUpdate


rickixie:
Perfect - just what i was imagining for the source / function side to remember the previous expanded state.

PR #1335, opened 2022-11-18
[Data Mapper] Dark theme code view

refortie:
Before
image

After
image


NicL9923 (approved):

rickixie:
woo where did you get the dark theme color (for the syntax)? @refortie

refortie:
> woo where did you get the dark theme color (for the syntax)? @refortie

We were coding in the theme in 2 spots, one spot was doing both light and dark, and the other spot was hard coded to the azure light theme. So I just snagged the equivalent built in dark one

PR #1339, opened 2022-11-18
[DM | Tree] Update how schema file names are shown/handled in schema trees

NicL9923:
-Separate schema file name out of the actual Tree
-schemaTreeRoot (the true root node, now displayed as top of tree) is now expanded by default, and can be collapsed again (was previously forced expanded)

Target schema tree:
image

Source schema tree:
image


refortie (approved):

rickixie (commented):
Is it possible remove the chevron spacing for the the file name one so the file and the root node don't feel like the same level?
CleanShot 2022-11-18 at 13 47 34@2x


Also mentioned this in the call too to update the chevron size from 20 to 12. (just update the icon from Chevron20Regular to Chevron12Regular, Chevron20Filled to Chevron12Filled

NicL9923:
> Is it possible remove the chevron spacing for the the file name one so the file and the root node don't feel like the same level? CleanShot 2022-11-18 at 13 47 34@2x
> 
> Also mentioned this in the call too to update the chevron size from 20 to 12. (just update the icon from Chevron20Regular to Chevron12Regular, Chevron20Filled to Chevron12Filled

The chevrons got updated in a previous PR, these photos just didn't have those changes

Following up on Teams on the spacing

PR #1340, opened 2022-11-21
fixed 2 connections on instert bug

DanielleCogs:
1. Fixed issue where initial connection is left if function is inserted in repeated connection: https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16309159
2. Fixed logic that causes "parent connection added" message when not needed https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16322647

### DIFF
--- a/libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
+++ b/libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
@@ -188,5 +188,587 @@ describe('utils/DataMap', () => {
       expect(input1.reactFlowKey).toEqual('source-/ns0:Root/ManyToOne/SourceYear/SourceMonth');
       expect(input2.reactFlowKey).toEqual('source-/ns0:Root/ManyToOne/SourceYear');
     });
+
+    it('adds parent connections for indexed many-to-one', () => {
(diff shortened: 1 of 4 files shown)


libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
```diff
@@ -188,5 +188,587 @@ describe('utils/DataMap', () => {
       expect(input1.reactFlowKey).toEqual('source-/ns0:Root/ManyToOne/SourceYear/SourceMonth');
       expect(input2.reactFlowKey).toEqual('source-/ns0:Root/ManyToOne/SourceYear');
     });
+
+    it('adds parent connections for indexed many-to-one', () => {
+      const sourceNodeParent = flattenedLoopSource['source-/ns0:Root/ManyToOne/SourceYear'];
+      const targetNodeParent = flattenedLoopTarget['target-/ns0:Root/ManyToOne/Date'];
+      // const connectionsDict = indexed;
```
refortie, line 195: Do we still need these comments?

refortie (approved):

libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
DanielleCogs: ah I can remove them

PR #1341, opened 2022-11-22
Loop delete logic

DanielleCogs:
Deletion of repeating connection now deletes appropriate parent
- Should not delete parent connection if other child is repeating
- Deletes parent even if connection is not direct from source to target
- does not allow deletion of "N" connections, shows warning message


libs/data-mapper/src/lib/core/state/DataMapSlice.ts
```diff
@@ -677,18 +724,41 @@ export const deleteNodeWithKey = (curDataMapState: DataMapState, reactFlowKey: s
...
+  const connections = { ...curDataMapState.curDataMapOperation.dataMapConnections };
+
+  if (canDelete) {
+    deleteConnectionFromConnections(
+      connections,
+      getSourceIdFromReactFlowConnectionId(reactFlowKey),
+      getDestinationIdFromReactFlowConnectionId(reactFlowKey)
+    );
+    const tempConn = connections[getSourceIdFromReactFlowConnectionId(reactFlowKey)];
+    console.log(JSON.stringify(tempConn));
```
NicL9923, line 740: Raising this and Ln 742 just-in-case

libs/data-mapper/src/lib/core/state/DataMapSlice.ts
```diff
@@ -619,6 +619,53 @@ export const deleteConnectionFromConnections = (connections: ConnectionDictionar
   );
 };
 
+export const deleteParentRepeatingConnections = (connections: ConnectionDictionary, inputKey: string /* contains prefix */) => {
+  const parentId = getParentId(inputKey);
+  if (parentId.endsWith('source-')) {
```
NicL9923, line 624: I wanna say there's a const for 'source-' somewhere in either ReactFlow Utils or Constants

PR #1348, opened 2022-11-23
Tree view in vscode extension

ccastrotrejo:


derek1ee:
All other extensions (e.g. Functions) have moved from have a single top-level entry into "resources"/"workspaces", I think we should be consistent and do the same?

ccastrotrejo:
> All other extensions (e.g. Functions) have moved from have a single top-level entry into "resources"/"workspaces", I think we should be consistent and do the same?

I looked around at how this works and resources extension needs to implement the Logic Apps configuration over there. I talked with Nathan from the resource's extension team, and he told me that they are currently overhauling the API to integrate into the Resources view. That work is expected to be done early next year

NicL9923 (approved):

PR #1363, opened 2022-11-29
[Data Mapper] Shows indices in the prop pane

refortie:
image

image


NicL9923 (approved):

rickixie:
Should the data type for indices an integer? @refortie

refortie:
> Should the data type for indices an integer? @refortie

It should be Any, it's a symptom of needing it to be both an integer in some situations and the loop object in others.

rickixie:
> > Should the data type for indices an integer? @refortie
> 
> It should be Any, it's a symptom of needing it to be both an integer in some situations and the loop object in others.

I see, probably should note for the research that since it is doing double duty, might not be easy to understand from the user perspective. cc @derek1ee

PR #1366, opened 2022-11-30
[DM] adding popover for repeating elements

DanielleCogs:
looks like this: 

https://user-images.githubusercontent.com/99753874/204801705-345888e7-21dd-4c2e-8b0a-e93e9a610999.mov


NicL9923 (approved):

rickixie (commented):
Can we make sure using the **Tooltip** V9 component as PopOver is mainly use for more rich content. They are also more accessible properties such as the prop relationship: label which will provide aria-label

https://react.fluentui.dev/?path=/docs/components-tooltip--default

PR #1370, opened 2022-11-30
[DM | PropPane] Fix schema node input value check to account for empty custom (string) values

NicL9923:
-Check explicit undefined so that empty custom values are properly assigned to InputDropdowns


libs/data-mapper/src/lib/components/propertiesPane/tabs/SchemaNodePropertiesTab.tsx
```diff
@@ -108,7 +108,7 @@ export const SchemaNodePropertiesTab = ({ currentNode }: SchemaNodePropertiesTab
 
     if (connection?.inputs && connection.inputs[0].length === 1) {
       const input = connection.inputs[0][0];
-      newInputValue = !input ? undefined : isCustomValue(input) ? input : input.reactFlowKey;
+      newInputValue = input === undefined ? undefined : isCustomValue(input) ? input : input.reactFlowKey;
```
refortie, line 111: Do we want to allow empty strings?

refortie (approved):

PR #1374, opened 2022-12-01
[Data Mapper] Direct Index Access

refortie:
Does not include deserialization work yet. That will come in the next PR

image

image

image

### DIFF
--- a/libs/data-mapper/src/lib/components/inputDropdown/InputDropdown.tsx
+++ b/libs/data-mapper/src/lib/components/inputDropdown/InputDropdown.tsx
@@ -4,9 +4,16 @@ import type { SchemaNodeDataType, SchemaNodeExtended, SchemaNodeProperty } from 
 import { NormalizedDataType } from '../../models';
 import type { ConnectionUnit, InputConnection } from '../../models/Connection';
 import type { FunctionData } from '../../models/Function';
-import { indexPseudoFunctionKey } from '../../models/Function';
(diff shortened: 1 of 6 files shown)


libs/data-mapper/src/lib/components/inputDropdown/InputDropdown.tsx
```diff
@@ -289,10 +296,15 @@ export const InputDropdown = (props: InputDropdownProps) => {
 
       const inputs = connectionDictionary[key].inputs[0];
       const sourceNode = inputs && inputs[0];
-      const nodeName =
-        node.key === indexPseudoFunctionKey && isConnectionUnit(sourceNode) && isSchemaNodeExtended(sourceNode.node)
-          ? calculateIndexValue(sourceNode.node)
-          : getFunctionOutputValue(fnInputValues, node.functionName);
+      let nodeName: string;
+      if (node.key === indexPseudoFunctionKey && isConnectionUnit(sourceNode) && isSchemaNodeExtended(sourceNode.node)) {
```
NicL9923, line 300: *Line 302* Does this need the isConnectionUnit/isSchemaNodeExtended check too?

NicL9923 (approved):

libs/data-mapper/src/lib/components/inputDropdown/InputDropdown.tsx
refortie: Nope. Those checks are type converting the sourceNode and we don't need to do that in 302

PR #1395, opened 2022-12-05
labugfix: Connection type and parameter bugs

### LINKED ISSUE #1364: Service Provider connections not being correctly created in new designer
### Describe the Bug

Creating a new connection for a service provider connecter in the new designer does not work correctly. The connection gets created without the "type", this makes the connection invalid and logic app won't run. 

image Shows the difference between the two designers when creating a connection.

Also, the secret for AADOAuth should not be saved as plain text in connections but rather an appsetting, this seems like it's a bug with existing designer.

1. Go to new designer
2. Place a SQL action
3. Create multiple connections (via change connection, add) for the SQL provider with all available types.

### Workflow JSON

_No response_

### Expected behavior

Since type is not included in connection json, the workflow won't start. The connection json should include the auth type.

### Screenshots or Videos

_No response_

### Platform And Browser

N/A

### Additional context

Contact Abode if you need more context here 

AB#16438914

rllyy97:
## Main Changes
- Fixes #1364
  - Selected auth type now passed with parameters
- Fixed a regression bug where connection parameters were hidden


libs/utils/src/lib/helpers/hooks.ts
```diff
@@ -37,3 +37,5 @@ export const useWindowDimensions = () => {
 
   return windowDimensions;
 };
+
+export const useConsoleLog = (value: any) => useEffect(() => console.log(value), [value]);
```
rllyy97, line 41: I just use this during development, not intended to ever commit a usage but helpful to have available.

hartra344 (approved):

PR #1403, opened 2022-12-07
[DM] small refactor and re-added simple test

DanielleCogs:


libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
```diff
@@ -163,6 +164,18 @@ describe('utils/DataMap', () => {
       expect(resultEntries[3][1]).toBeTruthy();
     });
 
+    it('creates a simple conditional connection', () => {
+      simpleMap['ns0:Root'] = {
+        DataTranslation: {
+          EmployeeName: {
+            RegularFullTime: '$if(is-greater-than(/ns0:Root/ConditionalMapping/ItemPrice, /ns0:Root/ConditionalMapping/ItemQuantity))',
```
refortie, line 171: I'm not sure this is a valid mapping? I could be wrong, but I thought we needed to use "if-then-else" in something like `if-then-else(is-greater-than(/ns0:Root/ConditionalMapping/ItemPrice, /ns0:Root/ConditionalMapping/ItemQuantity), $true, $nil`)

libs/data-mapper/src/lib/utils/DataMap.Utils.ts
```diff
@@ -495,6 +492,28 @@ const parseDefinitionToConnection = (
   }
 };
 
+type UnknownNode = SchemaNodeExtended | FunctionData | undefined;
+
+export const getDestinationNode = (targetKey: string, functions: FunctionData[], schemaTreeRoot: SchemaNodeExtended): UnknownNode => {
+  const dashIndex = targetKey.indexOf('-');
+  const destinationFunctionKey = dashIndex === -1 ? targetKey : targetKey.slice(0, dashIndex); // what is the purpose of this??
```
refortie, line 499: Still need that comment?

libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
```diff
@@ -163,6 +164,18 @@ describe('utils/DataMap', () => {
...
+          },
+        },
+      };
+      const result = convertFromMapDefinition(simpleMap, extendedSource, extendedTarget, [greaterThanFunction, conditionalFunction]);
```
refortie, line 175: Instead of importing the individual functions, you could just use "functionMock" it has all of the functions we use in standalone

refortie (approved):

libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
DanielleCogs: I can send it to Sanjiv's team to confirm, but that is what we generate. I know that we also need to check if-then-else, but I was thinking including this simple test could be helpful for me as a sanity check.

refortie: Let's chat offline about this

PR #1411, opened 2022-12-08
Clean up some of the worst npm issues

refortie:


package.json
```diff
@@ -106,6 +105,7 @@
     "react-use": "17.4.0",
     "reactflow": "^11.2.0",
     "regenerator-runtime": "0.13.9",
+    "request": "^2.88.2",
```
refortie, line 108: Had to add this in manually, since we were getting it from the `json-schema-generator` chain.

@hartra344 Probably should move off of request/request-promise since both are super deprecated

hartra344 (approved):

PR #1413, opened 2022-12-08
[DM | Canvas] Restrict panning to certain bounds

NicL9923:
Totally not shamelessly piggybacking off @rllyy97 (Thank you! :^) )

*Tradeoff (as rllyy97 probably discovered as well) is moving the canvas' zoom to controlled state, but as the comment I left suggests, it's that, or useViewport which re-renders not only each time zoom changes, but on any sort of panning action == lag

-It does confirmed-ly adapt to the diagram as it expands (see the 5-10 ToStrings in the GIF)
-It will need to a little more fine-tuned once the Code View gets some love again (as this currently just uses the canvas width as if the CV was never open - not a huge deal though)

CanvasPanningBounds

### DIFF
--- a/libs/data-mapper/src/lib/ui/ReactFlowWrapper.tsx
+++ b/libs/data-mapper/src/lib/ui/ReactFlowWrapper.tsx
@@ -6,8 +6,10 @@ import { SchemaCard } from '../components/nodeCard/SchemaCard';
 import { Notification } from '../components/notification/Notification';
 import { SchemaNameBadge } from '../components/schemaSelection/SchemaNameBadge';
 import { SourceSchemaPlaceholder } from '../components/schemaSelection/SourceSchemaPlaceholder';
+import { schemaNodeCardHeight, schemaNodeCardWidth } from '../constants/NodeConstants';
 import {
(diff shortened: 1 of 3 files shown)


ccastrotrejo (approved):

libs/data-mapper/src/lib/ui/ReactFlowWrapper.tsx
```diff
@@ -152,21 +156,38 @@ export const ReactFlowWrapper = ({ canvasBlockHeight }: ReactFlowWrapperProps) =
...
   return (
     <ReactFlow
       ref={reactFlowRef}
-      onKeyDown={keyDownHandler}
       nodeTypes={nodeTypes}
       edgeTypes={edgeTypes}
       nodes={nodes}
       edges={edges}
+      onPaneClick={onPaneClick}
+      // Not ideal, but it's this or useViewport that re-renders 3000 (due to x/y changes)
```
rllyy97, line 181: The state update that happens here only triggers the dependent useMemo if the value changes as far as I know, so it should only be recalculating the TE on zoom value changes!  We should be fine using this as is

PR #1417, opened 2022-12-09
[DM | Utils] Combine connection input setter methods + other minor refactors

NicL9923:


libs/data-mapper/src/lib/core/state/__test__/DataMapSlice.spec.ts
```diff
@@ -29,11 +24,19 @@ describe('DataMapSlice', () => {
   const destinationId = addReactFlowPrefix(node.key, SchemaType.Target);
   const functionId = createReactFlowFunctionKey(concatFunction);
 
-  describe('addNodeToConnections', () => {
+  describe('Simulate handle-drawn / deserialization-made connection', () => {
```
refortie, line 27: I'd name this "setConnectionInputValue" since that's the method it's testing

libs/data-mapper/src/lib/utils/Connection.Utils.ts
```diff
@@ -40,152 +47,138 @@ export const createConnectionEntryIfNeeded = (
...
+ * both those that are deserialized, handle-drawn, or InputDropdown-set.
+ *
+ * @param connections The connections dictionary to add the entry to
+ * @param targetNode The node to create the connection entry for
+ * @param targetNodeReactFlowKey The reactFlowKey of `targetNode`
+ * @param inputIndex The index of the input to set the value for (only needed if NOT `isHandleDrawnOrDeserialized`)
+ * @param value The value to set the input to (InputConnection, or `null` to delete an unbounded input)
+ * @param isHandleDrawnOrDeserialized Whether the new connection being made is handle-drawn or a deserialization
+ */
+export const setConnectionInputValue = (
```
refortie, line 62: I'll be honest, I'm not thrilled about this name. I get why you picked it and it's not inaccurate.

libs/data-mapper/src/lib/utils/Connection.Utils.ts
```diff
@@ -40,152 +47,138 @@ export const createConnectionEntryIfNeeded = (
...
+ * Creates or updates a connection entry's input value.
+ * This is the gateway for *setting ANY input values* in dataMapConnections - including
+ * both those that are deserialized, handle-drawn, or InputDropdown-set.
+ *
+ * @param connections The connections dictionary to add the entry to
+ * @param targetNode The node to create the connection entry for
+ * @param targetNodeReactFlowKey The reactFlowKey of `targetNode`
+ * @param inputIndex The index of the input to set the value for (only needed if NOT `isHandleDrawnOrDeserialized`)
+ * @param value The value to set the input to (InputConnection, or `null` to delete an unbounded input)
+ * @param isHandleDrawnOrDeserialized Whether the new connection being made is handle-drawn or a deserialization
```
refortie, line 60: Instead of the motion causing it, maybe this could be named after the action that results from it? Or if that doesn't make sense, what about the inverse so it's not named "is A Or B", but is named "is C" and we worry about the neg case

libs/data-mapper/src/lib/utils/__test__/ConnectionUtils.spec.ts
```diff
@@ -89,7 +88,7 @@ describe('utils/Connections', () => {
     });
   });
 
-  describe('addNodeToConnections (primarily drawn connections)', () => {
+  describe('setConnectionInputValue - Drawn connections', () => {
```
refortie, line 91: Nit: make this describe just "setConnectionInputValue" and then layer in another describe "Drawn Connections" and then below do the same for input drop down

libs/data-mapper/src/lib/utils/Connection.Utils.ts
```diff
@@ -10,8 +10,15 @@ import { addReactFlowPrefix, addTargetReactFlowPrefix } from './ReactFlow.Util';
...
+/**
+ * Creates a connection entry in the connections dictionary if it doesn't already exist.
+ * This is the gateway for any entries made into dataMapConnections - meaning all default
+ * inputs/etc can safely be managed in this singular spot.
+ *
+ * @param connections The connections dictionary to add the entry to
+ * @param node The node to create the connection entry for
+ * @param reactFlowKey The reactFlowKey of the node
+ */
 export const createConnectionEntryIfNeeded = (
```
refortie, line 22: I see in a lot of spots, you've deleted calls to this method, is it still necessary to be exported? It might be even if there are no calls to it. Thoughts?

refortie (changes requested):

PR #1435, opened 2022-12-19
[DM] Display when input values aren't correct types

NicL9923:
~~Depends on #1432~~

image

image


hartra344 (approved):

rickixie:
@NicL9923 

**Two feedbacks**
1. When a node is in error state, the border should be red (even it is selected) as it is more important
2. Same for the dropdown and text field

**Nice to have**
1. Add an error icon in front of the error message in the dropdown or text field ([See Fluent 9 validation](https://react.fluentui.dev/?path=/docs/preview-components-field--label#validation-state))
CleanShot 2023-01-04 at 10 33 56@2x

PR #1448, opened 2022-12-27
laFeat: OperationManifestService for Consumption Workflows

rllyy97:
## Main Changes
- Added OperationManifestService for Consumption
  - All manifests for implemented consumption built-in operations were added
  - They can now be added to the workflow and display correct inputs/outputs

AB#16681400


ccastrotrejo (approved):

Eric-B-Wu (approved):

libs/services/designer-client-services/src/lib/consumption/operationmanifest.ts
```diff
@@ -0,0 +1,48 @@
...
+      } = response;
+
+      // TODO: Remove below patching of connection when backend api sends correct information for service providers
+      const operationManifest = {
+        properties: {
+          brandColor,
+          description,
+          iconUri,
+          connection: equals(operationType, 'serviceprovider') ? { required: true, type: ConnectionType.ServiceProvider } : undefined,
+          ...manifest,
```
preetriti1, line 39: This whole code will not work in consumption, just get from supportedManifest if not then there is no service in consumption

libs/services/designer-client-services/src/lib/standard/manifests/compose.ts
```diff
@@ -0,0 +1,45 @@
+import { coreBadge } from '../../badges';
+import { OutputSecureDataMode, SettingScope } from '@microsoft/utils-logic-apps';
+import type { OperationManifest } from '@microsoft/utils-logic-apps';
+
+export const composeManifest = {
```
preetriti1, line 5: This file should be in src\lib\consumption/manifests since standard already has compose manifest and the operation manifest service should add it explicitly to the base value which is common between standard and consumption

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
```diff
@@ -0,0 +1,210 @@
...
+    displayName: 'XML Operations',
+  },
+};
+
+const settings: any = {
+  secureData: {},
+  trackedProperties: {
+    scopes: [SettingScope.Action],
+  },
+};
```
preetriti1, line 18: Did we verify from consumption portal this action only has these two settings?

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
```diff
@@ -0,0 +1,210 @@
...
+      type: 'object',
+      required: ['content', 'mapName', 'xsltParameters'],
+      properties: {
+        containerId: {
+          name: 'containerId',
+          title: 'Container Id',
+          summary: 'Function container',
+          description: 'The function container',
+          type: 'string',
+          visibility: 'internal',
```
preetriti1, line 37: There is no visibility internal anymore in new designer, we will have to use hideInUI attribute to just hide in UI but serialie anyways.

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
```diff
@@ -0,0 +1,210 @@
...
+          'x-ms-dynamic-values': {
```
preetriti1, line 38: x-ms-dynamic-values are not in manifest they are swagger based dynamic extension, we should be using x-ms-dynamic-list.

libs/services/designer-client-services/src/lib/standard/operationmanifest.ts
```diff
@@ -1,5 +1,6 @@
 import type { IHttpClient } from '../httpClient';
```
preetriti1, line 1: Strongly oppose changes in this file :) these are already sent by service in standard so we should even try to hardcode these manifests.

preetriti1 (changes requested):

preetriti1:
Maybe we can work on testing the newly added operations separately and the next PR can focus on fixing the manifests. Lets just get the new service added and proper structuring of the new operations only on consumption side

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
rllyy97: As far as I can tell these settings match the current consumption portal settings

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
rllyy97: Where should I get `x-ms-dynamic-list` property information from?  I don't see any mention of it in BPMUX.
Should I also just remove the `x-ms-dynamic-values`?  From your comment I thought the two props used the same information but they don't really match up.

libs/services/designer-client-services/src/lib/standard/operationmanifest.ts
rllyy97: I went ahead and reverted a lot of this file, and I separated the services into base/standard/consumption versions so everything should be separated and organized corrected now.  If you think I need to make more changes I will adjust as needed!

libs/services/designer-client-services/src/lib/standard/manifests/xml.ts
preetriti1: lets leave the manifest updates in this PR, we will work on it separately since we have to make sure this manifest works for the runtime and authoring.

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -294,7 +294,7 @@ function getVisibility(parameter: any) {
 
 function shouldHideInUI(parameter: ResolvedParameter): boolean {
   const visibility = getVisibility(parameter);
-  return parameter?.hideInUI || equals(visibility, 'hideInUI') || equals(visibility, Visibility.Internal);
+  return parameter?.hideInUI || parameter?.schema?.hideInUI || equals(visibility, 'hideInUI') || equals(visibility, Visibility.Internal);
```
preetriti1, line 297: When do we have hideInUI inside parameter schema?

libs/services/designer-client-services/src/lib/base/operationManifest.ts
```diff
@@ -0,0 +1,641 @@
...
+
+  isSupported(operationType: string, _operationKind?: string): boolean {
+    const { supportedTypes } = this.options;
+    const normalizedOperationType = operationType.toLowerCase();
+    return supportedTypes
+      ? supportedTypes.indexOf(normalizedOperationType) > -1
+      : supportedManifestTypes.indexOf(normalizedOperationType) > -1;
+  }
+
+  async getOperationInfo(definition: any, isTrigger: boolean): Promise<OperationInfo> {
```
preetriti1, line 194: This operation should be abstract and should be implemented by standard and consumption separately, reason being serviceprovider is not a valid operation in consumption.

libs/services/designer-client-services/src/lib/base/operationManifest.ts
```diff
@@ -0,0 +1,641 @@
...
+
+    return {
+      connectorId: 'Unknown',
+      operationId: 'Unknown',
+    };
+
+    //throw new UnsupportedException(`Operation type: ${definition.type} does not support manifest.`);
+  }
+
+  async getOperationManifest(_connectorId: string, _operationId: string): Promise<OperationManifest> {
```
preetriti1, line 212: Lets make it abstract

libs/services/designer-client-services/src/lib/consumption/operationmanifest.ts
```diff
@@ -0,0 +1,44 @@
...
+// import { equals, ConnectionType } from '@microsoft/utils-logic-apps';
+import { composeManifest } from './manifests/compose';
+import { flatFileDecodingManifest, flatFileEncodingManifest } from './manifests/flatfile';
+import { inlineCodeManifest } from './manifests/inlinecode';
+import { integrationAccountArtifactLookupManifest } from './manifests/integrationaccountartifactlookup';
+import { liquidJsonToJsonManifest, liquidJsonToTextManifest, liquidXmlToJsonManifest, liquidXmlToTextManifest } from './manifests/liquid';
+import { xmlTransformManifest, xmlValidationManifest } from './manifests/xml';
+import type { OperationManifest } from '@microsoft/utils-logic-apps';
+
+export class ConsumptionOperationManifestService extends StandardOperationManifestService {
```
preetriti1, line 12: Shouldn't this be extending from baseoperationmanifestservice?

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/helper.ts
rllyy97: At some point during our importing of the manifest we put the parameter object behind a `schema` prop, in the manifests I'm putting the `hideInUI` prop at the highest level and it still doesn't apply unless I add this schema step.

preetriti1: If this resolved parameter is similar to ParameterInfo then there should be method called createParameterInfo in utils/parameters/helper.ts that should be reading this prop corectly and putting it as first class property in parameterinfo.

preetriti1 (approved):

libs/designer/src/lib/core/utils/parameters/helper.ts
rllyy97: The `createParameterInfo` function gets the `hideInUI` prop from this `shouldHideInUI` function I'm referencing above ^^

PR #1455, opened 2022-12-30
[DM] Adding icons to prod CDN

DanielleCogs:
Icons now uploaded to PROD CDN. 
Note:
- used Fluent V8 because it has a fallback option, in case the icon cannot be found
- Utility is used if the icon does not load
image

### DIFF
--- a/libs/data-mapper/src/lib/utils/Icon.Utils.tsx
+++ b/libs/data-mapper/src/lib/utils/Icon.Utils.tsx
@@ -197,9 +197,11 @@ export const iconForSchemaNodeDataType = (
   return bundled ? bundleIcon(icons[1], icons[0]) : icons[0];
 };
 
+export const iconBaseUrl = 'https://logicappsv2resources.blob.core.windows.net/icons/datamapper/';
+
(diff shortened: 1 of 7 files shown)


hartra344 (approved):

libs/data-mapper/src/lib/utils/Icon.Utils.tsx
```diff
@@ -200,41 +197,48 @@ export const iconForSchemaNodeDataType = (
...
+  const functionIconUrlPrefix = `${iconBaseUrl}${FunctionCategoryIconPrefix}`;
+  if (functionCategories.indexOf(functionCategory) >= 0) {
+    return `${functionIconUrlPrefix}${functionCategory.toLowerCase().replace(' ', '')}.svg`;
+  } else {
+    return `${functionIconUrlPrefix}${FunctionCategory.Utility.toLowerCase()}.svg`;
   }
 };
 
 export const iconUriForIconImageName = (iconImageName: string) => {
   // TODO Temp CDN, will need to be moved into a production location
```
NicL9923, line 213: This comment should be fine to remove now huh?

PR #1465, opened 2023-01-03
laFeat: Optional Gateway Parameters

### LINKED ISSUE #1399: Adding a SharePoint transformation (Get File Properties just hangs)
### Describe the Bug

THe logic app designer just spins and does nothing when I add a Get File Properties SharePoint transformation to the workflow.

Just add a SharePoint Get FIle Properties transformation and add to your workflow.

### Workflow JSON

_No response_

### Expected behavior

The transformation should just show up.

### Screenshots or Videos

_No response_

### Platform And Browser

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]
 

### Additional context

_No response_

AB#16528029

rllyy97:
## Main Changes
- Implemented optional gateway parameters during connection creation
  - This does not happen on many connectors but it was needed for some of them to function properly
  - Reworked the connection page a little bit to separate these optional gateway params from the rest
  - Fixes #1399 

image
image

AB#16708127


preetriti1:
How would this look like when there are gateways with a non oauth connection type, will it show the connection parameters above or below the optional check for gateways... I think azure SQL has that

rllyy97:
> How would this look like when there are gateways with a non oauth connection type, will it show the connection parameters above or below the optional check for gateways... I think azure SQL has that

Currently it renders this checkbox above all else, but I can make it below other params / above gateway params.
For that Azure SQL connector, on old-designer we don't show the optional gateway checkbox on multi-auth connections, so it just shows all the gateway parameters.  Is that incorrect / should we be showing the checkbox instead?

ccastrotrejo (approved):

libs/designer-ui/src/lib/createConnection/UniversalConnectionParameter.tsx
```diff
@@ -0,0 +1,113 @@
+import GatewayPicker from './gatewayPicker';
```
tonytang-microsoft-com, line 1: Curious, shall the filename be in pascal casing like any other filenames?

libs/designer-ui/src/lib/createConnection/index.tsx
```diff
@@ -103,35 +103,71 @@ export const CreateConnection = (props: CreateConnectionProps): JSX.Element => {
...
+  );
+
   const isParamVisible = useCallback(
     (parameter: ParamType) => {
       const constraints = parameter?.uiDefinition?.constraints;
       if (constraints?.hidden === 'true' || constraints?.hideInUI === 'true') return false;
       const dependencyParam = constraints?.dependentParameter;
       if (dependencyParam && parameterValues[dependencyParam.parameter] !== dependencyParam.value) return false;
       if (parameter.type === ConnectionParameterTypes[ConnectionParameterTypes.oauthSetting]) return false;
+      if (!optionalGatewayEnabled && isOptionalGatewayParam(parameter)) return false;
```
tonytang-microsoft-com, line 139: 1. What is the scenario we are trying to achieve? I'm curious weather this feature is needed, because seems it add unnecessary complexity and doesn't align with how capabilities shall be used originally.
2. This code adds ad hoc logic that couples capabilities with visibility and type.

libs/designer-ui/src/lib/createConnection/index.tsx
```diff
@@ -324,94 +365,55 @@ export const CreateConnection = (props: CreateConnectionProps): JSX.Element => {
...
+              setValue={(val: any) => setParameterValues({ ...parameterValues, [key]: val })}
+              isLoading={isLoading}
+              selectedSubscriptionId={selectedSubscriptionId}
+              selectSubscriptionCallback={selectSubscriptionCallback}
+              availableGateways={availableGateways}
+              availableSubscriptions={availableSubscriptions}
+            />
+          ))}
+
+        {/* OptionalGateway Check */}
```
tonytang-microsoft-com, line 383: shall these code be moved into UniversalConnectionParameter?

libs/designer-ui/src/lib/createConnection/index.tsx
preetriti1: I think what tony means here is, instead of having it as gateway option or gateway parameter, it should be a capability option or capability parameter, because that is how it was modeled i think.. capabilities are array of string, it can have cloud, gateway.. the previous code kept it very hard coded logic as to what capability we are looking for

preetriti1: maybe you can have this as a function and write the specific logic there, this can check for a capability to enable and if the parameter only belongs to that capability?

preetriti1 (approved):
I am fine with the changes so signing off

PR #1482, opened 2023-01-09
[DM Feature] Indent nested loop source nodes

DanielleCogs:
Three main changes with this PR:
1. Parents of repeating items are added when child is added, not when connection is made (still kept code to add back parent if child is connected, in case it was removed)
2. In-between ancestors are added if the repeating element has a higher ancestor on the canvas
3. When the source canvas changes, all nodes sizes are recalculated


libs/data-mapper/src/lib/utils/Schema.Utils.ts
```diff
@@ -47,6 +55,35 @@ export const parsePropertiesIntoNodeProperties = (propertiesString: string): Sch
   return [];
 };
 
+export const setWidthForSourceNodes = (sourceNodes: SourceSchemaNodeExtended[]) => {
```
DanielleCogs, line 58: this is the more efficient version that needs to be refactored

libs/data-mapper/src/lib/utils/ReactFlow.Util.ts
```diff
@@ -178,6 +178,26 @@ const convertSourceToReactFlowParentAndChildNodes = (
     return reactFlowNodes;
   }
 
+  const sourceNodesCopy = [...combinedSourceSchemaNodes];
+  sourceNodesCopy.filter((node) => node.nodeProperties.includes(SchemaNodeProperty.Repeating));
+  const sourceKeySet: Set<string> = new Set();
+  sourceNodesCopy.forEach((node) => sourceKeySet.add(node.key));
+  const widthDict: Map<string, number> = new Map<string, number>();
+  let maxSize = 200;
```
refortie, line 186: Nit: Maybe add some of these numbers to constants and then reference them here?

libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
```diff
@@ -20,6 +23,35 @@ import {
 } from '../__mocks__';
 
 describe('utils/DataMap', () => {
+  describe('addAncestorNodesToCanvas', () => {
+    const sourceSchema: Schema = layeredLoopSourceMockSchema;
+    const extendedSourceSchema: SchemaExtended = convertSchemaToSchemaExtended(sourceSchema);
+    const flattenedSchema = flattenSchemaIntoDictionary(extendedSourceSchema, SchemaType.Source);
+    it('includes direct parent', () => {
```
refortie, line 30: Nit: some new lines here and below to break the `it()` from each other

refortie (approved):

libs/data-mapper/src/lib/models/Schema.ts
```diff
@@ -55,11 +55,13 @@ export interface SchemaExtended extends Schema {
   schemaTreeRoot: SchemaNodeExtended;
 }
 
+export type SourceSchemaNodeExtended = SchemaNodeExtended & { width?: number };
 export interface SchemaNodeExtended extends SchemaNode {
   children: SchemaNodeExtended[];
   nodeProperties: SchemaNodeProperty[];
   // Inclusive of the current node
   pathToRoot: PathItem[];
+  width?: number;
```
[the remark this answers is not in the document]
DanielleCogs: done!

libs/data-mapper/src/lib/utils/ReactFlow.Util.ts
DanielleCogs: done!

libs/data-mapper/src/lib/utils/__test__/DataMapUtils.spec.ts
DanielleCogs: done!

PR #1484, opened 2023-01-09
[DM | Deserialization] Direct Access support (non-loop)

NicL9923:
NOTE: May already work mostly/completely with in-loop DA's as well, but that'll be verified in its separate task

image

AB#16746415

### DIFF
--- a/libs/data-mapper/src/lib/mapDefinitions/MapDefinitionDeserializer.ts
+++ b/libs/data-mapper/src/lib/mapDefinitions/MapDefinitionDeserializer.ts
@@ -2,7 +2,7 @@
 import { mapNodeParams, reservedMapDefinitionKeysArray } from '../constants/MapDefinitionConstants';
 import { sourcePrefix, targetPrefix } from '../constants/ReactFlowConstants';
 import { addParentConnectionForRepeatingElements } from '../core/state/DataMapSlice';
-import type { FunctionData, MapDefinitionEntry, SchemaExtended, SchemaNodeDictionary } from '../models';
+import type { FunctionData, MapDefinitionEntry, SchemaExtended, SchemaNodeDictionary, SchemaNodeExtended } from '../models';
 import { SchemaType } from '../models';
 import type { ConnectionDictionary } from '../models/Connection';
 import { setConnectionInputValue } from '../utils/Connection.Utils';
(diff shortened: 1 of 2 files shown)


libs/data-mapper/src/lib/mapDefinitions/MapDefinitionDeserializer.ts
```diff
@@ -192,19 +192,37 @@ const createConnections = (
...
+  let mockDirectAccessFnKey: string | undefined = undefined;
+  const [daOpenBracketIdx, daClosedBracketIdx] = [amendedSourceKey.indexOf('['), amendedSourceKey.indexOf(']')];
 
-  const sourceNode =
-    sourceEndOfFunction > -1
-      ? findFunctionForFunctionName(amendedSourceKey.substring(0, sourceEndOfFunction), functions)
-      : findNodeForKey(amendedSourceKey, sourceSchema.schemaTreeRoot);
+  // Identify source schema node, or Function(Data) from source key
+  let sourceNode: SchemaNodeExtended | FunctionData | undefined = undefined;
+  if (sourceEndOfFunction >= 0) {
```
NicL9923, line 204: Callout: Felt ">= 0" ever so slightly better represented the condition being tested than "> -1"

Can revert if any objections

libs/data-mapper/src/lib/mapDefinitions/MapDefinitionDeserializer.ts
```diff
@@ -192,19 +192,37 @@ const createConnections = (
...
+    // We found a Function in source key -> let's find its data
+    sourceNode = findFunctionForFunctionName(amendedSourceKey.substring(0, sourceEndOfFunction), functions);
+  } else if (daOpenBracketIdx >= 0 && daClosedBracketIdx >= 0) {
+    // One of the source key's chunks contained a Direct Access, so let's format it
+    // into the Function syntax the deserializer can parse
+    sourceNode = directAccessPseudoFunction;
+
+    mockDirectAccessFnKey = `directAccess(`;
```
NicL9923, line 212: Callout: Could obviously include "directAccess(" and ")" in the inner string literals, but I personally prefer them as is to sort of clearly visually mark this string-building block

refortie (approved):

NicL9923:
Sure - it's the same condition, it just felt it made a little more sense, due to the fact that we're checking if it passed (meaning we got a valid index >=0 back), to have that as the actual condition/check, but not even remotely a big deal, so I'll revert it real quick

PR #1498, opened 2023-01-12
[DM | Deserialization] Index variable + direct access test case

NicL9923:
+ fixes index variable input being wrong source loop node


__mocks__/mapDefinitions/FullTranscriptMapDefinition.ts
```diff
@@ -45,7 +45,7 @@ ns0:Root:
     Trips:
       $for(/ns0:Root/Looping/VehicleTrips/Trips, $i):
         Trip:
-          VehicleRegistration: /ns0:Root/Looping/VehicleTrips/Vehicle[is-equal(VehicleID, /ns0:Root/Looping/VehicleTrips/Trips[$i]/VehicleId)]/VehicleRegistration
+          VehicleRegistration: /ns0:Root/Looping/VehicleTrips/Vehicle[is-equal(VehicleId, /ns0:Root/Looping/VehicleTrips/Trips[$i]/VehicleId)]/VehicleRegistration
```
refortie, line 48: This is currently testing nothing in it's use case: `is-equal(VehicleId, /ns0:Root/Looping/VehicleTrips/Trips[$i]/VehicleId)`

Both VehicleId are referring to the Trips vehicle id. I wonder if it should be: `is-equal(VehicleId, /ns0:Root/Looping/VehicleTrips/Vehicle[$i]/VehicleID)` instead.

libs/data-mapper/src/lib/mapDefinitions/__test__/MapDefinitionDeserializer.spec.ts
```diff
@@ -619,14 +619,14 @@ describe('mapDefinitions/MapDefinitionDeserializer', () => {
...
-    it.skip('creates a looping connection w/ index variable and direct access', () => {
+    it('creates a looping connection w/ index variable and direct access', () => {
       simpleMap['ns0:Root'] = {
         Looping: {
           Trips: {
             '$for(/ns0:Root/Looping/VehicleTrips/Trips, $i)': {
               Trip: {
                 VehicleRegistration:
-                  '/ns0:Root/Looping/VehicleTrips/Vehicle[is-equal(VehicleID, /ns0:Root/Looping/VehicleTrips/Trips[$i]/VehicleId)]/VehicleRegistration',
+                  '/ns0:Root/Looping/VehicleTrips/Vehicle[is-equal(VehicleId, /ns0:Root/Looping/VehicleTrips/Trips[$i]/VehicleId)]/VehicleRegistration',
```
refortie, line 629: Same comment as above

refortie (approved):

__mocks__/mapDefinitions/FullTranscriptMapDefinition.ts
NicL9923: Will follow up w/ backend to revisit Transcript map definition for cases such as these

PR #1501, opened 2023-01-12
[DM] Quick cleaning

NicL9923:


libs/data-mapper/src/lib/core/state/DataMapSlice.ts
```diff
@@ -224,31 +223,15 @@ export const dataMapSlice = createSlice({
...
     removeSourceSchemaNodes: (state, action: PayloadAction<SchemaNodeExtended[]>) => {
-      // TODO: So far we only ever remove one node at a time, but if that changes, we need to alter this
+      // NOTE: So far we only ever remove one node at a time, but if that changes, we need to alter this
       // as currently each node deletion will generate a new undo/redo state
       action.payload.forEach((srcSchemaNode) => {
         deleteNodeWithKey(state, addSourceReactFlowPrefix(srcSchemaNode.key));
       });
     },
 
-    toggleSourceSchemaNode: (state, action: PayloadAction<SchemaNodeExtended>) => {
```
NicL9923, line 235: NOTE: This wasn't used anywhere else at all, so pretty sure it was just our old version of the current "addSourceSchemaNodes"

refortie: Yeah this was used in the original implementation of adding nodes to the canvas.

refortie (approved):

PR #1530, opened 2023-01-19
[DM] Update Utility to Utilities to match latest Function manifest

NicL9923:


refortie (approved):

libs/data-mapper/src/lib/constants/FunctionConstants.tsx
```diff
@@ -45,7 +45,7 @@ export const stringBranding: FunctionGroupBranding = {
 };
 
 export const utilityBranding: FunctionGroupBranding = {
-  displayName: 'Utility',
+  displayName: 'Utilities',
```
refortie, line 48: We might want to double check with Alex on this. We might want to keep the branding display singular

NicL9923: I don't even know that we use this particular property anywhere - let me double check on that, but otherwise I know the Function List uses the value of FunctionCategory.Utility anyways as is

NicL9923: Yeah we only use the 'displayName' property for FunctionData's, never FunctionGroupBranding

PR #1540, opened 2023-01-20
[DM | Mocks + Standalone] Comprehensive mock schemas + map definition & Standalone improvements

NicL9923:
API Tester:
image

Wrapping Dev Toolbox:
image


__mocks__/schemas/ComprehensiveSourceSchema.json
```diff
@@ -0,0 +1,454 @@
...
+  "name": "ComprehensiveSource.xsd",
+  "type": "XML",
+  "targetNamespace": "http://tempuri.org/ComprehensiveSource.xsd",
+  "namespaces": {
+    "ns0": "http://tempuri.org/ComprehensiveSource.xsd",
+    "xs": "http://www.w3.org/2001/XMLSchema"
+  },
+  "schemaTreeRoot": {
+    "key": "/ns0:SchemaRoot",
+    "name": "SchemaRoot",
```
refortie, line 11: I might update this to "Source" (SourceSchemaRoot) and all the subsequent ones, that way it's easy to look at a map and see that source/target is being assigned to the correct spots, even if the names are the same

__mocks__/schemas/PlaygroundSourceSchema.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "PlaygroundSourceSchema.xsd",
+  "type": "XML",
+  "targetNamespace": "http://tempuri.org/Source.xsd",
+  "namespaces": { "ns0": "http://tempuri.org/Source.xsd", "xs": "http://www.w3.org/2001/XMLSchema" },
+  "schemaTreeRoot": {
+    "key": "/ns0:PlaygroundRoot",
+    "name": "PlaygroundRoot",
```
refortie, line 8: Same comment as above to add in Source/Target to the path so it's easy to see at a glance

libs/data-mapper/src/lib/mapDefinitions/__test__/MapDefinitionDeserializer.spec.ts
```diff
@@ -579,7 +579,7 @@ describe('mapDefinitions/MapDefinitionDeserializer', () => {
       );
     });
 
-    it('creates a direct index connection', () => {
+    it('creates a direct access connection', () => {
```
refortie, line 582: Nit: this is actually specifically for custom value index access instead of using the index function. Maybe `creates a custom value direct access connection`?

refortie (approved):

PR #1567, opened 2023-01-27
Adding swift mt operations to list of supported manifests

msnehagup:
The new designer is unable to load the card for the SWIFT MT Decode and Encode actions because these operations are not present in the list of supported manifests. Old designer recognizes these already and is able to display the cards.

AB#16907547


msnehagup:
@hartra344 / @preetriti1 - can you please review?

hartra344 (approved):

PR #1587, opened 2023-01-31
[Data Mapper] Add FunctionIcon component

refortie:
Adding in a FunctionIcon component so that we can de-couple the icon from the internet if need by (aka the customer can run the data mapper completely offline using the fallback icons)


libs/data-mapper/src/lib/components/nodeCard/functionCard/ExpandedFunctionCard.tsx
```diff
@@ -62,6 +64,7 @@ export const ExpandedFunctionCard = (props: NodeProps<FunctionCardProps>) => {
   const handlesForCollapsedHeader = functionData.inputs.map((input) => {
     return <Handle key={input.name} id={input.name} type="target" position={Position.Left} style={{ top: '20px', visibility: 'hidden' }} />;
   });
+
```
NicL9923, line 67: Nice, this is a good piece of code here

NicL9923 (approved):

PR #1597, opened 2023-02-02
[Data Mapper] Map Checker

refortie:
image

image

image


rickixie:
Hey - the icon in the individual error card is a little bit big - i was thinking reusing the same toast we have in here. Is that possible

refortie:
> Hey - the icon in the individual error card is a little bit big - i was thinking reusing the same toast we have in here. Is that possible

Can do. I'll reference that size

PR #1598, opened 2023-02-02
LaFeat: Load dynamic run instance data to monitoring view mode

ccastrotrejo:
### Main Code Changes
- Implementation of loading and displaying Run mocked data dynamically through the toolkit dropdown
- Set Run data in each node's metadata
- Implementation of properties panel in monitoring view
- Implement dynamic status and duration in card and scopecard
- Move monitoring view tab code from `designer-ui` to `designer` folder
- Added run instance mock data for simple scoped workflow
- Added `RunInstanceDefinition` to logic apps types

### What is missing
- Get the retry history to differentiate between `SucceededWithRetries` and `Succeeded` status icons
- Outputs and inputs panels in monitoring tab when click in connector. Currently with static data

### Implementation

https://user-images.githubusercontent.com/102700317/216443126-37c00bdf-d3ae-4c4d-8f02-3587d95768dc.mov

AB#17012949


libs/designer/src/lib/ui/CustomNodes/OperationCardNode.tsx
```diff
@@ -39,6 +39,8 @@ const DefaultNode = ({ targetPosition = Position.Top, sourcePosition = Position.
   const operationInfo = useOperationInfo(id);
   const isTrigger = useMemo(() => metadata?.graphId === 'root' && metadata?.isRoot, [metadata]);
 
+  const { status: statusRun, duration: duratioRun, error: errorRun } = metadata?.runData ?? {};
```
rllyy97, line 42: Typo on `duratioRun`

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
```diff
@@ -44,6 +44,8 @@ const ScopeCardNode = ({ data, targetPosition = Position.Top, sourcePosition = P
   const graphNode = useWorkflowNode(scopeId) as WorkflowNode;
   const metadata = useNodeMetadata(scopeId);
 
+  const { status: statusRun, duration: duratioRun, error: errorRun } = metadata?.runData ?? {};
```
rllyy97, line 47: Same typo on `duratioRun`

libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
```diff
@@ -323,3 +333,46 @@ const throwIfMultipleTriggers = (definition: LogicAppsV2.WorkflowDefinition) =>
     );
   }
 };
+
+const addTriggerInstanceMetaData = (runInstance: LogicAppsV2.RunInstanceDefinition | null) => {
+  if (!isNullOrUndefined(runInstance)) {
```
rllyy97, line 338: Just a tiny nit but something to help keep code cleaner, I do this all the time:
This is a condition where we can just invert the condition at the top, and return right away instead of having the whole function inside an if statement and then returning at the end

rllyy97 (approved):
I had a couple nits but otherwise looks good!

preetriti1:
Why do you have other tabs like settings, codeview, parameters along with monitoring? Are these tabs in editable state, if yes then are we allowing them to update and save the definition.

ccastrotrejo:
@preetriti1 All of the tabs are in read-only mode, so user are not able to edit/save those values.

PR #1600, opened 2023-02-03
[DM | Theming] Add Fluent PortalCompatProvider

NicL9923:
-Gets our tooltips showing up in Standalone

-Can also, should the need arise, tweak things so that V9 controls maintain their styles on V8 surfaces (Ex: Panel) (see: https://codesandbox.io/s/fluentui-v9-theming-v8-surfaces-forked-sskgei?file=/src/index.js)

Don't ask:
image

Function node tooltips:
image


libs/data-mapper/src/lib/core/DataMapperDesignerProvider.tsx
```diff
@@ -49,34 +50,37 @@ export interface DataMapperDesignerProviderProps {
...
-// NOTE: Leaving ThemeProvider here as we still use Fluent V8 components
 export const DataMapperDesignerProvider = ({ theme = 'light', locale = 'en', options, children }: DataMapperDesignerProviderProps) => {
   return (
     <AppInsightsContext.Provider value={reactPlugin}>
       <ReduxProvider store={store}>
         <DataMapperWrappedContext.Provider value={options}>
-          <FluentProvider
-            theme={theme === 'light' ? extendedWebLightTheme : extendedWebDarkTheme}
+          <ThemeProvider
+            theme={theme === 'light' ? AzureThemeLight : AzureThemeDark}
```
ccastrotrejo, line 59: Very small nit, maybe you can use the Theme enum in here `libs/utils/src/lib/models/theme.ts`

ccastrotrejo (approved):

libs/data-mapper/src/lib/core/DataMapperDesignerProvider.tsx
NicL9923: @ccastrotrejo Could you take a look at why the tests are failing the designer-ui? The only changes I made that could've affected that were bumping the FluentUI versions, but I wouldn't think they'd do that much damage...

PR #1602, opened 2023-02-03
LaFeat: Monitoring tab

ccastrotrejo:
### Main Code Changes
- Update default tabs (not showing parametersTab and show monitoringTab when is in monitoring mode)
- Add host service to list of designer provider services
- Provide readOnly to settings tab (runAfter, security, tracking, general, dataHandling)
- Implement fetchAndDisplayContent when click in `See raw inputs` and `See raw ouputs`

##### Monitoring tab in new designer
image

##### Monitoring tab in portal

image

### Implementation

https://user-images.githubusercontent.com/102700317/216733484-5607985f-aa1e-4aab-bd22-866390c87ca5.mov


Eric-B-Wu (approved):
I'm wondering if we should rename the tab something else as it's not limited to just parameters (but would probably need to discuss with Parth)

PR #1604, opened 2023-02-04
Add Lexical to dependencies and make React a peer dependency

ek68794998:
This PR has a few changes:

1. Update `reactflow` imports to use `~reactflow`. This should prevent downstream consumers from having to manually include the CSS files. From William: *... the current import is relative and does not work, the proposed change works with webpack's css-loader (https://webpack.js.org/loaders/css-loader/#import).*
2. Move React (and ReactDOM) to be a peer dependency. This is the recommendation from the React team, so that consumers can use their own versions of React. From William: *It's expected consumers will already have a version of react specified and conflicting versions are problematic as explained here: https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react*
3. Move Lexical to be a direct dependency. The project directly consumes Lexical but does not currently include it in `package.json`.

AB#17017890


wifun (approved):

ek68794998:

hartra344 (changes requested):
This change removes react as a dependency from the generated package.json but it doesn't add it as a peer dependency. This package.json is the monorepo overarching package.json not the one that gets generated for individual libs. To make react a peer dep you'll need to dos omething with the workspace config. NX should have some docs on this.

ek68794998:
> This change removes react as a dependency from the generated package.json but it doesn't add it as a peer dependency. This package.json is the monorepo overarching package.json not the one that gets generated for individual libs. To make react a peer dep you'll need to dos omething with the workspace config. NX should have some docs on this.

This seems impossible. Based on the Nx documentation, you have to either select `dependencies` or `peerDependencies` for it to pull from the monorepo `project.json`. You can't mix both&mdash;this is controlled by `buildableProjectDepsInPackageJsonType` which can either make **everything** a dependency, or **everything** a peer dependency. (`https://github.com/nrwl/nx/issues/10550`) 

The only way I can get around this is by adding the peer dependencies for `react` and `react-dom` to all of the buildable projects&mdash;that is, `.\libs\designer\package.json` and similar.

@hartra344 Is the above an acceptable workaround? If not, do you have an alternative suggestion, given that you are likely more familiar with Nx than I am?

hartra344:
> > This change removes react as a dependency from the generated package.json but it doesn't add it as a peer dependency. This package.json is the monorepo overarching package.json not the one that gets generated for individual libs. To make react a peer dep you'll need to dos omething with the workspace config. NX should have some docs on this.
> 
> This seems impossible. Based on the Nx documentation, you have to either select `dependencies` or `peerDependencies` for it to pull from the monorepo `project.json`. You can't mix both—this is controlled by `buildableProjectDepsInPackageJsonType` which can either make **everything** a dependency, or **everything** a peer dependency. (`https://github.com/nrwl/nx/issues/10550`)
> 
> The only way I can get around this is by adding the peer dependencies for `react` and `react-dom` to all of the buildable projects—that is, `.\libs\designer\package.json` and similar.
> 
> @hartra344 Is the above an acceptable workaround? If not, do you have an alternative suggestion, given that you are likely more familiar with Nx than I am?

If there's no config for it then it's likely that it would need a custom build executor which is something I would prefer not to do. The adding it ot the package.json workaround will likely be fine but only add it to /designer and /designer-ui. There are other libs that do not require react so don't want to add it to those. They are used in other projects.

ek68794998:
Updated to move `react` and `react-dnd` peer dependencies to their two respective projects.

libs/designer/package.json
```diff
@@ -3,5 +3,9 @@
   "version": "0.2.68",
   "scripts": {
     "yalcpush": "yalc push ../../dist/libs/designer"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0 || ^17.0.0 || ^18.0.0",
```
wifun, line 8: There were 4 packages that listed react as a dep, is changing these 2 files enough?
They were:
* @microsoft/logic-apps-designer
* @microsoft/designer-ui
* @microsoft-logic-apps/intl
* @microsoft-logic-apps/utils
ex. https://msazure.visualstudio.com/OneAgile/_artifacts/feed/power-platform-ux/Npm/@microsoft-logic-apps%2Futils/overview/0.2.30

ek68794998: Need Travis to chime in on this one. I did find a reference in `libs/utils` to React:

image

ek68794998: `libs/intl` has no `react` or `react-dom` references; `libs/utils` has no `react-dom` references.

Therefore I think adding this to `libs/utils/package.json` should be sufficient?

image

ek68794998: Spoke with Travis offline&mdash;we don't want React in `libs/utils`, so he will be removing that reference.

hartra344 (approved):

PR #1625, opened 2023-02-09
[DM | Map Checker] Slightly polish no-items scenario

NicL9923:
image


refortie (approved):

rickixie:
lol let's not use the clippy image. I will send you a empty state placeholder.

rickixie:
image
We can just do something like this. But if we want to have some easter egg now clippy is okay for public preview :)

NicL9923:
For future reference let's please make a work item to follow up - speaking for at least myself, I don't reference PRs once closed

PR #1634, opened 2023-02-13
[DM | Schema tree] Filter by data types

NicL9923:
image

*Fixed icon for Array not acting right since taking this screenshot
image

image

image

AB#17159587


libs/data-mapper/src/lib/models/Schema.ts
```diff
@@ -40,15 +40,15 @@ export enum SchemaNodeProperty {
 }
 
 export enum NormalizedDataType {
+  Any = 'Any',
```
NicL9923, line 43: NOTE: Just alphabetizing here

libs/data-mapper/src/lib/components/functionList/FunctionList.tsx
```diff
@@ -154,9 +157,40 @@ export const FunctionList = () => {
...
+  });
+
+  const clearLoc = intl.formatMessage({
+    defaultMessage: 'Clear',
+    description: 'Clear',
+  });
+
   return (
     <>
-      <TreeHeader onSearch={setSearchTerm} onClear={() => setSearchTerm('')} />
```
NicL9923, line 159: NOTE: Un-using the abstraction here in favor of specializing the schema tree one (without trying to conditional-ize stuff) - figured it's an okay tradeoff as FunctionList is pretty simple, and it'd be extremely low effort to abstract the new stuff here if we did need a reusable *simple* searchbar (i.e. this) in the future

NicL9923:
Disregard Test Runner failure - unrelated

libs/data-mapper/src/lib/components/tree/SchemaTreeSearchbar.tsx
```diff
@@ -0,0 +1,182 @@
...
+                      <Checkmark20Filled style={{ visibility: filteredDataTypes[dataTypeValue] ? 'visible' : 'hidden' }} />
+                      <DataTypeIcon
+                        style={{
+                          marginRight: 4,
+                          height: 20,
+                          color: filteredDataTypes[dataTypeValue] ? tokens.colorCompoundBrandForeground1 : undefined,
+                        }}
+                        filled={filteredDataTypes[dataTypeValue]}
+                      />
+                      <Text>{dataTypeKey}</Text>
```
refortie, line 150: We might want to use a localized name or at the very least display name. I think "ComplexType" should at least be "Complex Type" and tbh that might be better as "Object"

libs/data-mapper/src/lib/components/tree/SchemaTreeSearchbar.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    visibility: Object.values(filteredDataTypes).some((isFiltered) => !isFiltered) ? 'hidden' : 'visible',
+                    marginRight: 2,
+                  }}
+                />
+                <Text style={{ color: tokens.colorBrandForegroundLink }}>Select all</Text>
+              </Stack>
+            </MenuItem>
+
+            {Object.entries(NormalizedDataType).flatMap(([dataTypeKey, dataTypeValue]) => {
+              const getMenuItemForDataType = (dataTypeKey: string, dataTypeValue: DataTypeFilterKeys) => {
```
refortie, line 132: I think this is the spot where the icon is. Should we be using the colored icons or not, depending on whether they are selected? It feels weird that they are changing depending on selected

refortie (approved):

libs/data-mapper/src/lib/components/tree/SchemaTreeSearchbar.tsx
NicL9923: Per the spec, yes

libs/data-mapper/src/lib/components/tree/SchemaTreeSearchbar.tsx
NicL9923: I agree on 'Object', but otherwise I would say most of the types (the keys of NormalizedDataType which is what's displayed) are already pretty well matched up with what I'd expect to see (Boolean, String, Number, etc)

PR #1640, opened 2023-02-14
[DM | Function list] Search by category name + enumerate FunctionData.type

NicL9923:
image

image

### DIFF
--- a/libs/data-mapper/src/lib/models/Function.ts
+++ b/libs/data-mapper/src/lib/models/Function.ts
@@ -6,6 +6,11 @@ export interface FunctionManifest {
   transformFunctions: FunctionData[];
 }
 
+export enum FunctionType {
+  PseudoFunction = 'PseudoFunction',
+  TransformationFunction = 'TransformationFunction',
+}
+
 export interface FunctionData {
   key: string;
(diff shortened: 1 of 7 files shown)


libs/data-mapper/src/lib/models/Function.ts
```diff
@@ -6,6 +6,11 @@ export interface FunctionManifest {
   transformFunctions: FunctionData[];
 }
 
+export enum FunctionType {
+  PseudoFunction = 'PseudoFunction',
+  TransformationFunction = 'TransformationFunction',
```
refortie, line 11: Did you pull the manifest and check to see if there are other ones of these?

refortie (approved):

libs/data-mapper/src/lib/models/Function.ts
NicL9923: I did not, lemme do that just to confirm

NicL9923: Yeah, all the ones that come from the manifest are "TransformationFunction", so all good there.

Side note, that makes me wonder what its purpose is in the first place :P

PR #1647, opened 2023-02-15
[Data Mapper] Major reworks to ReactFlow.utils.ts

refortie:
* Added WholeMapOverview - tentative name. No connections, no functions yet
* Reworked how widths are calculated, should hopefully be sturdier and simpler to follow
* Fixed bug in minimap node sizes
* Refactored react flow node creation to go back to using 1 method for both source and target

Before Overview:
image

After Overview (should be no difference):
image

Before Mapping:
image

After Mapping (should be no difference):
image

Whole map:
image

AB#17192386


NicL9923 (approved):

refortie:
image

Fix for status icons in the overview

PR #1648, opened 2023-02-15
[DM | Mocks] Add JSON schemaTrees + remove schemas now encompassed by Comprehensive

NicL9923:
*see commit messages for more details*

image


__mocks__/mapDefinitions/NestedLoopMapDefinition.ts
```diff
@@ -1,20 +0,0 @@
-export const nestedLoopMapDefinition = `$version: 1
```
refortie, line 1: Comprehensive already has a nested loop example in it? I don't see any update here to add one

refortie (approved):

__mocks__/mapDefinitions/NestedLoopMapDefinition.ts
NicL9923: Correct, it covers all of our base scenarios already (including ManyToMany, ManyToOne, etc)

PR #1677, opened 2023-02-27
[DM | Overview] Rename WholeView to GlobalView and expose it on the command bar

NicL9923:
image


Localize/lang/strings.json
```diff
@@ -526,6 +527,7 @@
   "TPV6j/": "Paging count invalid",
   "TRpSCQ": "Action Type",
   "TUaunO": "Enter a value",
+  "TVrS/G": "Whole view",
```
refortie, line 530: Elliott suggested the name "Global View" and I think I like it better then our current choices

refortie (changes requested):

PR #1698, opened 2023-03-01
[VSFeat] Monitoring implementation

ccastrotrejo:
### Main Code Changes
- Full functional in local and remote workflows
- Add monitoring view logic for remote and local runs
- Add base monitoring view class that extends the base designer class
- Add monitoring view designer command bar items

### Pending
- Refresh and resubmit buttons in command bar
- Secure Inputs and Outputs 
- Implement a try catch for run service

#### Local workflow
https://user-images.githubusercontent.com/102700317/222278120-a33d2280-e19f-4ad6-832b-0ce3a393d065.mov

#### Remote workflow

https://user-images.githubusercontent.com/102700317/222284711-1b2e2e07-f7f4-4e09-9735-9ee1c01618e1.mov


libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
```diff
@@ -360,7 +360,7 @@ const addActionsInstanceMetaData = (nodesMetadata: NodesMetadata, runInstance: L
   const updatedNodesData = { ...nodesMetadata };
 
   Object.entries(updatedNodesData).forEach(([key, node]) => {
-    const nodeRunData = runInstanceActions[key];
+    const nodeRunData = runInstanceActions && runInstanceActions[key];
```
rllyy97, line 363: Fun fact (or nit) optional chaining works with [ ] property references too, so you could do this and it makes it null safe on both sides, returning undefined if either fails
```suggestion
    const nodeRunData = runInstanceActions?.[key];
```

libs/designer/src/lib/ui/panel/panelTabs/monitoringTab/monitoringTab.tsx
```diff
@@ -5,21 +5,32 @@ import { useNodeMetadata } from '../../../../core/state/workflow/workflowSelecto
...
   const runMetaData = nodeMetadata?.runData;
+  const [inputOutputs, setInputsOutputs] = useState({ inputs: {}, outputs: {} });
+
+  useEffect(() => {
+    async function getActionInputsOutputs() {
+      const actionsInputsOutputs = await RunService().getActionLinks(runMetaData, selectedNodeId);
+      setInputsOutputs(actionsInputsOutputs);
+    }
+    getActionInputsOutputs();
+  }, [runMetaData]);
```
rllyy97, line 27: This might be better stored as a useQuery if we're not doing any modification of these values, then we will also get access to things like error state data and loading states

libs/services/designer-client-services/src/lib/standard/run.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (!accessToken) {
+      return undefined;
+    }
+
+    return new Headers({
+      Authorization: accessToken,
+    });
+  };
+
+  async getMoreRuns(continuationToken: string): Promise<Runs> {
```
rllyy97, line 69: Our search service has a couple methods set up already to handle requests with continuation tokens too, this may not be for this PR but we should at some point pull those out to a generic file so they can use the same logic (they're essentially doing the same thing just with different data types).  If you want to look at them they are `getAzureResourceRecursive` for requesting in series and `batchAzureResourceRequests` for requesting in parallel

rllyy97 (approved):

libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
ccastrotrejo: Thanks, I will update it

libs/designer/src/lib/ui/panel/panelTabs/monitoringTab/monitoringTab.tsx
ccastrotrejo: Sure thing, I will move them to useQuery

libs/services/designer-client-services/src/lib/standard/run.ts
ccastrotrejo: Thanks! Yeah I was planning on doing a separate PR for the run service.

libs/services/designer-client-services/src/lib/standard/run.ts
```diff
@@ -0,0 +1,188 @@
...
+    const response = await httpClient.get<any>({
+      uri,
+    });
+    return response;
+
+    /*const contentType = response.headers.get('Content-Type');
+    if (contentType?.startsWith('application/json')) {
+      return response.json();
+    } else if (contentType?.startsWith('text/')) {
+      return response.text();
```
Eric-B-Wu, line 48: nit: remove

Eric-B-Wu (approved):

Eric-B-Wu (approved):

PR #1701, opened 2023-03-02
[Data Mapper] Deleting unbounded inputs should leave the total inputs alone

refortie:


libs/data-mapper/src/lib/core/state/DataMapSlice.ts
```diff
@@ -726,32 +736,18 @@ export const deleteNodeWithKey = (curDataMapState: DataMapState, reactFlowKey: s
...
     return;
   }
 
   // Item to be deleted is a connection
   const connections = { ...curDataMapState.curDataMapOperation.dataMapConnections };
 
-  /*
-    const sourceId = getSourceIdFromReactFlowConnectionId(reactFlowKey);
-    const sourceSchema = curDataMapState.curDataMapOperation.flattenedSourceSchema;
-    const canDelete = canDeleteConnection(sourceId, sourceSchema);
```
NicL9923, line 742: I can't recall if we use canDeleteConnection() anywhere else - if not, might be good to go ahead and delete that too (and its associated test, again IIRC)

NicL9923 (approved):

libs/data-mapper/src/lib/core/state/DataMapSlice.ts
refortie: Can do

PR #1706, opened 2023-03-03
Remove styles targeting host layer's 'body'

ek68794998:
Currently the designer library has some code in it which targets `body` and `body > #root`, which are app-layer DOM elements that should not be affected by library styles. This PR removes those styles.

AB#17424033


hartra344:
@rllyy97 before merge this in, can you go ahead and add this to CSS in portal so it doesn't break when we forget to?

ek68794998:
Superseded by PR #1710.

PR #1707, opened 2023-03-04
[Data Mapper][JSON] JSON object arrays serialization

refortie:


libs/data-mapper/src/lib/mapDefinitions/MapDefinitionSerializer.ts
```diff
@@ -24,6 +24,7 @@ import yaml from 'js-yaml';
 interface OutputPathItem {
   key: string;
   value?: string;
+  arrayIndex?: number;
```
refortie, line 27: This is so that we can eventually support :

```yml
  Array1: 
  - /root/Count
  - /root/Count1
  - /root/Num
  ```
  
  where we need to be able to let the customer add new target nodes/assign multiple inputs to target nodes in a specific order. Right now this will only ever be '0'

NicL9923 (approved):

PR #1721, opened 2023-03-08
Add support for OpenAPI parameter types

ek68794998:
## Primary changes

Currently, loading OpenAPI actions results in a duplicated set of parameters (and the duplicates are the ones populated with any previously saved data):

image

This change adds proper handling for OpenAPI actions so that they are not duplicated:

image

## Other changes

- Due to an issue we had in the past, I've updated the logic from `equals()` to handle Greek and Latin letters as well

AB#17466033


edgarguoms:
Looks good to me.

edgarguoms (approved):

hartra344 (approved):

PR #1724, opened 2023-03-09
Remove fuse.js as a dev dependency

ek68794998:
The `fuse.js` library is set as both a dependency and dev dependency, which causes Nx to get confused and not bundle it in `package.json`.

I made the change locally and did a full clean and rebuild before and after. Below is the `dist/libs/designer/package.json` content before and after.

## Before

image

## After

image

AB#17490406


hartra344 (approved):

rllyy97 (approved):

ek68794998:
@rllyy97 or @hartra344 are you able to complete the PR? Even with approvals I do not have permission to complete it. Thanks!

PR #1737, opened 2023-03-09
Fix display names of enums using 'x-ms-enum-values'

ek68794998:
## Issue

Many connectors return an object matching the following shape:

```js
{
  enum: {
    // whole objects
    { displayName: 'Aa', value: 'a' },
    { displayName: 'Bb', value: 'b' },
  },
  schema: {
    enum: { 'a', 'b' }, // values only
    ...,
  },
  ...,
}
```

However, OpenAPI connectors return their data in the following shape instead:

```js
{
  schema: {
    enum: { 'a', 'b' }, // values only
    'x-ms-enum-values': {
      // whole objects
      { displayName: 'Aa', value: 'a' },
      { displayName: 'Bb', value: 'b' },
    },
    ...,
  },
  ...,
}
```

In the current implementation, these OpenAPI display name are ignored, resulting in the raw values being displayed. For example:

image

## Changes

The changes in this pull request update the logic so that `x-ms-enum-values` is also accounted for.

image

## Tests

The `helper.spec.ts` file has been updated with tests. These tests were written **before** making changes to `helper.ts` to prevent regressions, and all were passing *except* the test that uses `x-ms-enum-values`, which failed with the following error:

image

After this change, all tests pass.

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/__test__/helper.spec.ts
+++ b/libs/designer/src/lib/core/utils/parameters/__test__/helper.spec.ts
@@ -1,1063 +1,1041 @@
-import { parameterValueToJSONString, parameterValueToString } from '../helper';
-import type { ParameterInfo, ValueSegment } from '@microsoft/designer-ui';
+import { getParameterEditorProps, parameterValueToJSONString, parameterValueToString } from '../helper';
+import type { DictionaryEditorItemProps, ParameterInfo, ValueSegment } from '@microsoft/designer-ui';
 import { TokenType, ValueSegmentType } from '@microsoft/designer-ui';
-import { ExpressionType } from '@microsoft/parsers-logic-apps';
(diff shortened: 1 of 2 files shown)


libs/designer/src/lib/core/utils/parameters/__test__/helper.spec.ts
```diff
@@ -1,131 +1,530 @@
-import { parameterValueToJSONString, parameterValueToString } from '../helper';
-import type { ParameterInfo, ValueSegment } from '@microsoft/designer-ui';
+import { getParameterEditorProps, parameterValueToJSONString, parameterValueToString } from '../helper';
```
ek68794998, line 1: This diff is terrible&mdash;all the existing code is unchanged except that it's wrapped in a new `describe` block.

hartra344 (approved):

edgarguoms (approved):

PR #1760, opened 2023-03-15
laFeat: Search + Browse speed boost

rllyy97:
## Main Changes
_**SPEED**_ (Or at least the illusion of speed)

- Converted our main preload `useQuery` to a `useInfiniteQuery`
  - UIQ is normally used for loading paged content / infinite scroll content
  - Instead I set it up to load our batches of resource requests sequentially
- Split operations preload from connectors preload
  - Split connectors preload up into Azure, Custom, Built-In requests

This has two main outcomes:
- Connectors load almost instantly, and if it's not fully done it's enough to fill the view
- Operations populate at varying levels of speed, but from just playing around with it it does have the illusion of being faster
  - It will potentially lead to faster searches, or the same, but should never be slower than it is currently
  - This is the fastest we can get operation content to render without getting faster endpoints 

This also makes search + browse a little more resilient since if there is an issue getting operations / connectors from one query, the rest of the queries are not held back.

## See for yourself 
(Both recordings I started the recording as soon as designer loaded inside of portal after a refresh)

### OLD CONNECTORS
old_preload

### NEW CONNECTORS
new_preload

---

### OLD SEARCH
old_preload_blogger

### NEW SEARCH 
(Ignore the loading spinner switching, it's fixed now)
((Not a huge speed difference but anything helps))

new_preload_blogger

AB#17575792


Eric-B-Wu:
Much needed change! Really improving the search experience

Eric-B-Wu (approved):

Eric-B-Wu:
also just want to make sure this doesn't affect any of the vscode services

ccastrotrejo (approved):

PR #1775, opened 2023-03-18
Add support for using ARM resources in the standalone designer for devleopment

hartra344:
NOTES:
* Requires Azure CLI to be installed in order to use. If you have azure cli installed and logged in you only need to run "npm generateArmToken" and it'll generate and place an arm token for the app to pick up
* There is no ARM batching to calls can be very slow, could be made better in the future?
* Currently no OAuth support, didn't migrate that service from portal yet
* There are no UX niceties like loading indicators. Could be added but I viewed as unimportant here since it's for development purposes

General Use:

https://user-images.githubusercontent.com/13208452/226130916-c9752f1d-17d1-4868-bafd-113e40f256a7.mp4

Search/Browse:

https://user-images.githubusercontent.com/13208452/226131129-d1c2dfbb-019e-4e08-a693-acd9bae79967.mp4

AB#17596385


Eric-B-Wu:
wow, this will be an amazing QOL improvement for us

Eric-B-Wu (approved):

PR #1780, opened 2023-03-19
Updated Lexical, Minor Changes

Eric-B-Wu:
Changes
- Updated Lexical to 0.9.0, fixed breaking changes
- Adding setting option to view local workflows even when having a valid ARM token
- Fixed Token Picker's expression editor being glitchy when decreasing expression editor size

Minor Changes
- Decreased AppId Dropdown vw from 100%, so that can click outside if no option is selected (when having a smaller window previously, the width of the dropdown would take the whole screen)
- Disabled Workflow Dropdown when no workflow options in logic app


apps/designer-standalone/src/environments/armToken.json
```diff
@@ -1 +0,0 @@
-{}
```
Eric-B-Wu, line 1: @hartra344 I know we have the pre-commit to prevent any token from ever populating this, but do we actually need this file? Seems like it can be created when generating ARM token, and by having this file deleted we prevent git from tracking, as the file is already in the repo

Eric-B-Wu: ah nevermind, i see why the file is needed when we grab the token in the environment.ts file

hartra344: yep, that's the reason haha, otherwise it required async loading stuff which would be flakey when the app first started. This was the easiest, most reliable way to do it

hartra344 (approved):

PR #1781, opened 2023-03-20
Add ability to search nodes with keyboard shortcuts

hartra344:
With mock data but large node set:

https://user-images.githubusercontent.com/13208452/226222785-514c454c-d368-4fa8-ab94-8f488a4bc994.mp4

With a real Logic App but small node set:

https://user-images.githubusercontent.com/13208452/226222975-e08ad072-98cb-4eba-87e3-7ef5e12451a1.mp4

maybe more we could do with this but this opens a lot of avenues for keyboard navigation


Eric-B-Wu (approved):

rllyy97 (approved):
I like it!  Huge usability improvement especially for larger workflows

PR #1782, opened 2023-03-20
Run parameter validaiton trigger when panel is collapsed

### LINKED ISSUE #1607: Workflow is saving without filling the mandatory fields
### Describe the Bug

Expected : .Workflow should not be saved also show a inline error line.

Actual : Workflow is saving without filling the mandatory fields.

If we create a new action with mandatory fields, we can save workflow immediately without triggering any mandatory validation.

This is happening because we are currently validating onChange, when we add a new action, the action is not in "dirty state" by design so that it doesn't automatically show errors. However maybe something should be done to prevent saving

### Workflow JSON

_No response_

### Expected behavior

Expected : .Workflow should not be saved also show a inline error line.

Actual : Workflow is saving without filling the mandatory fields.

### Screenshots or Videos

_No response_

### Platform And Browser

All Browsers and Platforms
saving issue

### Additional context

_No response_

AB#17036660

hartra344:
fixes #1607

https://user-images.githubusercontent.com/13208452/226255508-0372f3d6-2312-489d-9304-9137172f46e1.mp4

https://user-images.githubusercontent.com/13208452/226256096-f7c80dc7-ca7b-43ed-b3f4-77a79a1eb88a.mp4

AB#17606178


Eric-B-Wu (approved):

Eric-B-Wu:
do you know if this will also cover the case where we create a new action and then immediately click another node/immediately save? I know in preview designer we don't have any validation and can immediately save

rllyy97:
With this and Travis's portal change, I think the option I'm for is running validation over everything on command bar save, but before the actual save callback.  I feel like that's the least intrusive while still actually handling all validation errors 100% of the time.  (I still think validating on panel close is needed though)

Eric-B-Wu:
> With this and Travis's portal change, I think the option I'm for is running validation over everything on command bar save, but before the actual save callback. I feel like that's the least intrusive while still actually handling all validation errors 100% of the time. (I still think validating on panel close is needed though)

the validation check would happen on the designer side still, right?

PR #1789, opened 2023-03-21
laFix: Small dark mode changes

rllyy97:
A couple small dark-mode changes.
This was intended to be a fix for the dark mode issues we've been having but they look like they've been fixed on their own.

[Portal PR](https://msazure.visualstudio.com/One/_git/AzureUX-LogicAppsPortal/pullrequest/7817990)

AB#17625841


hartra344:
Once this is in, it might actually be worth a hotfix

PR #1798, opened 2023-03-22
Add `serializeHost` case for OpenAPI

ek68794998:
Currently, serialization of OpenAPI actions fails due to it not being handled by `serializeHost`. This PR adds a case to that method for handling OpenAPI.

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -455,6 +455,14 @@ interface FunctionConnectionInfo {
   };
 }
 
+interface OpenApiConnectionInfo {
+  host: {
+    apiId: string;
+    connectionName: string;
+    operationId: string;
+  };
+}
+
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -455,6 +455,14 @@ interface FunctionConnectionInfo {
   };
 }
 
+interface OpenApiConnectionInfo {
+  host: {
+    apiId: string;
+    connectionName: string;
```
wifun, line 461: should this be `connection` if it's related to this? https://github.com/Azure/LogicAppsUX/blob/main/types/LogicAppsV2.d.ts#L108

ek68794998: Interesting find. This makes sense. I can actually just use that type.

hartra344 (approved):

preetriti1 (approved):

PR #1799, opened 2023-03-22
laFix: Dynamic Data value initialization fix

rllyy97:
## Main Changes
- A recent change caused dynamic data value initialization to fail due to invalid key paths
- This fix adjusts to account for both the previous default paths, as well as the OpenApi key paths that were being added


hartra344 (approved):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -577,3 +581,8 @@ function getSwaggerTypeFromVariableType(variableType: string): string | undefine
       return undefined;
   }
 }
+
```
preetriti1, line 584: We should check here for aliasing supported know this parameter object, I think should present if it has alias then its open api format

preetriti1 (changes requested):

preetriti1 (approved):

PR #1818, opened 2023-03-27
Update parameter validation to use OpenAPI display names

ek68794998:
Currently, OpenAPI parameters which do not validate (i.e., have error messages) show the incorrect display name for the parameter:

image

This PR updates the logic to check if there is a schema associated with the parameter first and, if so, use that schema's title or summary fields to fill in the display name. Otherwise, fall back to the current logic.

Result:

image


ek68794998:
One thing I want to check is how this logic compares to the logic used above the field as well, for consistency.

PR #1873, opened 2023-03-31
lafeat: Dont preload search operations before finished loading workflow data

hartra344:
Preloading the search data too quickly meant the workflow data would often load much slower. We want to preload the search data but we don't want it to be at the expense of loading the  workflow data.

https://user-images.githubusercontent.com/13208452/229059324-ae307d25-1216-4440-8c69-52ad7c1518af.mp4


libs/designer/src/lib/ui/Designer.tsx
```diff
@@ -108,16 +109,20 @@ export const CanvasFinder = () => {
...
   const { backgroundProps } = props;
 
   const [nodes, edges, flowSize] = useLayout();
   const isEmpty = useIsGraphEmpty();
   const isReadOnly = useReadOnly();
   const dispatch = useDispatch();
-
-  usePreloadOperationsQuery();
-  usePreloadConnectorsQuery();
+  const [initializationStage, setInitializationStage] = useState<'Start' | 'Loading' | 'Finished'>('Start');
```
hartra344, line 125: Future note: This might be worth tracking in a way as well that portal can use it or redux can use it to start tracking dirty state.

Eric-B-Wu (approved):

rllyy97 (approved):

PR #1877, opened 2023-03-31
Export initializeServices for host services to update designer services

shuktika12163:
Currently, designer services are initialized once in BJSWorkflowProvider and stored globally. They are not re-initialized on re-mounting the component. Exporting the function for host services to re-initialize the designer services as per need.

### DIFF
--- a/libs/designer/src/lib/core/index.ts
+++ b/libs/designer/src/lib/core/index.ts
@@ -9,3 +9,4 @@ export { useNodeDisplayName } from './state/workflow/workflowSelectors';
 export { serializeWorkflow } from './actions/bjsworkflow/serializer';
 export { clearPanel, switchToWorkflowParameters } from './state/panel/panelSlice';
 export { useSelectedNodeId } from './state/panel/panelSelectors';
+export { initializeServices } from './state/designerOptions/designerOptionsSlice';
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/logicappsux",
-  "version": "2.0.14",
+  "version": "2.0.16",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@microsoft/logicappsux",
-      "version": "2.0.14",
+      "version": "2.0.16",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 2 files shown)


edgarguoms (approved):

libs/designer/src/lib/core/index.ts
```diff
@@ -9,3 +9,4 @@ export { useNodeDisplayName } from './state/workflow/workflowSelectors';
 export { serializeWorkflow } from './actions/bjsworkflow/serializer';
 export { clearPanel, switchToWorkflowParameters } from './state/panel/panelSlice';
 export { useSelectedNodeId } from './state/panel/panelSelectors';
+export { initializeServices } from './state/designerOptions/designerOptionsSlice';
```
wifun, line 12: this is just for convenience right? currently the host can import from `.../designerOptionsSlice` since it's already exported

ek68794998: Good point, it looks like this is actually available:

```ts
import { initializeServices } from '@microsoft/logic-apps-designer/lib/core/state/designerOptions/designerOptionsSlice';
```

Should we just do that?

shuktika12163: I remember it not getting recognized and there being build errors when I tried using it without this change. Let me try again.

shuktika12163: This is the error I get when I try to import it without the change in LAUX:

Module not found: Can't resolve '@microsoft/logic-apps-designer/lib/core/state/designerOptions/designerOptionsSlice' in 'D:\power-platform-ux\packages\powerautomate-core\src\designer\v3Designer'

shuktika12163: image

hartra344 (approved):

libs/designer/src/lib/core/index.ts
hartra344: It's not for convenience. It tells rollup and our build processes that it needs to be made available when it's bundled.

wifun: Hm ok, so it seems like the typing shouldn't be generated then because the IDE doesn't complain

hartra344: I mean you can import types that way, it just wont work if you try to import js code that way. A lot of libraries export this way.

PR #1878, opened 2023-03-31
[VS Bug] Operations not having workflow kind and onClick getting the first operation

### LINKED ISSUE #1875: Get operation group API no workflowkind API filter

Issue where the workflowKind API filter is not passed to below API. This is causing all the actions used for stateful/stateless workflows to show in both the workflow kinds. And the filtering logic filters some internal operations as well used for dynamic schema

### Workflow JSON

_No response_

### Screenshots or Videos

image

#### Highlighted action should only be visible in the stateless workflows. And other actions like get queues/topics etc are hidden and uses the same filter.
image

### Browser

VSCode and Portal

### Additional context

_No response_

ccastrotrejo:
### Main code changes
- Add workflow kind to search service to show actions according to app kind
- Add apiId to validate the action user clicked to the action that is going to be added to the designer
Closes #1875 

#### First issue - Just listing action according to the kind of app
- Before
Screenshot 2023-03-31 at 3 36 49 PM

- After 
MicrosoftTeams-image (11)

#### Second issue
- Before

https://user-images.githubusercontent.com/102700317/229244976-a2f85b4a-1a60-427f-81da-36d0ce556af8.mov

- After

https://user-images.githubusercontent.com/102700317/229244967-9ab59386-b952-4ede-ba48-fb8e5fb0d6c6.mov


libs/designer/src/lib/ui/Designer.tsx
```diff
@@ -134,7 +134,8 @@ export const Designer = (props: DesignerProps) => {
   const isFetchingInitialData = useIsFetching({
     predicate: (query) => {
       const queryKeysToWatch = ['connections', 'manifest', 'apiWithSwaggers'];
-      return (query.queryKey as string[]).some((val) => queryKeysToWatch.includes(val));
+      const queryKeyArray = query.queryKey && isString(query.queryKey) ? [query.queryKey] : (query.queryKey as string[]);
```
hartra344, line 137: you can remove this, I have another PR fixing the same issue already haha

hartra344 (approved):

libs/designer/src/lib/ui/Designer.tsx
ccastrotrejo: Alright, thanks!!

PR #1882, opened 2023-04-01
lafix: Fix serializer to flatten paths for OpenAPI

ek68794998:
Currently, OpenAPI inputs are serialized in a nested format (see left side of image below). However, OpenAPI expects inputs to be serialized in a flat list (see right side of image below).

image

This change updates the OpenAPI serialization code to flatten the list if it is an OpenAPI serialization.


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -303,9 +308,9 @@ const getOperationInputsToSerialize = (rootState: RootState, operationId: string
   }));
 };
 
-const serializeOperationParameters = (inputs: SerializedParameter[], manifest: OperationManifest): Record<string, any> => {
+const serializeParametersFromManifest = (inputs: SerializedParameter[], manifest: OperationManifest): Record<string, any> => {
   const inputsLocation = (manifest.properties.inputsLocation ?? ['inputs']).slice(1);
-  const inputPathValue = constructInputValues('inputs.$', inputs, false /* encodePathComponents */);
+  const inputPathValue = constructInputValues('inputs.$', inputs, { encodePathComponents: false, flattenPaths: true });
```
preetriti1, line 313: Please do not do it like this, you are assuming all manifest code path should follow this, this is only specific to parameters which have the parameterAlias. Use that property "alias" from serializedParameter in the code logic method "serializeParameter" to determine if you need to flatten it or not. This would fail all existing manifests.

preetriti1 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
ek68794998: @preetriti1 There is no `alias` property on the parameter. Are you referring to `info.alias`?

image

preetriti1: Yes info.alias,

Swagger or manifest can have any property name so I can name a property to have slashes, but it should not be assumed to be open api... please move your logic for aliasing specific, we should not assume and do pattern matching in client code unless there is no other way, this has lead to bugs before.

ek68794998: Fixed. Need to re-run my E2E scenarios but tests are passing.

CHANGELOG.md
```diff
@@ -1,26 +1,24 @@
 # Logic Apps Designer
```
ek68794998, line 1: This file keeps getting updated by the linter when I resolve merge conflicts.

preetriti1 (approved):

PR #1904, opened 2023-04-03
Set designer tools left spacing conditionally based on panel location

jaimereyesmx:
Allow extra spacing for when panel location is to the left


libs/designer/src/lib/ui/styles.less
```diff
@@ -69,6 +69,10 @@
   box-sizing: border-box;
 }
 
+.msla-designer-tools-left-panel {
+  left: 52px !important; // Add extra 32px taken from the collapsed panel
```
hartra344, line 73: Is it possible to do this without !important? I really like to make sure we go through every possible alternative before settling on !important

ek68794998: If it doesn't take by default, changing the selector to `.msla-designer-tools.msla-designer-tools-left-panel` should cause it to overwrite the base `.msla-designer-tools` styles.

jaimereyesmx: Will try that

jaimereyesmx: Fixed it

ek68794998 (approved):

PR #1908, opened 2023-04-04
fix(designer): Fixing the subsequent ListDynamicValues calls by adding correct cache key

bnarayanared:
Fixing the cache-key for getListDynamicValues calls by adding a correct cache key since the parameters were always empty and parameterAlias distinguishes the different ListDynamicValue calls. 
Before: 
image

After: 
image


edgarguoms (approved):

libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -55,7 +55,7 @@ export const getListDynamicValues = async (
       (connectionId ?? '').toLowerCase(),
       connectorId.toLowerCase(),
       operationId.toLowerCase(),
-      getParametersKey(parameters).toLowerCase(),
+      parameterAlias,
```
preetriti1, line 58: You are assuming here every parameter has parameterAlias, only openApi connectors may have parameterAlias, parametersKey should be unique for parameters even if they have an alias. If you are not getting the key then there is a bug somewhere in that code.
Key is supposed to be unique.

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -87,7 +87,7 @@ export async function getDynamicValues(
     return getListDynamicValues(
       connectionReference?.connection.id,
       operationInfo.connectorId,
-      operationInfo.operationId,
+      dynamicState.operationId,
```
preetriti1, line 90: full dynamic state is already passed into getListDynamicValues, look at the last argument, so you should not be sending this operationId, this breaks the contract and the services implementing it. Maybe your service is not using the correct operationId to make the actual dynamic calls, please check la standard dynamic service.

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
bnarayanared: I can revert this as it's not required to make the cache-key unique

libs/designer/src/lib/core/queries/connector.ts
preetriti1: Can you give me an example of such manifest? Point me to your connector manifest and it should be debugged. We have manifests (non open api) which have multiple dynamic values and they do work... parameterKey empty meaning the code which generates parameter key is buggy, your scenario fix should not be here

bnarayanared: image
I'm not sure if this is what you're asking for.

ek68794998: I think the issue here might be arising from where the parameter key is expected to be found. In our `inputsDefinition` block, we do not have a `parameterKey`, and `parameters` is just an empty object:

image

This means that the local scope from `getListDynamicValues` looks like the following, and produces an empty parameter key:

image

It looks like our parameter keys are stored in `nodeInputs`, which I don't think is being considered here:

image

In fact, I can indeed see the parameter keys properly set in the redux state:

image

So it looks like the `parameters` being included here might not be considering those from `nodeInputs` in OpenAPI. Is this something we need to fix in the OpenAPI deserializer?

ek68794998: Digging a bit further, it looks like `getDynamicValues` wants to get the `parameters` value from `definition.extension`, but we don't use that for OpenAPI.

image

image

This might actually be on our side. I'll take a deeper look.

ek68794998: I think I figured this one out.

The way this works is that we use the `dependencies.inputs` object to retrieve `dependencyInfo`. However, this object only contains information about the parameter that it depends on, not the parameter itself.

image

image

This means that any input which has no dependencies will have a blank parameter key. Unless non-OpenAPI manifests have some other way of handling `dependencies.inputs`, I expect all manifests will see the same issue, where input parameter keys are blank.

So this does indeed look like a bug in LAUX. Can we confirm that this exists for manifest operations in LA Portal today?

cc @preetriti1

preetriti1: No, to fetch dynamic content why would you need the parameter key of the parameter on which dynamic call is being made, at that point all you need is connector information, the operationId (the operation making the dynamic call like GetAllTeams) and the dependent parameters.
This is how it works in backend, unless you guys are going something with the parameter key of the parameter on which dynamic calls needs to be made.

bnarayanared: As discussed offline, merging the parameters for dynamicState and inputs to fetch the right parametersKey to make QueryCacheKey distinct

preetriti1 (approved):

PR #1931, opened 2023-04-05
fix(designer): Trigger Spliton issue

maturpi:
Triggers that have spliton option are producing errors due to the outputs not being parsed properly when the spliton is a property that is nested. For example, sharepoint's "When an item is created" triggers' spliton string is `triggerOutputs()?['body/value']`.
This causes an error when checking for property `body/value` in the getUpdatedManifestForSpiltOn function since there actual property is `value` with `body` being the parent property.

image

Previous designer versions allowed us to implement our own OperationManifest function that would convert the outputs using spliton string (see: getSplitOnOutputs). 

The changes in this PR fixes the issue on our (client) side. I had to remove the following code as it was adding 'body' twice when encountering `triggerBody()?['body/value']`. 
```
      if (equals(name, 'triggerBody')) {
        properties.push('body');
      }
```

I couldn't find any tests failing after removing the above code, however I'm very doubtful that it would have no impact on your side. I am unsure how you are currently handling nested spliton strings. Clarity on that subject might help in suggesting a better fix. We might need to go back to having to an interface function for OperationManifest service that allows clients to define their own logic.


maturpi:
Seems like slash notation is expected to remain the same and not be transformed to segments as seen here:
https://github.com/Azure/LogicAppsUX/blob/1aa6e8a50f863cfe0ced2bbe51f91da3a1203666/libs/designer/src/lib/core/utils/parameters/__test__/tokensegment.spec.ts#L226

It does say 'when configured to do so' which is interesting. If we have some type of setting that we can enable as a client that would force LAUX to handle slash notation then this would fix our issue.

While dot notation seems to be converted and parsed correctly:
https://github.com/Azure/LogicAppsUX/blob/1aa6e8a50f863cfe0ced2bbe51f91da3a1203666/libs/designer/src/lib/core/utils/parameters/__test__/tokensegment.spec.ts#L207

This leads me to think that some schema transformation might be happening on the LAUX side to convert the array + slash notation format to dot notation, most likely when loading/fetching the definition/manifest. I cannot find or confirm this behavior however, would appreciate insight.

libs/designer/src/lib/core/utils/__test__/outputs.spec.ts
```diff
@@ -66,15 +67,25 @@ describe('Outputs Utilities', () => {
           },
         },
       };
-      const splitOn = "@triggerOutputs()?['body']";
```
preetriti1, line 69: Can you check if just @triggerBody() in splitON works?

preetriti1 (changes requested):

libs/parsers/src/lib/expression/parser.ts
```diff
@@ -132,11 +139,23 @@ export class ExpressionParser {
...
         token = ExpressionParser._getTokenOrThrowException(scanner, ExpressionTokenType.RightSquareBracket);
-        dereferences.push({
-          isSafe,
-          isDotNotation: false,
-          expression: expression,
-        });
+
+        if (expression.type === ExpressionType.StringLiteral) {
+          // takes care of expressions that are nested such as ['body/value']
+          for (const expressionValue of (expression as ExpressionLiteral).value.split('/')) {
```
preetriti1, line 145: You cannot assume here / means it is openApiConnection aliasing format, what if properties actually have this character, so please add some kind of aliasing support flag in the parser so it knows it has to assume / as open api alias parameter name

libs/designer/src/lib/core/utils/__test__/outputs.spec.ts
maturpi: Instead of removing the code that was pushing 'body' when the name of the string was 'triggerBody', I added a condition to check if the dereferences already contain 'body' at the first index so it avoids converting triggerBody()?['body'] to body.body for example.

`      // prevents duplicate 'body' property when parsing OpenApi manifests
      if (equals(name, 'triggerBody') && dereferences.length && !equals((dereferences[0].expression as ExpressionLiteral).value, 'body')) {
        properties.push('body');
      }`

libs/parsers/src/lib/expression/parser.ts
maturpi: Changed the code to this:
```
if (expression.type === ExpressionType.StringLiteral && scanner.referenceKeyFormat === ConnectionReferenceKeyFormat.OpenApi) {
          // takes care of expressions that are nested such as ['body/value']
          for (const expressionValue of (expression as ExpressionLiteral).value.split('/')) {
            dereferences.push({
              isSafe,
              isDotNotation: false,
              expression: { type: ExpressionType.StringLiteral, value: expressionValue },
            });
          }
```

libs/parsers/src/lib/expression/scanner.ts
```diff
@@ -5,19 +5,22 @@ import { ScannerException } from '../common/exceptions/scanner';
...
 import type { ExpressionToken } from '../models/expression';
 import { ExpressionTokenType } from '../models/expression';
+import type { ConnectionReferenceKeyFormat } from '@microsoft/utils-logic-apps';
 import { equals } from '@microsoft/utils-logic-apps';
 
 export class ExpressionScanner {
   private _expression: string;
   private _startPosition: number;
   private _currentToken: ExpressionToken;
+  private _referenceKeyFormat: ConnectionReferenceKeyFormat | undefined;
```
preetriti1, line 15: this is a string parser/scanner we cannot introduce product / connector specific thing in this class that way this is no longer a library which can be used outside the context of LA or PA... I am not seeing where you are using this as class property.

preetriti1 (changes requested):

libs/parsers/src/lib/expression/scanner.ts
maturpi: I'm using the reference in the parser file on line 143. This was introduced at your request to identify if it is openapi:
> You cannot assume here / means it is openApiConnection aliasing format, what if properties actually have this character, so please add some kind of aliasing support flag in the parser so it knows it has to assume / as open api alias parameter name

```       
 if (expression.type === ExpressionType.StringLiteral && scanner.referenceKeyFormat === ConnectionReferenceKeyFormat.OpenApi) {
          // takes care of expressions that are nested such as ['body/value']
          for (const expressionValue of (expression as ExpressionLiteral).value.split('/')) {
```

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -91,12 +91,14 @@ export const getUpdatedManifestForSpiltOn = (manifest: OperationManifest, splitO
...
-    const parsedValue = ExpressionParser.parseTemplateExpression(splitOn);
+    const parsedValue = ExpressionParser.parseTemplateExpression(splitOn, manifest.properties.connectionReference?.referenceKeyFormat);
     const properties: string[] = [];
     let manifestSection = updatedManifest.properties.outputs;
     if (isSupportedSplitOnExpression(parsedValue)) {
       const { dereferences, name } = parsedValue as ExpressionFunction;
-      if (equals(name, 'triggerBody')) {
+
+      // prevents duplicate 'body' property when parsing OpenApi manifests
+      if (equals(name, 'triggerBody') && dereferences.length && !equals((dereferences[0].expression as ExpressionLiteral).value, 'body')) {
```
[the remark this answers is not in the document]
maturpi: On PowerAutomate side it comes out to triggerBody()['body/value']. Like I mentioned at the beginning of this merge request, there's most likely some operation manifest manipulation/transformation being done on your side to make this work. This is most likely why the designer also used to have a function for getSplitOnOutputs in OperationManifest that was implemented by clients.

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -861,7 +861,11 @@ export function generateExpressionFromKey(
   let rootMethod = method;
   if (!isInsideArray && segments[0]?.value?.toString()?.toLowerCase() === OutputSource.Body) {
     segments.shift();
-    rootMethod = actionName ? `${OutputSource.Body}(${convertToStringLiteral(actionName)})` : constants.TRIGGER_BODY_OUTPUT;
+    rootMethod = actionName
+      ? `${OutputSource.Body}(${convertToStringLiteral(actionName)})`
+      : method === constants.TRIGGER_OUTPUTS_OUTPUT
+      ? method
+      : constants.TRIGGER_BODY_OUTPUT;
```
shuktika12163, line 868: We realized this part was causing issues on our end because it was converting things of the format "triggerOutputs()[body/value]" to "triggerBody()[body/value]", making it have double "body" since the result below doesn't remove "body" from the segments. If the method is triggerOutputs, we would like to keep it as trigger outputs. We would like someone from LAUX to confirm this doesn't break spliton triggers or anything else on LogicApps side.

shuktika12163: Method comes from Redux state which has triggerOutputs in it, but gets converted to triggerBody which is wrong:
image

shuktika12163: After making this change, we are able to see the option correctly as "@triggerOutputs()?['body/value']"

libs/parsers/src/lib/expression/scanner.ts
shuktika12163: Updated to a generic bool to signify this instead

libs/parsers/src/lib/expression/parser.ts
```diff
@@ -43,14 +50,14 @@ export class ExpressionParser {
     },
   ];
 
-  public static parseExpression(expression: string): Expression {
-    const scanner = new ExpressionScanner(expression);
-    const parsedExpression = ExpressionParser._parseExpressionRecursively(scanner);
+  public static parseExpression(expression: string, isAliasPathParsingEnabled?: boolean): Expression {
+    const scanner = new ExpressionScanner(expression, /*prefetch*/ true);
```
preetriti1, line 54: You do not need to set prefetch here explicitly since it is already a default set value to true in the constructor

libs/parsers/src/lib/expression/parser.ts
```diff
@@ -62,14 +69,14 @@ export class ExpressionParser {
...
-        return ExpressionParser.parseExpression(expression.substring(1));
+        return ExpressionParser.parseExpression(expression.substring(1), isAliasPathParsingEnabled);
       }
     } else {
       return ExpressionParser._parseStringInterpolationExpression(expression);
     }
   }
 
-  private static _parseExpressionRecursively(scanner: ExpressionScanner, index = 0): Expression {
+  private static _parseExpressionRecursively(scanner: ExpressionScanner, index = 0, isAliasPathParsingEnabled?: boolean): Expression {
```
preetriti1, line 79: Better to keep this argument isAliasPathParsingEnbaled as required in all private functions so we can make sure it is getting passed down from all callers, else some code paths can be missed.

libs/parsers/src/lib/expression/parser.ts
```diff
@@ -132,11 +139,23 @@ export class ExpressionParser {
       if (scanner.getTokenForTypeAndValue(ExpressionTokenType.LeftSquareBracket)) {
         const expression = this._parseExpressionRecursively(scanner);
         token = ExpressionParser._getTokenOrThrowException(scanner, ExpressionTokenType.RightSquareBracket);
-        dereferences.push({
-          isSafe,
-          isDotNotation: false,
-          expression: expression,
-        });
+
+        if (expression.type === ExpressionType.StringLiteral && isAliasPathParsingEnabled) {
```
preetriti1, line 143: Are we sure we do not want to do it for ExpressionType stringinterpolation?

libs/designer/src/lib/core/utils/parameters/helper.ts
preetriti1: segments[0]?.value?.toString()?.toLowerCase() is validating if first segment is body, so it first segment is body its method is converting into triggerBody(), in your case the first segment shouldn't be coming as body as your segment should be body/value.
Can you point me which caller is calling this generateExpreesionfromKey method?

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/helper.ts
preetriti1: I see the caller now, in your case we should just send aliasing attribute to skip this check, because if we do this for all LA might break;

shuktika12163: Ahh, just realized it is the OpenApi causing duplicate/repetitive paths for the parameters. Eric (on our team) solved a few of these in other areas. We'll do something similar to that and have it behind the parameterAlias flag. Thank you!

For more context (from Eric):
Each section of the key has the previous one prefixed onto it. It can get really cumbersome:
body.body/value.body/value/foo.body/value/foo/bar
which is really just body -> value -> foo -> bar

Some sample PRs from before:
https://github.com/Azure/LogicAppsUX/pull/1756
https://github.com/Azure/LogicAppsUX/pull/1762
https://github.com/Azure/LogicAppsUX/pull/1725

libs/parsers/src/lib/expression/parser.ts
maturpi: I'm not sure for this, I'll leave a comment in case we run into issues that indicate string Interpolation should be inlcuded

libs/parsers/src/lib/expression/parser.ts
```diff
@@ -43,14 +50,14 @@ export class ExpressionParser {
     },
   ];
 
-  public static parseExpression(expression: string): Expression {
+  public static parseExpression(expression: string, isAliasPathParsingEnabled: boolean): Expression {
```
preetriti1, line 53: There is too many updates, can we make this value default as false, so only true values are being passed from callers? Similar for below method 'parseTemplateExpression'

We just need to make sure all private methods have this as required parameter.

preetriti1 (changes requested):

libs/parsers/src/lib/expression/parser.ts
maturpi: @preetriti1 makes sense, addressed this in latest commit.

preetriti1 (approved):

PR #1949, opened 2023-04-07
feat(powerautomate): Add UI component for floating action menu

Ram-G:
Bringing over floating action menu from BPMUX repo. This component is needed by Power Automate.

Collapsed:
image

Expanded:
image

How tested:
In Power Automate Portal (power-platform-ux repo) by loading a flow that contains the manual button trigger

In follow-up PRs:
- Bring over DynamicallyAddedParameter ([WI 17890957](https://msazure.visualstudio.com/One/_workitems/edit/17890957))


libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+      e.preventDefault();
+      e.stopPropagation();
+      toggleExpanded();
+    }
+  };
+
+  const renderMenuButton = (): JSX.Element => {
+    return (
+      <div role="button" className="msla-floating-action-menu" onClick={toggleExpanded} onKeyDown={toggleExpandedOnKeyDown} tabIndex={0}>
+        <span className="msla-floating-action-menu-plus-icon">{'+'}</span>
```
Ram-G, line 45: should there be double quotes around the +? `"+"`?

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }
+  };
+
+  const handleMenuItemSelected = (item: FloatingActionMenuItem) => {
+    toggleExpanded();
+    props.onMenuItemSelected(item);
+  };
+
+  const handleMenuItemSelectedOnKeyDown = (e: React.KeyboardEvent<HTMLElement>, item: FloatingActionMenuItem) => {
+    const { keyCode } = e;
```
Ram-G, line 126: VS Code told me `keyCode` is deprecated. Based on a quick search, we can use `e.code` instead. What is the  new recommended new way to identify keyboard events by keypress?

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -212,6 +214,21 @@ const TokenField = ({
     case 'recurrence':
       return <ScheduleEditor readOnly={readOnly} type={editorOptions?.recurrenceType} initialValue={value} onChange={onValueChange} />;
 
+    case 'floatingactionmenu': {
+      const menuItems = getMenuItemsForDynamicAddedParameters(editorOptions?.supportedTypes);
```
Ram-G, line 218: this function call although reasonable breaks the trend of simply returning JSX components in this `switch case`. Should I structure this code differently?

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/logicappsux",
-  "version": "2.0.24",
+  "version": "2.0.25",
```
Ram-G, line 3: Is this package version update automatic? I tried to recreate the PR without it, but every time I do `git commit -m "..."`, this change gets added.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -0,0 +1,148 @@
+import { KeyCodes } from '@fluentui/react';
+import { useBoolean } from '@fluentui/react-hooks';
+import React from 'react';
+import { useIntl } from 'react-intl';
+
+export enum DynamicallyAddedParameterType {
```
Ram-G, line 6: Perhaps we'll move this enum into a new file for `DynamicallyAddedParameter` once that component gets added? Put it here for now

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
hartra344: no single quotes unless the string has a single quote inside of it. Prettier will always enforce this.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
hartra344: Please don't use typescript enums. It'll break the babel typescript compiler used to build the production version. Use key/value constants instead.

libs/designer-ui/src/lib/floatingactionmenu/helper.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const supportedTypesArray = supportedTypes as Array<string>;
+
+  const menuItems = [];
+  if (supportedTypesArray.includes('text')) {
+    menuItems.push({
+      label: 'Text',
+      id: DynamicallyAddedParameterType.Text,
+      icon: '', // TODO: Add icon
```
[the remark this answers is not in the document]
hartra344: I'm assuming they're svgs and you'll need to copy them over and import them. Look for other svgs in the code as an example.

libs/designer-ui/src/lib/floatingactionmenu/helper.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!supportedTypes) {
+    throw new ValidationException(ValidationErrorCode.INVALID_PARAMETERS, 'supportedTypes are necessary.');
+  }
+
+  const supportedTypesArray = supportedTypes as Array<string>;
+
+  const menuItems = [];
+  if (supportedTypesArray.includes('text')) {
+    menuItems.push({
+      label: 'Text',
```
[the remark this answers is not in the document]
hartra344: One benefit of using react-intl is how easy it is to include the localizations in the code. the tooling does all of the work to extract it and put it in a resjson format. I would suggest getting in the habit of always localizing during development to avoid any accidental misses.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
hartra344: I'm not sure, the go to seems to just be using string representations of the keys now. I don't particularly care for it either due to lack of type safety but should be fine as long as it's tested.

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
hartra344: Yes please, this style makes it difficult for the react runtime to do proper optimizaitons.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: Prettier puts these to single quotes, going with that

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: changed to be enum strings instead of number, as this is how other files in designer-ui seem to be doing it

package-lock.json
Ram-G: leaving this as is for now, I expect will run into merge conflict at some point, will find out what to do then

libs/designer-ui/src/lib/floatingactionmenu/helper.ts
```diff
@@ -0,0 +1,59 @@
...
...AgIDxnIGlkPSJUZXh0X2ljb24iIHRyYW5zZm9ybT0idHJhbnNsYXRlKDEyLjAwMDAwMCwgOTMuMDAwMDAwKSI+CiAgICAgICAgICAgICAgICAgICAgPGcgaWQ9InRleHQiPgogICAgICAgICAgICAgICAgICAgICAgICA8Y2lyY2xlIGlkPSJPdmFsLTU1NC1Db3B5IiBmaWxsPSIjOUY2Q0Q5IiBjeD0iMjIiIGN5PSIyMiIgcj0iMjIiPjwvY2lyY2xlPgogICAgICAgICAgICAgICAgICAgICAgICA8cGF0aCBkPSJNMjUuMjAwNjYxMywxMy43NSBMMjAuMzI1NTA2MywyOC4zNDU2MjkzIEwxNy41ODgyOTMsMjAuMTQ0MjE4OCBMMTYuMjQ5OTQ4MywyMC4xNDQyMTg4IEwxMi44MzMzMzMzLDMwLjM3Mjc1MjYgTDE0LjE4MTkwNzQsMzAuMzcyNzUyNiBMMTUuMDMxNzk4OSwyNy44MTYyNTg1IEwxOC44MDY0NDI0LDI3LjgxNjI1ODUgTDE5LjY1NjMzNCwzMC4zNzI3NTI2IEwyMS4wMDQ5MDgsMzAuMzcyNzUyNiBMMjIuNzAzODM4NiwyNS4yNTgwNTk1IEwyOS4wMzY2ODExLDI1LjI1ODA1OTUgTDMwLjczNDc1OTIsMzAuMzcyNzUyNiBMMzIuMDgzMzMzMywzMC4zNzI3NTI2IEwyNi41Mzk4NTg0LDEzLjc1IEwyNS4yMDA2NjEzLDEzLjc1IFogTTI1Ljg2OTgzMzYsMTUuNzc4ODI4MyBMMjguNjA3MDQ2OSwyMy45ODAyMzg3IEwyMy4xMzI2MjA0LDIzLjk4MDIzODcgTDI1Ljg2OTgzMzYsMTUuNzc4ODI4MyBaIE0xNi45MTkxMjA3LDIyLjE3MTM0MjIgTDE4LjM3NzY2MDcsMjYuNTM2NzMyOCBMMTUuNDYwNTgwNywyNi41MzY3MzI4IEwxNi45MTkxMjA3LDIyLjE3MTM0MjIgWiIgaWQ9IlBhZ2UtMSIgZmlsbD0iI0ZGRkZGRiI+PC9wYXRoPgogICAgICAgICAgICAgICAgICAgIDwvZz4KICAgICAgICAgICAgICAgIDwvZz4KICAgICAgICAgICAgPC9nPgogICAgICAgIDwvZz4KICAgIDwvZz4KPC9zdmc+',
```
Ram-G, line 14: These are generated strings copied from BPMUX. I'm not thrilled about having these as strings, but the code seems to be sensitive to how the svg string is generated, and doing the .svg -> string conversion in this file in this repo wasn't having the icons render correctly. So copied the generated strings from BPMUX

Ram-G:
Currently investigating why if I use the two arrows at the top-right of the left panel to collapse it, the flow complains of validation errors 

image

hartra344:
> Currently investigating why if I use the two arrows at the top-right of the left panel to collapse it, the flow complains of validation errors
> 
> image

It's complaining because you have a parameter marked as required in the swagger but aren't providing a value for it. It runs validation when the parameter panel is closed so users can find where they have errors.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
hartra344: I mean don't use the enum keyword at all. It does not exist in JS and it generates code. It should not be used in the codebase. Here is an article on valid alternatives: https://maxheiber.medium.com/alternatives-to-typescript-enums-50e4c16600b1

hartra344 (changes requested):

preetriti1:
Can you please have your button trigger manifest here, so we can compare?

Ram-G:
Here is a link to the code where we add inputs for button trigger manifest: https://msazure.visualstudio.com/One/_git/Flow-RP?path=/src/processsimple/Roles/ProcessSimple.Data/DataProviders/BuiltInOperationDataProvider.cs&version=GBmaster&line=1961&lineEnd=1964&lineStartColumn=1&lineEndColumn=115&lineStyle=plain&_a=contents

Could you point me to the v1 manifest and/or help understand what's different here? I wonder if this will also help me with the required paramater error I posted a screenshot above; I wonder if that error is a difference in manifest 

CC @preetriti1, thank you

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: Thank you for providing the pointer to the article. I see now.

Removed enum in latest iteration

Ram-G:
Planning take the fix for the parameter error in follow-ups -- it's something we are working parallelly on, to figure out the story for manifest for manutal trigger. Good news is that this UI component is unused in Logic Apps

Requesting re-review. Thank you

hartra344 (approved):

PR #1990, opened 2023-04-12
fix(designer): Add UI for invoker connection support

bnarayanared:
The UI changes required to add the "Use Invoker's connection" toggle. 
How this looked in v1:
image
Changes to v3:
image

A few things to keep in mind:
This setting needs to be shown under general tab only for Dataverse actions if the trigger is a Dataverse CUD trigger.


hartra344:
@bnarayanared can you also please include a description of what is getting changed here? Images also help.

bnarayanared:
> @bnarayanared can you also please include a description of what is getting changed here? Images also help.

Sorry forgot that bit, added the UI screenshots for v1 v/s v3

libs/designer/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -307,3 +309,14 @@ const getOperationType = (operation: DiscoveryOperation<DiscoveryResultTypes>):
       : Constants.NODE.TYPE.API_CONNECTION
     : operationType;
 };
+
+export const getRootNodeDetails = (workflowState: WorkflowState, connectorId: string): string | undefined => {
+  if (workflowState.graph?.children && connectorId.indexOf(Constants.INVOKER_CONNECTION.DATAVERSE_CONNECTOR_ID) > -1) {
```
hartra344, line 314: This is super specific to a particular connector, something we want to avoid. Is there a way we can generalize this behavior. Also can you explain why it needs this special treatment at all?

libs/designer/src/lib/ui/settings/index.tsx
```diff
@@ -234,15 +234,31 @@ function GeneralSettings({ nodeId, readOnly }: { nodeId: string; readOnly?: bool
...
   };
 
+  const onInvokerConnectionToggle = (checked: boolean): void => {
+    dispatch(
+      updateNodeSettings({
+        id: nodeId,
+        settings: {
+          invokerConnection: {
+            isSupported: !!invokerConnection?.isSupported,
+            value: { value: invokerConnection?.value?.value ?? undefined, enabled: checked },
```
hartra344, line 244:
```suggestion
            value: { value: invokerConnection?.value?.value, enabled: checked },
```

This means the same thing

libs/designer/src/lib/ui/settings/sections/general.tsx
```diff
@@ -104,12 +107,22 @@ export const General = ({
     defaultMessage: 'Array',
     description: 'title for array dropdown input setting',
   });
+  const invokerConnectionTitle = intl.formatMessage({
+    defaultMessage:'Use Invoker\'s Connection',
```
hartra344, line 111:
```suggestion
    defaultMessage:"Use Invoker's Connection",
```

always prefer double quotes if the string includes a single quote. This helps avoid errors/bugs related to escapes.

hartra344 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/add.ts
bnarayanared: The invoker connection is very specific to Dataverse actions. It should only be shown as a setting on Dataverse actions if the trigger being used is a Dataverse CUD trigger and so that's why it needs the specific checks. I'm open to taking suggestions if there's a way to generalize this but afaik this is very specific to DV actions.

hartra344 (approved):

PR #2063, opened 2023-04-19
fix(designer): Add invoker connection support on toggle

bnarayanared:
Adding the invoker connection support for DV actions on CUD trigger:
image

image


libs/designer/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -203,7 +203,22 @@ const initializeOperationDetails = async (
   // Re-update settings after we have valid operation data
   const rootNodeId = getRootNodeDetails(state.workflow, connectorId);
   const operation = getState().workflow.operations[nodeId];
-  const settings = getOperationSettings(isTrigger, operationInfo, initData.nodeOutputs, manifest, swagger, operation, rootNodeId);
+
+  const connectionReferences = state.connections.connectionReferences;
+  if (state.connections.connectionReferences !== undefined) {
```
ek68794998, line 208: No need for the `!== undefined` here, can just do `if (!state.connections.connectionReferences) {`.

libs/designer/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -203,7 +203,22 @@ const initializeOperationDetails = async (
...
+    dispatch(addInvokerSupport({ connectionReferences }));
```
ek68794998, line 209: This dispatch may not complete before the call to `getOperationSettings` below. Is that intended?

libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
```diff
@@ -136,7 +136,9 @@ export const initializeOperationDetailsForManifest = async (
   nodeId: string,
   _operation: LogicAppsV2.ActionDefinition | LogicAppsV2.TriggerDefinition,
   isTrigger: boolean,
-  dispatch: Dispatch
+  dispatch: Dispatch,
```
ek68794998, line 139: I'm a little concerned about passing `dispatch` around like this. It's generally intended to be used within the scope of a component so that other functions which might not have access to the app dispatch can still use it.

Would like input from LAUX team on how to best handle this. I think a callback argument might be the best option.

libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
```diff
@@ -190,7 +195,8 @@ export const getOperationSettings = (
     },
     invokerConnection: {
       isSupported: isInvokerConnectionSupported(isTrigger, nodeType, rootNodeId),
-      // value: TODO in next PR
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
ek68794998, line 198: Shouldn't need to do this. `invokerConnection` should allow `connectionReferences: ConnectionReferences | undefined`, then the `!` can be removed here.

libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
bnarayanared: This `dispatch` was already present in this method. I did not add it myself.

hartra344 (approved):

hartra344 (commented):
Please make the changes @ek68794998 suggested and I'll merge after.

Issue #2047, opened 2023-04-17
Dynamic data inputs do not populate for existing values since 2.4.2


ek68794998:
### Describe the Bug with repro steps

1. Open or create a Logic App.
2. Add a new action that has dynamically populated dropdown inputs. (I used a SharePoint action.)
3. Select an input from a dropdown.
4. Close and reopen the parameters pane.

### Screenshots or Videos

This is what it should look like (and what it looks like on 2.4.1):

![image](https://user-images.githubusercontent.com/1350074/232616293-6c403dc6-a7da-445e-8780-73512f662e6f.png)

And this is what it looks like (on 2.4.2) after closing and reopening the pane (or loading an existing Logic App with prepopuated values):

![image](https://user-images.githubusercontent.com/1350074/232616369-76e92868-69bb-4baa-9402-b4060043d082.png)

### Browser

Chrome 112.0.5615.121, Windows 11 x64.

ek68794998:
I expect this was caused by changes to how enum dropdowns work in: #2032

Eric-B-Wu:
Discussed offline, but this happens when clicking the node before the dynamic data has populated, and was happening before 2.4.2

Eric-B-Wu:
to be honest, I'm not sure why it was working before, it might be because we reduced the initial load time so that nodes were clickable before everything had loaded, but either way there was a callback being called when loading was completing that shouldn't have been which was the root of the issue.

PR #2067, opened 2023-04-19
fix(designer): combobox showing UIUD when clicking on a loading dynamic data parameter

Eric-B-Wu:
Should now Update accordingly

combox


hartra344 (approved):

ek68794998:
Validated locally that this fixes both #2047 and the pre-existing issue of "UUIDs are shown while workflow is loading". Thank you Eric!

PR #2068, opened 2023-04-19
feat(powerautomate): Implement DynamicallyAddedParameter to support manual trigger

Ram-G:
Adding UI component DynamicallyAddedParameter and serialization/deserialization of it. This enables the following scenarios for dynamic parameters in a manual trigger
- Edit, add, delete parameters in new & existing flows
- Save & load flows in v3

Scenarios that may run into problems:
- Save flow in v3 and load flow in v1

TODOs (in follow-up PRs):
- Dynamic parameters don't show up in token picker
- Naming of parameters is different from v1 designer (BPMUX repo)

GIF:
GIF 4-20-2023 1-19-03 PM


libs/designer-ui/src/lib/dynamicallyaddedparameter/dynamicallyaddedparameter.less
```diff
@@ -0,0 +1,102 @@
+.msla-dynamic-added-param-header {
```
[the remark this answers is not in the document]
hartra344: It doesn't unless you import it here: https://github.com/Azure/LogicAppsUX/blob/main/libs/designer-ui/src/lib/styles.less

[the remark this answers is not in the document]
Ram-G: thanks! Imported it in styles.less for now. I haven't yet verified that the styling is what we want though, because some of the elements have changed and I copied the styles from BPMUX repo

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -132,9 +156,42 @@ export const FloatingActionMenu = (props: FloatingActionMenuProps): JSX.Element
...
+    const { icon, type, placeholder } = item;
+    const format =
+      type === DynamicallyAddedParameterType.Date || type === DynamicallyAddedParameterType.Email ? type.toLowerCase() : undefined;
+    dynamicParameterProps.push({
+      icon,
+      schemaKey: guid(), // TODO: generate schemaKey
+      properties: {
+        description: placeholder,
+        format,
+        title: enterVariableNamePlaceholder, // TODO: generate title based on schemaKey
```
Ram-G, line 174: The 1) property key here (called `schemaKey` in this code) and 2) `title` are named/generated differently from v1 designer (BPMUX repo).

In BPMUX, instead of a guid the keys would be named -- text, text_1, text_2, etc.
and the `title` would be generated based on those, for example -- "Text", "Text 1", "Text 2", etc.

Ram-G: Questions:
1) Is this functionality we can do without for now?
2) What repercussions does this have if we want to load a flow saved in v3 designer in v1 designer?

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -132,9 +156,42 @@ export const FloatingActionMenu = (props: FloatingActionMenuProps): JSX.Element
...
+        type,
+        'x-ms-content-hint': type,
+        'x-ms-dynamically-added': true,
+      },
+      required: true, // TODO: add functionality to allow making parameters optional
```
Ram-G, line 179: In v1 designer, this option looks like this in the menu -- 

image

libs/designer-ui/src/lib/dynamicallyaddedparameter/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    );
+  };
+
+  return (
+    <div className="msla-dynamic-added-param-container">
+      {renderDynamicParameterContainer()}
+      <div className="msla-dynamic-added-param-bottom-divider" />
+    </div>
+  );
+};
```
Ram-G, line 138: Error when loading flow in v1 designer below. This is because v1 expects certain properties to be structured in a way that is missing in v3.

TODO: figure out importance of this
image

libs/designer-ui/src/lib/dynamicallyaddedparameter/dynamicallyaddedparameter.less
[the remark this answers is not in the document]
Ram-G: update: going through all the styles in this file and making sure they're good

libs/designer-ui/src/lib/dynamicallyaddedparameter/index.tsx
Ram-G: This should not be blocking signoff for this PR. I will take as an action item to talk about it

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: Right now, this is not blocking -- we're in a time crunch so would like to see this PR merged.

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: Not blocking right now.

hartra344 (approved):

PR #2070, opened 2023-04-19
fix: add logic to fallback to connector data from Operation Manifest when Connection Service data is undefined

jaimereyesmx:
Reintroducing logic that uses Operation Manifest connector data that was removed as part of https://github.com/Azure/LogicAppsUX/commit/22d6c30d7977a9fb7ef17e10095084bf9e77c1aa#diff-ea2d1be19c1bdc8a7105f83da7156578f002d4dd84fcf9cc702d6c925d280bf9. The main change is that useConnectorByNodeId will give priority to Connection Service and then fallback to Operation manifest to avoid reintroducing a regression.

### DIFF
--- a/libs/designer/src/lib/core/state/connection/connectionSelector.ts
+++ b/libs/designer/src/lib/core/state/connection/connectionSelector.ts
@@ -1,5 +1,6 @@
 import type { ConnectionReference, ConnectionReferences } from '../../../common/models/workflow';
 import type { RootState } from '../../store';
+import { useOperationManifest, useOperationInfo } from '../selectors/actionMetadataSelector';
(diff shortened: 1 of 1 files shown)


hartra344:
@rllyy97 is there a reason in particular that this was removed that we should keep in mind?

rllyy97 (approved):

rllyy97:
@hartra344 Yes, it was removed due to one of the two connectors (previously the one with priority) having some connection data missing.  We didn't need a fallback in our implementation since they were otherwise identical, but PA needs the fallback, so having the same logic but just reversed priority solves both issues (I verified this with Jaime yesterday)

shuktika12163 (approved):

PR #2083, opened 2023-04-20
fix(designer): ensure all operations and connectors pages are fetched on fisrt search

### LINKED ISSUE #2081: Operations and Connectors search does not fetch all the pages

When adding an action for first time all Operations and Connectors should be fetched, if there is another page to be fetched a subsequent call needs to be made to fetch that page and continue until there are no more pages to fetch.

After https://github.com/Azure/LogicAppsUX/pull/1983 this behavior regressed and only one page is fetched resulting in three network calls (1 for Azure, 1 for Custom and 1 for BuiltIn), and subsequent pages are not fetched even though there is a next link.

Current behavior:
image

Expected behavior:
image

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

All browsers

### Additional context

_No response_

jaimereyesmx:
When adding an action for first time all Operations and Connectors should be fetched, if there is another page to be fetched a subsequent call needs to be made to fetch that page and continue until there are no more pages to fetch.

After https://github.com/Azure/LogicAppsUX/pull/1983 this behavior regressed and only one page is fetched resulting in three network calls (1 for Azure, 1 for Custom and 1 for BuiltIn), and subsequent pages are not fetched even though there is a next link. This change fixes that regression.

Current behavior:
image

New behavior:
image

Fixes #2081

### DIFF
--- a/libs/designer/src/lib/core/queries/browse.ts
+++ b/libs/designer/src/lib/core/queries/browse.ts
@@ -23,8 +23,32 @@ const pagedOpts = {
 /// Operations ///
 
 export const useAllOperations = () => {
-  const { data: azureOperations, isLoading: azureLoading, hasNextPage: azureHasNextPage } = useAzureOperationsLazyQuery();
-  const { data: customOperations, isLoading: customLoading, hasNextPage: customHasNextPage } = useCustomOperationsLazyQuery();
+  const {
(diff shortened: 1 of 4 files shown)


ek68794998:
LGTM from our side as well. Thanks!

PR #2095, opened 2023-04-20
feat(state): Implement basis for workflow state `isDirty`

ek68794998:
This PR adds an `isDirty` property to the `workflow` state and updates it when node parameters or connections are updated, or when nodes are added, deleted, or moved within the workflow. There is also a slice method to dispatch a change to manually set the dirty state, so when the workflow is saved, this can be invoked.

**Note:** `isDirty` is set to `true` if a parameter is clicked into and out of, even though the value is not changed. (This is due to `validationErrors` being set from `undefined` to `[]`.)

image


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1730,6 +1731,7 @@ export async function loadDynamicValuesForParameter(
     if (isDynamicDataReadyToLoad(dependencyInfo)) {
       dispatch(
         updateNodeParameters({
+          isUserAction: false,
```
hartra344, line 1734: can we actually just make this optional and treat undefined as false?

hartra344 (changes requested):

libs/designer/src/lib/core/utils/parameters/helper.ts
ek68794998: Sure. I had gone the route of being explicit but I have no actual preference. Will make that change.

hartra344 (approved):

PR #2116, opened 2023-04-24
fix(designer): Fix incorrect type when creating new DynamicallyAddedParameter to unblock Save Flow

Ram-G:
As part of the implementation of DynamicallyAddedParameter, we are passing the incorrect 'type' for its properties. The correct type is a swagger data type; instead we're passing in the 'floating action menu type'. This PR fixes that.

We now mimic the logic from V1 designer [operationoutputs.ts](https://msazure.visualstudio.com/One/_git/AzureUX-BPMUX?path=/src/core/stores/operationoutputs.ts&version=GBmaster&line=344&lineEnd=369&lineStartColumn=1&lineEndColumn=22&lineStyle=plain&_a=contents).

GIF 4-24-2023 2-11-25 PM

### DIFF
--- a/libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
+++ b/libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
@@ -2,6 +2,7 @@ import type { DynamicallyAddedParameterProps, DynamicallyAddedParameterTypeType,
 import { DynamicallyAddedParameterType } from '.';
 import type { ValueSegment } from '../editor';
 import { ValueSegmentType } from '../editor';
+import { getIntl } from '@microsoft/intl-logic-apps';
 import { guid } from '@microsoft/utils-logic-apps';
 
 export type DynamicallyAddedParameterIcon = string;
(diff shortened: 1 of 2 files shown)


libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -117,3 +118,50 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
     },
   ];
 }
+
+export function createDynamicallyAddedParameterProperties(
+  itemType: DynamicallyAddedParameterTypeType,
+  description: string
+): IDynamicallyAddedParameterProperties {
+  let format, fileProperties;
+  let type = '';
```
shuktika12163, line 127: Would be nice to have swagger type enum for list of possible values

Ram-G: on it

preetriti1 (approved):
You have multiple components added with same naming convention as old designer. We should have these under one editor folder, this should be refactored a bit later, changing name of your components so that when we create editor services it's all in one place and readable.

Ram-G:
Created this backlog item for us to track cleanup as per Priti's comment above: https://msazure.visualstudio.com/One/_workitems/edit/19567382

PR #2118, opened 2023-04-24
feat: Add prop to specify whether to display Runtime dropdown and labels

jaimereyesmx:
Allow the flexibility to specify whether Runtime dropdown and labels should be visible.

After this change the designer can be customized to not include runtime dropdown and labels.

"Add an action" panel
image

Search results
image

Triggers
image

### DIFF
--- a/libs/designer-ui/src/lib/panel/recommendationpanel/browseResults.tsx
+++ b/libs/designer-ui/src/lib/panel/recommendationpanel/browseResults.tsx
@@ -48,7 +48,7 @@ export const BrowseGrid = (props: BrowseGridProps) => {
         </div>
       );
     },
-    [forceSingleCol, onConnectorSelected]
+    [forceSingleCol, onConnectorSelected, displayRuntimeInfo]
   );
 
   const noResultsText = intl.formatMessage({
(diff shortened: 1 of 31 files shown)


libs/designer-ui/src/lib/panel/recommendationpanel/browseResults.tsx
```diff
@@ -41,6 +42,7 @@ export const BrowseGrid = (props: BrowseGridProps) => {
               brandColor={brandColor}
               onClick={onConnectorSelected}
               category={getConnectorCategoryString(connector.id)}
+              displayRuntimeInfo={displayRuntimeInfo}
```
ccastrotrejo, line 45: Missing dependency in react hook

jaimereyesmx: Nice catch, thanks!

jaimereyesmx: Resolved it.

ccastrotrejo:
Seems good to me. It seems that the default value for `displayRuntimeInfo` is true, however my concerns, and maybe I am not aware of this or didn't catch it in the PR, is in which scenarios when want it to be false and if we are going to implement that "switch" in vscode and portal  @hartra344

ccastrotrejo (approved):

hartra344 (approved):

PR #2125, opened 2023-04-25
fix(designer): Don't validate parameters and dispatch on blur if the user is still editing on the panel

almaasrawi:
Don't validate parameters and dispatch on blur if the user is still editing on the panel. This prevents the issue where the token picker for drop down list custom value would immediately get dismissed on click.

image

### DIFF
--- a/libs/designer-ui/src/lib/tokenpicker/index.tsx
+++ b/libs/designer-ui/src/lib/tokenpicker/index.tsx
@@ -38,6 +38,8 @@ const calloutStyles: Partial<ICalloutContentStyles> = {
 
 export type SearchTextChangedEventHandler = (e: string) => void;
 
+const tokenPickerClassName = 'msla-token-picker';
+
 export interface TokenPickerProps {
   editorId: string;
   labelId: string;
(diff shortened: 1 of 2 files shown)


libs/designer-ui/src/lib/tokenpicker/index.tsx
```diff
@@ -264,3 +266,12 @@ export function getWindowDimensions() {
     height,
   };
 }
+
+export function isTokenPickerElement(element: HTMLElement | null) {
+  for (let e = element; e !== null; e = e.parentElement) {
+    if (e.className.includes(tokenPickerClassName)) {
```
Eric-B-Wu, line 272: can we change this to:
`
if (e.className?.includes(tokenPickerClassName)) {
`
it will crash for elements without classNames

almaasrawi: Thanks for the catch !

hartra344:
We've actually decided to revert this initial PR. we found some more bugs that seemed ot be caused by it and it would be better t find another way of doing this validation trigger than to continue batting at hte issues that pop up.

PR #2127, opened 2023-04-25
fix(designer): Change V3 token picker classname diff from V1

yijyuan:
Need to distinguish the V3 token picker classname from V1 classname.
When we are support both V1 and V3, we are seeing [V1 styling](https://msazure.visualstudio.com/One/_git/AzureUX-BPMUX?path=/src/ui/tokenpicker/tokenpicker.less&version=GBmaster&line=55&lineEnd=58&lineStartColumn=1&lineEndColumn=32&lineStyle=plain&_a=contents) getting applied to V3 token picker.
image

### DIFF
--- a/libs/designer-ui/src/lib/tokenpicker/index.tsx
+++ b/libs/designer-ui/src/lib/tokenpicker/index.tsx
@@ -186,7 +186,7 @@ export function TokenPicker({
         }}
       >
         <div
-          className="msla-token-picker-container"
+          className="msla-token-picker-container-v3"
           style={
             fullScreen
               ? { height: windowDimensions.height - 100, width: windowDimensions.width - (parseInt(PanelSize.Medium, 10) + 40) }


libs/designer-ui/src/lib/tokenpicker/index.tsx
```diff
@@ -186,7 +186,7 @@ export function TokenPicker({
         }}
       >
         <div
-          className="msla-token-picker-container"
+          className="msla-token-picker-container-v3"
```
preetriti1, line 189: @Eric-B-Wu - Will this impact our css class?

preetriti1 (changes requested):

libs/designer-ui/src/lib/tokenpicker/index.tsx
Eric-B-Wu: I think this might not actually affect our CSS because msla-token-picker-container is not a selector actually used in our css

yijyuan: Yeah, we are not using 'msla-token-picker-container' in tokenpicker.less

preetriti1 (approved):

PR #2128, opened 2023-04-25
fix(designer): prioritize items with false/undefined isCustomApi property

jaimereyesmx:
When isCustomApi property is defined by the host non-custom items (false value) will be sorted before the non-custom items (true value). 
If the host does not include this property for its items, then the values will be set to false for all the items and therefore the sorting rule will not be applied.


hartra344 (approved):

libs/utils/src/lib/models/operationDiscoveryResults.ts
```diff
@@ -30,6 +30,9 @@ export interface SomeKindOfAzureOperationDiscovery {
   isNotification?: boolean;
   api: OperationApi;
   description: string;
+  operationGroup?: {
```
preetriti1, line 33: Api is the operation group, why was there a need to add operationGroup explicitly... api is the connector and operation Group is also the connector

preetriti1 (changes requested):

libs/utils/src/lib/models/operationDiscoveryResults.ts
hartra344: @preetriti1 we should be adding this logic to the searchservice, until we can move it though I told him he could go ahead and add htis in a way that doesn't impact us. This was an extensibility oversight on our side.

jaimereyesmx: I addressed your feedback. Thank you!

hartra344 (approved):

preetriti1 (approved):

PR #2137, opened 2023-04-25
fix(Power Automate): Add conditional string rendering for designer when in xrmConnectionReferenceMode

dhruvnm:
Solution flows in Power Automate use the concept of a "connection reference" which is essentially a wrapper around a connection. This is used to enable ALM scenarios like solution import. In the past, customers have been confused by labels that pertain to "connections" when we are really using "connection references." This change makes the necessary label changes if the designer is in connection reference mode.

Screenshots of the changes labels below:

image

image

image


dhruvnm:
> 

dhruvnm:
> @dhruvnm please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> @microsoft-github-policy-service agree company="Microsoft"
> Contributor License Agreement

hartra344 (approved):

preetriti1:
Every connection creation is actually creating a connection reference, but we do not expose the terminology to the customer because they do not know how it is stored in backend and how that is used by client UI.
We do not see such text changes in the old designer so i am not aware of this is a new scenario which is being enabled first time in designer. Also we do not expose connector specific logic in every layer or as a designer option as well..

dhruvnm:
> Every connection creation is actually creating a connection reference, but we do not expose the terminology to the customer because they do not know how it is stored in backend and how that is used by client UI. We do not see such text changes in the old designer so i am not aware of this is a new scenario which is being enabled first time in designer. Also we do not expose connector specific logic in every layer or as a designer option as well..

We definitely have this logic in the old V1 designer. Connection reference here refers to Dataverse connection references which are known by customers to be stored in their Dataverse db. We have has customer tickets in the past of confusion with the connection nomenclature being used with DV connection references which is why we had to make that change. This change in V3 designer is to stay consistent with what PAuto already has in place today.

PR #2139, opened 2023-04-25
fix(powerautomate): Fix DynamicallyAddedParameter menu getting collapsed on mouseenter

Ram-G:
Bug: We were using a call to a guid() function to generate a key for the delete menu item. When the user mouses over the menu item, this causes a re-render which causes the guid() function to be called, and the menu item key to be changed -- this causes the menu to be collapsed.

Fix: This PR changes the menu item key to be a string constant. Also removes some dead code.

Before:
GIF 4-25-2023 4-36-03 PM

After:
GIF 4-25-2023 4-36-29 PM


hartra344 (approved):

preetriti1:
Please use fix(powerautomate) as label not fix(designer-ui) 
We do not want to refer the package folder where the change is being made-- also this is power automate specific so the changelist will indicate only PA specific code was changed

PR #2168, opened 2023-04-27
fix(types): Move LogicApps types to exportable files

ek68794998:
- Move `LogicApps`, `LogicAppsV2`, `OpenAPIV2`, and `Swagger` types from `.d.ts` files to exportable `.ts` files
- Update `import type` statements and `tsconfig.json` files per the above change
- Export `NodeInputs` type

There are **no** functional changes in this PR. (Just type changes & type export/import changes.)

*Note: The order of some imports got swapped automatically when I added all my imports, as I tried to keep them consistent (`import type` first, then standard `import`).*


libs/designer-ui/tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "noFallthroughCasesInSwitch": true,
     "resolveJsonModule": true
   },
-  "files": ["../../types/Swagger.d.ts"],
+  "files": [],
```
ek68794998, line 15: Wasn't sure if I should remove this line entirely or leave `[]`. Let me know if I should remove the line.

hartra344 (approved):
If the build is happy on this one then I am happy

PR #2171, opened 2023-04-27
feat(designer): Recurrence parameter defaults

rllyy97:
## Main Changes

Actions and triggers with recurrence parameters now have fallback defaults based on SKU, matching the functionality of BPMUX.


hartra344 (approved):

preetriti1:
I thought we discussed we will have the manifest provide the default values rather than what was done in BPMUX.

PR #2172, opened 2023-04-28
fix: customize default node height for Open Api operations

jaimereyesmx:
When the host saves a flow and then returns the updated flow to the Designer, the updated flow might need to be rerendered. This causes a call initializeGraphState in [ParseReduxAction.ts](https://github.com/Azure/LogicAppsUX/blob/main/libs/designer/src/lib/core/parsers/ParseReduxAction.ts) that eventually sets the nodes to their default side; however, OpenApi nodes do not have the same size as the default nodes causing the designer to look messed up.

Before the fix:
image

After the fix:
image

### DIFF
--- a/libs/designer-ui/src/lib/panel/recommendationpanel/operationSearchHeader/index.tsx
+++ b/libs/designer-ui/src/lib/panel/recommendationpanel/operationSearchHeader/index.tsx
@@ -1,11 +1,7 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
-
-/* eslint-disable react/jsx-no-literals */
 import { DesignerSearchBox } from '../../../searchbox';
-import { Checkbox, Icon, IconButton, Link, Text } from '@fluentui/react';
+import { Checkbox } from '@fluentui/react';
 import type { IDropdownOption } from '@fluentui/react/lib/Dropdown';
-import { Dropdown, DropdownMenuItemType } from '@fluentui/react/lib/Dropdown';
-import { useCallback } from 'react';
+import { Dropdown } from '@fluentui/react/lib/Dropdown';
 import { useIntl } from 'react-intl';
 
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/utils/graph.ts
```diff
@@ -33,11 +34,15 @@ export const DEFAULT_NODE_SIZE = {
...
+export const OPENAPI_NODE_HEIGHT = 88;
+
 // Creating generic layout nodes and edges below
 
-export const createWorkflowNode = (id: string, type?: WorkflowNodeType): WorkflowNode => ({
+export const createWorkflowNode = (id: string, type?: WorkflowNodeType, operationType?: string): WorkflowNode => ({
   id,
-  ...DEFAULT_NODE_SIZE,
+  width: DEFAULT_NODE_SIZE.width,
+  height: operationType && isOpenApiConnectionType(operationType) ? OPENAPI_NODE_HEIGHT : DEFAULT_NODE_SIZE.height,
```
preetriti1, line 45: How is OpenApi different from ApiConnection or any other node which has connection, we do not have custom heights for other operations so the issue might be somewhere else, @rllyy97 and @Eric-B-Wu to comment on this.

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/graph.ts
rllyy97: Priti is correct, all of our nodes should have the same default sizing which is just a single-line title and no icon row.  You mentioned it in your issue already, but this is an issue with all nodes getting reset to default sizes in our layout algorithm.  The solution would be to either run the update again or block the reset to defaults from occurring.

I've created an issue for this [here](https://github.com/Azure/LogicAppsUX/issues/2175) to keep track of it.

PR #2176, opened 2023-04-28
fix(designer): Update token picker location during scroll

yijyuan:
Add Event Listener for scroll action on the side panel, so the token picker button will have updated location when the panel is scrolled.
Current behavior:
tok
Behavior after fix;
token_afterFix

### DIFF
--- a/libs/designer-ui/src/lib/editor/base/plugins/tokenpickerbutton/index.tsx
+++ b/libs/designer-ui/src/lib/editor/base/plugins/tokenpickerbutton/index.tsx
@@ -31,6 +31,7 @@ export const TokenPickerButton = ({ showOnLeft, openTokenPicker }: TokenPickerBu
   const [editor] = useLexicalComposerContext();
   const [anchorKey, setAnchorKey] = useState<NodeKey | null>(null);
   const boxRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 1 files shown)


rllyy97 (approved):

libs/designer-ui/src/lib/editor/base/plugins/tokenpickerbutton/index.tsx
```diff
@@ -31,6 +31,7 @@ export const TokenPickerButton = ({ showOnLeft, openTokenPicker }: TokenPickerBu
   const [editor] = useLexicalComposerContext();
   const [anchorKey, setAnchorKey] = useState<NodeKey | null>(null);
   const boxRef = useRef<HTMLDivElement>(null);
+  const panel = document.getElementsByClassName('ms-Panel-scrollableContent')[0];
```
rllyy97, line 34: Just to catch if it can't find the element for whatever reason
```suggestion
  const panel = document.getElementsByClassName('ms-Panel-scrollableContent')?.[0];
```

yijyuan: This should be fine, it should always return the first element or undefined - and not generate any error. Just tested locally, 
image

PR #2177, opened 2023-04-28
fix(powerautomate): Update manual trigger serialization to match manifest and render only dynamically added parameters

Ram-G:
Changes in this PR:
- Update serialization and deserialization of manual trigger parameters to match updated manifest (PR: https://msazure.visualstudio.com/One/_git/Flow-RP/pullrequest/8032179)
- Render parameters only if `x-ms-dynamically-added` is true -- this is set to true for all user-defined parameters (as opposed to static parameters from Flow-RP).

### DIFF
--- a/libs/designer-ui/src/lib/floatingactionmenu/index.tsx
+++ b/libs/designer-ui/src/lib/floatingactionmenu/index.tsx
@@ -209,12 +209,12 @@ export const FloatingActionMenu = (props: FloatingActionMenuProps): JSX.Element 
     <>
       <div className="msla-dynamic-added-params-container">
         {dynamicParameterProps.map((props) =>
-          props.properties['x-ms-dynamically-added'] === true ? <DynamicallyAddedParameter {...props} key={props.schemaKey} /> : undefined
+          props.properties['x-ms-dynamically-added'] === true ? <DynamicallyAddedParameter {...props} key={props.schemaKey} /> : null
         )}
       </div>
       <div className="msla-floating-action-menu-container">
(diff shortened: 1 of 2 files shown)


Ram-G:
Sorry for late ping -- we would _really_ like to have this PR merged for a bug bash on Monday. It's a small change and scoped. Is anyone from LA team able to take a look so we can get it in today's package? Thank you very much in advance!

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -208,9 +208,9 @@ export const FloatingActionMenu = (props: FloatingActionMenuProps): JSX.Element
   return (
     <>
       <div className="msla-dynamic-added-params-container">
-        {dynamicParameterProps.map((props) => (
-          <DynamicallyAddedParameter {...props} key={props.schemaKey} />
-        ))}
+        {dynamicParameterProps.map((props) =>
+          props.properties['x-ms-dynamically-added'] === true ? <DynamicallyAddedParameter {...props} key={props.schemaKey} /> : undefined
+        )}
```
preetriti1, line 213: You should be returning null not undefined when no component needs to be rendered

preetriti1 (changes requested):

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
Ram-G: Thank you, updated

preetriti1 (approved):

rllyy97 (approved):

PR #2179, opened 2023-04-28
fix(designer): expose `updateParameterValidation` helper function

almaasrawi:


preetriti1 (approved):

preetriti1:
Are you going to call validator function with the specific parameterId, groupId and validationErrors from host?
How is this function help to trigger validation for all parameters

almaasrawi:
> Are you going to call validator function with the specific parameterId, groupId and validationErrors from host? How is this function help to trigger validation for all parameters

Yes, I am calling the validator function and then calling the update validation from the host on save. The function wouldn't trigger the validation, we do that on flow save, which would help give us the errors to call the update method with. I have the changes tested locally with yalc, I just need this to get the official package that we can use instead of my local setup.

preetriti1:
> > Are you going to call validator function with the specific parameterId, groupId and validationErrors from host? How is this function help to trigger validation for all parameters
> 
> Yes, I am calling the validator function and then calling the update validation from the host on save. The function wouldn't trigger the validation, we do that on flow save, which would help give us the errors to call the update method with. I have the changes tested locally with yalc, I just need this to get the official package that we can use instead of my local setup.

So you are iterating over all parameters and calling the exposed validator function as well? I am trying to understand how is the state data being manipulated outside designer in the host?

almaasrawi:
> > > Are you going to call validator function with the specific parameterId, groupId and validationErrors from host? How is this function help to trigger validation for all parameters
> > 
> > 
> > Yes, I am calling the validator function and then calling the update validation from the host on save. The function wouldn't trigger the validation, we do that on flow save, which would help give us the errors to call the update method with. I have the changes tested locally with yalc, I just need this to get the official package that we can use instead of my local setup.
> 
> So you are iterating over all parameters and calling the exposed validator function as well? I am trying to understand how is the state data being manipulated outside designer in the host?

Correct, on save: I'm iterating over all the node inputs, calling the validator, keeping them on an observable state which then gets watched on the designer host component which calls the `updateParameterValidation` when that observable state changes. I can include you on the other PR once we have a new LogicAppsUX package.

ek68794998:
Given that we have an approval and this is a non-impacting change, I'm going to go ahead and merge.

PR #2182, opened 2023-04-28
fix(designer): Consumption APIM dynamic data fetched properly, recurrence defaults passed properly

rllyy97:
## Main Changes
- Fixed a consumption APIM manifest issue
  - API ID is now accessible to dynamic data
- Passed sku down to functions that need it
  - > NOTE: I'm not a huge fan of the bulk this adds, but to have this functionality we need to pass the sku, if we want to discuss just not having the different defaults I'd be fine with that just for the sake of simplicity


libs/services/designer-client-services/src/lib/consumption/manifests/apiManagement.ts
```diff
@@ -66,9 +66,14 @@ export const apiManagementActionManifest = {
...
                 dynamicState: {
                   operationId: 'getApimOperations',
-                  parameters: {},
                 },
-                parameters: {},
+                parameters: {
+                  type: 'object',
+                  apiId: {
+                    parameterReference: 'api.id',
+                    required: true,
```
preetriti1, line 74: Hoping these changes are tested in consumption with dynamic data.

preetriti1 (approved):

preetriti1 (changes requested):
Please remove the sku addition and we can define one set of defaults for everyone.

libs/services/designer-client-services/src/lib/consumption/manifests/apiManagement.ts
rllyy97: They were yes

preetriti1 (approved):

Issue #2175, opened 2023-04-28
Node layout sizing resets to defaults on workflow update


rllyy97:
### Describe the Bug with repro steps

When a workflow is resubmitted to the designer after a flow has already been processed, the designer resets all node sizes back to their default values (only in our layout algorithm, not in the actual viewport).  
Normally this results in the spacing between nodes becoming much smaller.

___

![image](https://user-images.githubusercontent.com/25409734/235185085-11c5b84b-0bf6-4b08-89d1-bd12c8699196.png)

Turns into => 

![image](https://user-images.githubusercontent.com/25409734/235185020-8c6518d5-341d-4def-85eb-7be6ec7ac176.png)

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Browser

Chrome

preetriti1:
@rllyy97 - Will this help in our rendering in Portal as well or we have to come up with different mechanism to handle this bug.
Current PA team is going to pass an option on designer component on when to not recalculate.

PR #2183, opened 2023-04-28
fix(designer): add parameter to specify whether the workflow graph should be recreated #2175

jaimereyesmx:
The host can specify whether the workflow graph should be recreated. This is used for scenarios where the flow is already loaded into the designer and recreating it will modify the existing node dimensions.

Before the fix:
image

After the fix:
image

Fixes https://github.com/Azure/LogicAppsUX/issues/2175

### DIFF
--- a/libs/designer/src/lib/core/BJSWorkflowProvider.tsx
+++ b/libs/designer/src/lib/core/BJSWorkflowProvider.tsx
@@ -14,15 +14,16 @@ export interface BJSWorkflowProviderProps {
   workflow: Workflow;
   runInstance?: LogicAppsV2.RunInstanceDefinition | null;
   children?: React.ReactNode;
+  shouldRecreateWorkflowGraph?: boolean;
 }
 
-const DataProviderInner: React.FC<BJSWorkflowProviderProps> = ({ workflow, children, runInstance }) => {
+const DataProviderInner: React.FC<BJSWorkflowProviderProps> = ({ workflow, children, runInstance, shouldRecreateWorkflowGraph }) => {
   const dispatch = useDispatch<AppDispatch>();
   useEffect(() => {
     dispatch(initWorkflowSpec('BJS'));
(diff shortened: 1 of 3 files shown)


preetriti1 (approved):

libs/designer/src/lib/core/BJSWorkflowProvider.tsx
```diff
@@ -14,15 +14,16 @@ export interface BJSWorkflowProviderProps {
   workflow: Workflow;
   runInstance?: LogicAppsV2.RunInstanceDefinition | null;
   children?: React.ReactNode;
+  shouldRecreateWorkflowGraph?: boolean;
 }
 
-const DataProviderInner: React.FC<BJSWorkflowProviderProps> = ({ workflow, children, runInstance }) => {
+const DataProviderInner: React.FC<BJSWorkflowProviderProps> = ({ workflow, children, runInstance, shouldRecreateWorkflowGraph }) => {
```
ek68794998, line 20: Does this get called even if `shouldRecreateWorkflowGraph` is already `true`? That is, if the data provider is set up with `shouldRecreateWorkflowGraph=false`, then we set `shouldRecreateWorkflowGraph=true`, it will dispatch a render. What happens if we then need to re-render again?

ek68794998: I have a feeling we should not use a prop here, and instead should have our own call to dispatch the initialization. @hartra344 ?

jaimereyesmx:
Eric looked at it this afternoon (thanks Eric!) and had some Perf concerns due to unnecessary rerenders. Eric suggested that instead of passing this variable we dispatch the action isWorflowGraphLocked=true when save starts and then unlock the Graph after save is done to avoid unnecessary rerenders due to changes to the variable. Closing this PR for now.

PR #2184, opened 2023-04-28
fix(openapi): Make parameters optional in OpenAPI builder

ek68794998:
Code added in [`openapi/inputsbuilder.ts`](https://github.com/Azure/LogicAppsUX/pull/1721/files#diff-75eeb5274c2d267589ddfb8d49e94407030df7fa8594036767f45a82f5c46694) in #1721 was based on the v1 designer which handles parameters differently. In v1, parameters are always truthy (`{}` is passed if no parameters are set). However, v3 designer does not do this, resulting in an uncaught error when `inputsInDefinition.parameters` is falsy:

image

This PR makes `parameters` nullable and adds a check in `_loadKnownParametersFromDefinition` to verify that they are set.

### DIFF
--- a/libs/designer/src/lib/core/utils/openapi/inputsbuilder.ts
+++ b/libs/designer/src/lib/core/utils/openapi/inputsbuilder.ts
@@ -1,7 +1,7 @@
 import type { InputParameter } from '@microsoft/parsers-logic-apps';
 
 export interface OpenApiConnectionSerializedInputs {
-  parameters: Record<string, unknown>;
+  parameters?: Record<string, unknown>;
 }
 
 export class OpenApiOperationInputsBuilder {
@@ -16,7 +16,7 @@ export class OpenApiOperationInputsBuilder {
     inputsInDefinition: OpenApiConnectionSerializedInputs | undefined,
     inputParameters: InputParameter[]
   ): InputParameter[] {
-    if (!inputsInDefinition) {
+    if (!inputsInDefinition?.parameters) {
       return inputParameters;
     }
 
--- a/libs/designer/src/lib/core/utils/openapi/__test__/inputsbuilder.test.ts
+++ b/libs/designer/src/lib/core/utils/openapi/__test__/inputsbuilder.test.ts
@@ -1,0 +1,69 @@
+import type { OpenApiConnectionSerializedInputs } from '../inputsbuilder';
+import { OpenApiOperationInputsBuilder } from '../inputsbuilder';
+
+describe('OpenApiOperationInputsBuilder', () => {
+  describe('loadStaticInputValuesFromDefinition', () => {
+    test('handles undefined definition', () => {
+      const builder = new OpenApiOperationInputsBuilder();
+
+      const inputParameters = [{ key: 'inputs.$.to', name: 'to', type: 'string' }];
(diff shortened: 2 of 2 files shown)


libs/designer/src/lib/core/utils/openapi/inputsbuilder.ts
```diff
@@ -1,7 +1,7 @@
 import type { InputParameter } from '@microsoft/parsers-logic-apps';
 
 export interface OpenApiConnectionSerializedInputs {
-  parameters: Record<string, unknown>;
+  parameters?: Record<string, unknown>;
 }
 
 export class OpenApiOperationInputsBuilder {
```
preetriti1, line 7: Why is there a need for this inputsBuilder? It could have been done using the parameter.alias and the manifest deserializer.

ek68794998: That wasn't brought up in #1721 but the logic can be moved there instead.

For now, I'd like to get this fix in so we don't have to deal with this error, and we can refactor the logic as needed after.

preetriti1: inputsLocation is not correctly added in your manifest, this should ne easily handled without a inputsbuilder, we cannot keep doing so many cleanups after GA... please fix it the right way...
make sure your openApiConnection manifests have [inputs, parameters] in the inputsLocation and then fix the code which serializes or deserializes if there are bugs.

ek68794998: The problem is that the code you're asking us to remove is already in, and if we didn't have this fix, the cleanup would be after GA anyway. This is a fix for that code. We risk breaking our scenarios at Build if we try to do some funky refactoring now.

Per your specific concern, yes, `inputsLocation` is correct:

image

The problem is that the object specified by that path can be `undefined`. I will look at the existing code for how we can leverage it, but effectively what you're asking us to do is jeopardize the functionality of OpenAPI before Build.

ek68794998: @preetriti1 Okay, this would be the alternative change. We'd remove the inputs builder, and do this instead:

image

image

I still feel this is too high-risk to push in on such a tight timeline. However, it does seem to resolve the issue fixed in #1721 and the issue in this current PR, and allows us to get rid of the inputs builder.

libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
```diff
@@ -103,53 +100,43 @@ export const getInputParametersFromManifest = (
   let primaryInputParametersInArray = unmap(primaryInputParameters);
 
   if (stepDefinition) {
-    const nodeType: string | undefined = stepDefinition.type;
```
ek68794998, line 106: This is a revert of #1721 and is easier to read with whitespace changes hidden.

libs/designer/src/lib/core/utils/openapi/inputsbuilder.ts
ek68794998: Updates have been made.

preetriti1 (approved):

Issue #2012, opened 2023-04-13
Designer stringifies boolean variable and dynamic values 


xuehongg:
### Describe the Bug with repro steps

This issue happens to both the exiting designer and the preview designer. 

Basically the designer stringifies any boolean value and boolean dynamics value.  For example, a boolean variable with a 'false' value is stringified as 'False' which is not a valid json value and cause problem with downstream problem.

Repro.
1. Create a boolean variable with a initial value of false
2. Add a create blob action and use the above boolean variable as part of the content.
3. Run the flow and check the raw inputs of the blob action, you will see 'False', instead of the expected 'false'.
4. See screenshots below as an example.

### Screenshots or Videos

Use boolean variable or boolean dynamic value as input to action
![image](https://user-images.githubusercontent.com/49498706/231842123-ee273132-0c45-44ba-9456-3e2239cb32f1.png)

The boolean variable comes out as "False", not the expected "false"
![image](https://user-images.githubusercontent.com/49498706/231842291-490936b4-689f-4184-8a71-5ab4aff6cf4d.png)

The problem is in the designer stringifies the variable or dynamic value in code.
![image](https://user-images.githubusercontent.com/49498706/231842377-70a3bfcc-58e3-4f8e-b407-8a43664bd1c6.png)

### Browser

Chrome

PR #2186, opened 2023-04-29
fix(designer): initializeVariable not showing correct boolean values

Eric-B-Wu:
Before: 
before

After:
after2


libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -140,8 +140,8 @@ const ParameterSection = ({
       const { viewModel } = newState;
       const parameter = nodeInputs.parameterGroups[group.id].parameters.find((param: any) => param.id === id);
       if (
-        (parameter?.type === constants.SWAGGER.TYPE.BOOLEAN && value.length === 1 && value[0]?.value === 'True') ||
```
preetriti1, line 143: I dont understand why are we patching the value here for boolean parameters in forst place, who added this code initially?

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
Eric-B-Wu: I added this code originally to fix https://github.com/Azure/LogicAppsUX/issues/2012, but you're right that this should be done in the serialization

preetriti1 (approved):

PR #2192, opened 2023-05-01
fix(designer): Fix default invoker toggle

bnarayanared:
Before:
By default the invoker connection toggle was on:
image

After the fix:
By default the "Use Invoker's Connection" toggle is off:
image


preetriti1 (changes requested):
Can you make sure your PR is only showing your changes, 372 files changed cant review

PR #2201, opened 2023-05-02
fix(designer): Fix expression for dynamic trigger output tokens

shuktika12163:
Dynamic tokens for trigger outputs have actionName in the outputInfo:
image

This causes the trigger outputs to get converted into @body('trigger_name') format which is invalid:
image

This PR fixes the issue by not setting actionName if the operation is a trigger. This makes the dynamic token show up as @triggerBody() instead:
image

### DIFF
--- a/libs/designer/src/lib/core/utils/outputs.ts
+++ b/libs/designer/src/lib/core/utils/outputs.ts
@@ -416,7 +416,13 @@ export const loadDynamicOutputsInNode = async (
         dispatch(
           addDynamicTokens({
             nodeId,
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -416,7 +416,13 @@ export const loadDynamicOutputsInNode = async (
         dispatch(
           addDynamicTokens({
             nodeId,
-            tokens: convertOutputsToTokens(nodeId, operationInfo.type, dynamicOutputs, { iconUri, brandColor }, settings),
+            tokens: convertOutputsToTokens(
+              isTrigger ? undefined : nodeId,
```
shuktika12163, line 420: This is how nodeId is passed in other areas where this function is used too.

PR #2218, opened 2023-05-03
Add LA Designer panel for token selecting

MStmaul:
Added to tester:
image

Panel:
image

Unit tests:
image

Issue with E2E test, are these still used? Maybe issue on my end:
image


preetriti1 (changes requested):
What is this feature regarding and why is it passing props through options whej the component is in the designer.
If there are custom components for designer feature and not being shared by all hosts then please consider adding service layer like old designer to get the custom from designer.

MStmaul:
Hi @preetriti1 I am closing this in favor of [this PR](https://github.com/Azure/LogicAppsUX/pull/2253). Changing to just exporting a few more things for our repo.

PR #2232, opened 2023-05-04
feat(designer): Increased search request size, search loading is now faster

rllyy97:
## Main Changes
- Increased search data load request page size to 2500 from 250
  - Per request, we are getting much more data in mostly the same request time
  - Should be a much faster load overall
- Removed our old request batching as well and are now requesting pages in series
  - Results in FAR less failed requests, and FAR less simultaneous requests
- Reintroduced search preloading
  - Now that we're not batching 50 requests on loading search (now it's just 1), we are able to start preloading on application start again without the performance / network bottleneck
- Reduced custom operation request page size to 100 at the request of custom connector engineers


hartra344:
@rllyy97  I think we can merge this in now. It should be fully out now

rllyy97:
Just tested in my local region and works great

PR #2235, opened 2023-05-04
fix(designer): Use skipValidation to skip throwing errors in serializeWorkflow

shuktika12163:
Currently, skipValidation option is not used in serializer. We need a way to get the latest workflow from designer even when parameters might be empty (to undo to the previous state of workflow while user is editing). Not having skipValidation option throws an error and doesn't serialize the workflow when there are errors, eg. parameters are empty.

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -58,26 +58,29 @@ export interface SerializeOptions {
 }
 
 export const serializeWorkflow = async (rootState: RootState, options?: SerializeOptions): Promise<Workflow> => {
-  const operationsWithSettingsErrors = (Object.entries(rootState.settings.validationErrors) ?? []).filter(
-    ([_id, errorArr]) => errorArr.length > 0
-  );
-  if (operationsWithSettingsErrors.length > 0) {
-    throw new Error(
-      'Workflow has settings validation errors on the following operations: ' +
-        operationsWithSettingsErrors.map(([id, _errorArr]) => id).join(', ')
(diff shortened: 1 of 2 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -58,26 +58,29 @@ export interface SerializeOptions {
 }
 
 export const serializeWorkflow = async (rootState: RootState, options?: SerializeOptions): Promise<Workflow> => {
```
shuktika12163, line 60: @Eric-B-Wu There are two places in LAUX repo where skipValidation is passed as "true". Can you please also confirm this wouldn't break those scenarios? Or I can update them to not pass the value for now since it wasn't being used before anyway.

https://github.com/Azure/LogicAppsUX/blob/b14642c7187a3b06907c270df9a60a3d364d534b/apps/designer-standalone/src/app/AzureLogicAppsDesigner/DesignerCommandBar.tsx#:~:text=skipValidation%3A%20true%2C

https://github.com/Azure/LogicAppsUX/blob/a8a9471148437222de7eb84af89ebc9479346f3c/apps/vs-code-designer-react/src/app/DesignerCommandBar/index.tsx#:~:text=skipValidation%3A%20true%2C

Eric-B-Wu (approved):

libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
shuktika12163: Updated to "skipValidation: false" in vs-code-designer-react as per discussion with Eric

PR #2238, opened 2023-05-04
fix(openapi): Fix dynamic schema params being duplicated

ek68794998:
In #1721, we had prevented OpenAPI parameters from executing the code path which populated **all parameters** that had values in the step definition, even if they were dynamic. However, in #2184, this code path was reactivated.

Now, parameters with a dynamic schema that have an existing value are populated twice: once using the value, which has a complete reference key (e.g., `body/body/recipient/to`), and once from the manifest, which has a dynamic schema reference key (e.g., `body/body/recipient`). This results in two objects being shown in the parameter list.

image

This fix addresses the issue so that if a parameter has a dynamic schema, we do not statically populate it in the parameters list.

image


rllyy97 (approved):

preetriti1:
Is this an issue only with aliasing parameter or any operation. Have we made sure this does not break any LA scenarios? @rllyy97

PR #2239, opened 2023-05-04
fix(designer): respect existing node dimensions when there is a workflow already loaded by the designer

jaimereyesmx:
For situations where there is an existing workflow, respect the node dimensions so that they are not reset.


libs/designer/src/lib/core/state/workflow/workflowInterfaces.ts
```diff
@@ -32,5 +32,6 @@ export interface WorkflowState {
   newlyAddedOperations: Record<string, string>;
   runInstance: LogicAppsV2.RunInstanceDefinition | null;
   isDirty: boolean;
+  /** @deprecated this will be removed in a follow up PR */
```
ek68794998, line 35: Can also add this for `setIsGraphLocked`.

ek68794998:
Looks good to me. Will let Riley review and sign off.

libs/designer/src/lib/core/parsers/ParseReduxAction.ts
```diff
@@ -27,11 +27,17 @@ export const initializeGraphState = createAsyncThunk<
...
+    // For situations where there is an existing workflow, respect the node heights so that
+    // they are not reset
+    const previousChildren = workflow?.graph?.children || [];
+    const deserializedChildren = deserializedWorkflow?.graph?.children || [];
+    for (const node of deserializedChildren) {
+      const previousNode = previousChildren.find((item) => item.id === node.id);
+      if (previousNode?.height) {
+        node.height = previousNode.height;
+      }
     }
```
rllyy97, line 39: This does not appear to be taking into account subgraphs / scope nodes, those nodes will have their own children so you will need to make this recursive or else only the top level will be considered

libs/designer/src/lib/core/parsers/ParseReduxAction.ts
```diff
@@ -27,11 +28,10 @@ export const initializeGraphState = createAsyncThunk<
     getConnectionsQuery();
     const { definition, connectionReferences, parameters } = workflowDefinition;
     const deserializedWorkflow = BJSDeserialize(definition, runInstance);
-    // The host can decide whether to make updates to the graph object. In situations like updating the workflow
-    // after save, it makes sense to not update the graph to avoid resetting things like dimensions
-    if (workflow.isGraphLocked && workflow.graph) {
-      deserializedWorkflow.graph = workflow.graph;
-    }
+    // For situations where there is an existing workflow, respect the node dimensions so that they are not reset
+    const previousGraphFlattened = flattenWorkflowNodes(deserializedWorkflow?.graph?.children || []);
```
jaimereyesmx, line 32: this should be workflow.graph

preetriti1:
@rllyy97 - why dont we have this issue in our portal? is there scenario different from ours?

rllyy97:
@preetriti1 Their use-case involves returning updated workflow objects to the designer after the initial load, we don't do this in portal currently, we make all our changes and save in-place

hartra344:
@preetriti1 @rllyy97 This is likely an issue we'll start running into when we start integrating CoPilot ourselves.

hartra344 (approved):

PR #2245, opened 2023-05-05
feat(designer): Workflow Parameters - Dirty State

rllyy97:
## Main Changes
- Added workflow parameters dirty state
  - Combined workflow and workflow parameters dirty state to a single selector and reset action


libs/designer/src/lib/core/index.ts
```diff
@@ -9,6 +9,7 @@ export type { NodeInputs } from './state/operation/operationMetadataSlice';
 export { useOperationsInputParameters } from './state/operation/operationSelector';
 export { discardAllChanges, setFocusNode, setIsGraphLocked, setIsWorkflowDirty } from './state/workflow/workflowSlice';
 export { useIsWorkflowDirty, useNodeDisplayName } from './state/workflow/workflowSelectors';
+export { useIsDesignerDirty, resetDesignerDirtyState } from './state/global';
```
ek68794998, line 12: Could you also export `useIsWorkflowParametersDirty`? I have a feeling we may want to consume that.

ccastrotrejo (approved):

libs/designer/src/lib/core/index.ts
rllyy97: Yep I can do that 👍

PR #2270, opened 2023-05-08
fix(designer): fix serialization and deserialization of simple query builder (oldFormat Condition Editor)

Eric-B-Wu:
simplequerybuilderfix


libs/designer-ui/src/lib/querybuilder/SimpleQueryBuilder.tsx
```diff
@@ -117,27 +117,28 @@ const convertRootPropToValue = (rootProps: RowItemProps): string => {
   return `@${rootProps.operator}(${op1},${op2})`;
 };
 
-const convertValueToRootProp = (value: ValueSegment[]): GroupItems => {
+const convertValueToRootProp = (value: ValueSegment[], items: RowItemProps): GroupItems => {
   const input = value[0].value;
   const operation: string = input.substring(input.indexOf('@') + 1, input.indexOf('('));
   const operations = input.split(',');
-  const operand1: ValueSegment[] = [
-    { id: guid(), type: ValueSegmentType.LITERAL, value: removeQuotes(operations[0].substring(operations[0].indexOf('(') + 1).trim()) },
```
preetriti1, line 125: Do we make sure multiple segments are not added by user here? basically does operand1 and operand2 only take 1 token as input and replace if you try to add another (old designer behavior), else you might have to handle multiple segments case here.

preetriti1: So it will pick up the first segment, if people file bug we can fix it then.

preetriti1 (approved):

PR #2275, opened 2023-05-08
feat(designer): Handling partially loaded or error nodes during serialization.

preetriti1:
This tries to serialize the error nodes correctly -
1. For nodes in critical error state, just get the loaded definition.
2. For connection references not present do not bother to add an empty one.
3. For connection errors show the error in UI.

Pending:
1. Handle failure in dynamic inputs load, this should show required parameters so runtime does not fail.
2. When nodes are in recommendation phase.


hartra344 (approved):

libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -59,13 +60,35 @@ export interface SerializeOptions {
...
 export const serializeWorkflow = async (rootState: RootState, options?: SerializeOptions): Promise<Workflow> => {
   if (!options?.skipValidation) {
+    const intl = getIntl();
+
+    const operationsWithConnectionErrors = Object.entries(rootState.operations.errors).filter(
+      ([_id, errors]) => !!errors[ErrorLevel.Connection]
+    );
+    if (operationsWithConnectionErrors.length > 0) {
+      throw new Error(
+        intl.formatMessage(
```
hartra344, line 70: is it possible to show a intl message to the use but throw a english error message. This is going to make it hard to look for in logs.

preetriti1: I thought this message would be visible in UI, but yes we can throw an exception which will have innerException details which i guess should be logged by Portal right?

PR #2276, opened 2023-05-08
fix(designer): Fix non-awaited dynamic data update call

ek68794998:
A regression was introduced in #2254 where a call that used to be instant (i.e., synchronous) had an `async` function added to it but that call was not `await`-ed. This caused an issue where dynamic data could not be populated when other events fired, leading to it disappearing from the UI.

This PR adds the appropriate `await` to ensure all dynamic data updates are completed before completing the deserialization process.

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
@@ -491,6 +491,7 @@ export const updateDynamicDataInNodes = async (getState: () => RootState, dispat
     tokens: { variables },
     connections,
   } = rootState;
+  const dynamicDataUpdatePromises: Promise<void>[] = [];
   const allVariables = getAllVariables(variables);
   for (const [nodeId, operation] of Object.entries(operations)) {
     if (!errors[nodeId]?.[ErrorLevel.Critical]) {
@@ -502,22 +503,25 @@ export const updateDynamicDataInNodes = async (getState: () => RootState, dispat
       const nodeOperationInfo = operationInfo[nodeId];
       const connectionReference = getConnectionReference(connections, nodeId);
 
-      updateDynamicDataForValidConnection(
-        nodeId,
-        isTrigger,
-        nodeOperationInfo,
-        connectionReference,
-        nodeDependencies,
-        nodeInputs,
-        nodeMetadata,
-        nodeSettings,
-        allVariables,
-        dispatch,
-        rootState,
-        operation
+      dynamicDataUpdatePromises.push(
+        updateDynamicDataForValidConnection(
+          nodeId,
+          isTrigger,
+          nodeOperationInfo,
+          connectionReference,
+          nodeDependencies,
+          nodeInputs,
+          nodeMetadata,
+          nodeSettings,
+          allVariables,
+          dispatch,
+          rootState,
+          operation
+        )
       );
     }
   }
+  await Promise.all(dynamicDataUpdatePromises);
 };
 
 const updateDynamicDataForValidConnection = async (


libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
```diff
@@ -502,22 +503,25 @@ export const updateDynamicDataInNodes = async (getState: () => RootState, dispat
...
-        connectionReference,
-        nodeDependencies,
-        nodeInputs,
-        nodeMetadata,
-        nodeSettings,
-        allVariables,
-        dispatch,
-        rootState,
-        operation
+      dynamicDataUpdatePromises.push(
```
hartra344, line 506: why put it into an array and awaiting Promise.all rather than just waiting the function call?

ek68794998: If we await every call, then there's no concurrency. If we have calls A, B, and C, then we have to wait for A to finish before B can start, and B to finish before C can start. With `Promise.all`, all 3 get kicked off and we just wait for them all to finish concurrently.

If you prefer simple `await`, I can verify that solves the problem as well.

hartra344: right, I know the reason for Promise.all but I only see 1 promise being added to the array so no concurrency needed?

preetriti1: I dont understand this was supposed to be fire and forget, since the next set of calls need not wait for updatedynamicdata to be completed. Previously also updateDynamicDataInNode was async and there was no await. What changed then?

ek68794998: @hartra344 There is a `for` loop on line 496. I assume that could run more than once?

@preetriti1 I wonder if we are running into a race condition here. We have no way of knowing when the dynamic data is populated from this callback, so we retrieve it after deserialization and expect it to be populated at that time. It's possible that previously, the call to get dynamic data completed quickly enough, but the addition of the connection check has made the call take too long.

ek68794998: In fact, with this change, we are only `await`-ing the call on line 537 (the connection validation), but not the actual dynamic data update (line 540).

If we want to make this clearer as a fire-and-forget, we could pull the connection validation out of that method, `await` it in this loop, and then fire-and-forget the `update*` call. Thoughts?

preetriti1: The main reason for conection check is to show the errors, And getting the connection should actually not be making a api call because connections have to be present at this stage.
n parseredux action nobody is awaiting the dynamic call to be completed, it was always supposed to be at the end when everything else is loaded. If we put awaits then it is going to caused the designer load to be slow and this call was not supposed to be waited call..
For every node dynamic call would complete at random times and would update hte card when you click on it, we can show some kind of loading indicator if needed (currently it is not).

So this was the change 

Before:
for () {
   callMethod()
}
callMethod() -> async method [which initializes all the dynamic cals and have so many awaits in it

After:
for() {
callMethodX()
}

callMethodX() {
  await ..
  callmethod()
}
again this is just an async method.

So from the caller it is not different from before than now, since it was never awaiting on the async call.

Maybe we need to understand which scenario is causing the race condition, because we do update UI when all these calls complete with dynamic data.

ek68794998: I think the call that is taking a while is the call to `getOperationManifest` in `isConnectionReferenceValid`. Let me add some benchmarks to illustrate the issue.

ek68794998: I am doing a bit more investigation but we may not need this fix. I pulled latest `main` and found that the issue is actually gone. It looks like this PR may have fixed it by retaining the dynamic output after its first fetch: [`b2ce1c1f4`](https://github.com/Azure/LogicAppsUX/commit/b2ce1c1f4a774ce77ecc37ff4ba7df192efee44b#diff-5443fb7dd2f0805c5e29c8285daae73c8c270087abf2502f40a32337743ce545)

I will do a test deployment of our bits with v2.15.14 and see if this resolves our issue.

ek68794998: Okay, I finally got a repro with my debugging statements in. Here is the issue&mdash;the `getConnection` call took 600ms, and during that time, a call was made to get the current workflow data. During this time, the dynamic data is not populated. Thus, when we use that workflow data later on, it's missing the dynamic data that it should have.

Here's the example output data. Note the perf timestamp of when the call is made (20812), and that it's before `getConnection` finishes:

image

Our calls to get the workflow come directly from the store, and we use `BJSSerializeWorkflow`. Below, you can see that the call that is made at `20812` (which matches the one above) has no `body/body/receipient/to` key (it is dynamic), and then the subsequent calls *do* have it&mdash;but it has an empty value. This is the bug we're hitting now.

image

We do not have a good measure of when the workflow is fully loaded and okay to be serialized, so we rely on the serialization to return us the correct result. But there seems to be a short period, now, when that data is *not* available in the serialization, due to the `getConnection` call.

ek68794998: I _think_ I have a workaround on the host side for the value disappearing (by preventing us from fetching the workflow until it's edited by the user). However, I don't think this is a good long-term fix as the user could edit the flow _immediately_ on load, and the UI bug of the field disappearing and reappearing still exists.

Should I abandon this PR and open a tracking issue to get this bug resolved?

preetriti1: Yes, I would suggest open a tracking bug, because we need to deal with this state to know when deserialization is completely fixed. We used to have a loading state for all nodes in old designer and we can have a similar incorporation here as well.
So i see there are these issues -
1. Dynamic content [dropdown values/inputs] disappearing and appearing...
2. Host to know when deserialization is completely done so we can make the parameters in editable state, or let the user know loading is still happening so they know if they edit now something can break.
3. Even when serialization is called before, we should be returning the data from old definition until we know operation becomes in editable state.

Can you mention the above file filing the tracking issue?

ek68794998: I will open the tracking bug now, but wanted to give an update.

Item 1 that you outlined above actually has a profound negative impact on us. It looks like the call to `updateDynamicDataInNode` happens at the time when the dynamic parameters disappear from the state. As a result, we do not update dynamic data for those nodes.

image

This means that those inputs do not get their dynamic data populated (they remain as the ugly ID values rather than usable dropdowns):

image

I have to test and see if the `await` fixes this. It might be the only workaround we can use to ship on time.

preetriti1: So the dynamic list is not updated for dynamic schema properties.... there are still awaits inside the updatedynamicdatainnodes, so i wonder putting await on previous method is going to fix this

ek68794998:
This PR is not needed but will look into making some tracking items for related issues with regard to how dynamic data is loaded.

PR #2282, opened 2023-05-08
fix(designer): Fixed style change for token picker

spanvalkar:


hartra344:
how important is this for you? I'm hesitant to bring this change in while we're prepping for GA/release because of the issues it could potentially cause.

spanvalkar:
This change fixes this bug: https://dev.azure.com/msazure/One/_workitems/edit/20553107

It's a style/appearance change, so it could potentially be pushed out. What issues do you foresee that may happen?

hartra344:
> This change fixes this bug: https://dev.azure.com/msazure/One/_workitems/edit/20553107
> 
> It's a style/appearance change, so it could potentially be pushed out. What issues do you foresee that may happen?

What I'm worried most about is that it was done for portal integration on our side and that remove it and moving that much lower will cause things not to show up in our portal. We've already gotten to our "ask mode" for our GA release so I don't want to take anything that would add extra risk to that.

spanvalkar:
That sounds reasonable. Let's keep this PR open for now and we can review/merge after the GA release is complete for you.

Within what timeframe do you estimate that we can merge this change?

Eric-B-Wu:
Our GA release has already taken a snapshot of new designer and is in our testing environment. However, we'll be rolling out hotfixes to the branch throughout the week as we fix customer/vendor testing bugs. The release will complete 05/19

hartra344:
We will be in ask mode however until BUILD in case there are important fixes that need to go in

Eric-B-Wu (approved):

PR #2284, opened 2023-05-09
feat(designer): Adding folder picker for manifest based operations

shuktika12163:
**Folder picker before**:
image

**Folder picker after**:
image

image

**Also able to navigate the picker, save and run the flow**:
image


libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -235,12 +235,18 @@ const ParameterSection = ({
     },
     getDisplayValueFromSelectedItem: (selectedItem: any): string => {
       const dependency = dependencies.inputs[parameter.parameterKey];
-      const propertyPath = dependency.filePickerInfo?.fullTitlePath ?? dependency.filePickerInfo?.browse.itemFullTitlePath;
+      const propertyPath =
+        dependency.filePickerInfo?.fullTitlePath ??
+        dependency.filePickerInfo?.browse.itemFullTitlePath ??
+        dependency.definition?.extension?.dynamicState?.dynamicExtension?.browse.itemFullTitlePath;
```
shuktika12163, line 241: Is there a better way to pass or get the paths from dynamicState?

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -291,12 +293,18 @@ export async function getFolderItems(
...
-  const { definition, filePickerInfo } = dependencyInfo;
-  if (isLegacyDynamicValuesTreeExtension(definition) && filePickerInfo) {
-    const { open, browse } = filePickerInfo;
-    const { connectorId } = operationInfo;
-    const connectionId = connectionReference?.connection.id as string;
-    const { operationId, parameters: referenceParameters } = selectedValue ? browse : open;
+  const { definition, filePickerInfo, parameter } = dependencyInfo;
+
+  if (!filePickerInfo) {
+    throw new UnsupportedException('File picker information does not exist.');
```
shuktika12163, line 299: Can change this to throw the same error as below "`Dynamic extension '${definition.type}' is not implemented yet or not supported`", or something else. I thought it's cleaner to read the code if we return early in this condition.

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
preetriti1: There should be one place where paths are atleast for one extension why do we have two fallbacks for dynamictree?

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
preetriti1: During initialization this is always set this should not be undefined can we remove this check?

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
preetriti1: These callbacks are provided so UI code does not assume how to get these values.

libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -181,6 +181,45 @@ export const getLegacyDynamicTreeItems = async (
...
+
+  const values = await queryClient.fetchQuery(
+    [
+      'dynamictreeitems',
+      connectionId.toLowerCase(),
+      connectorId.toLowerCase(),
+      operationId?.toLowerCase(),
+      getParametersKey(parameters).toLowerCase(),
+    ],
+    () => service.getTreeDynamicValues(connectionId, connectorId, operationId, parameterAlias, parameters, dynamicState)
```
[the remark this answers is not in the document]
shuktika12163: Ok. We have the parameterAlias dependency in a few places I believe. I'll keep this one for consistency for now for Build deadline (getListDynamicValues call also has this) and follow up with the team about how to remove dependency on parameterAlias in different places.

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
shuktika12163: Lint complains that this could be undefined here. I'll move the filePickerInfo check back to the existing if condition. Was just trying to take out redundant common code for the two cases below.

libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -181,6 +181,45 @@ export const getLegacyDynamicTreeItems = async (
...
+          id: treeDynamicValue.dynamicState?.selectionState?.id,
+          [dynamicState.dynamicExtension.browse.itemValuePath]: treeDynamicValue.value,
+          [dynamicState.dynamicExtension.browse.itemFullTitlePath]: treeDynamicValue.fullyQualifiedDisplayName,
+        },
+      };
+    });
+  }
+
+  return values;
+};
```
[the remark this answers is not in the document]
preetriti1: the connectors [x-ms-dynamic-tree].browse.itemFullTitlepath value should be equal to the dynamicitem returned from the call[itemFullTitlePath] if that is not the case then itemFullTitlePath should never be set in the connector because noOne would be using it in the client.

Something is redundant here based on the metadata.
And please do all value modifications in your srvice it should not be done here, this is comon code and LA might not send results like this.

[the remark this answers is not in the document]
shuktika12163: Yes, I'm trying to move the value modifications to our service but am unable to get things working end to end without a change to getPickerCallbacks function and need your help here.

For more context:

During initialization of action, we don't have itemFullTitlepath etc. values in dynamicState.browse. They exist in dynamicState.dynamicExtension.browse instead. This is not a big problem. I can add the values to dynamicState.browse too on our end, so that filePickerInfo.browse has these values:
image

But the problem is when I'm trying to get the values from tree dynamic call and use those values to set the display in parameter (getPickerCallbacks being used to set the value for display), filePickerInfo values don't match the propertyPath in selectedValue.

For example:

For the call to get Root for folder picker, our tree dynamic call returns these values:
image

We want to use the value and fullyQualifiedDisplayName from here to display the value in parameter.

Result of get dynamic tree call on LAUX side (I can update the dynamic extension browse parameters here in my service to use fullyQualifiedDisplayName instead):
image

But this doesn't seem to help with filePickerInfo.browse which is used to display the value in parameter panel.

It might be easier to show and solve if you have 10 mins for a call.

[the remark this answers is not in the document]
shuktika12163: Basically, is it possible to update filePickerInfo.browse based on this tree dynamic call or is it okay to use "dependency.definition?.extension?.dynamicState?.dynamicExtension?.browse.itemFullTitlePath" or similar for fallback in the getPickerCallbacks function?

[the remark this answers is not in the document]
preetriti1: If you are already updating the returned item to have 'Path': '/' , now this returned item is the selectedItem in pickerCallbacks
then in the picker callbacks - when the call comes for getdisplayValueFromSelectedItem it will get 
selectedItem['Path'] because we do selectedItem[dependency.filepickerInfo.browse.itemFullPath]

dynamic tree extensions are supposed to contain all metadata inside open/browse.

extension is the x-ms-dynamic-tree which is prsent in parameter schema in manifest, i dont know where the confusion is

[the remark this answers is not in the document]
shuktika12163: Actually, I think I found a way to pass it correctly so that filePickerInfo works. I'll try a couple more things first.

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
shuktika12163: Removed changes from this file.

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
shuktika12163: Updated

libs/designer/src/lib/core/queries/connector.ts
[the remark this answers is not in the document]
shuktika12163: Fixed this to only return values.

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -320,6 +328,20 @@ export async function getFolderItems(
...
+    const pickerParameters = Object.keys(referenceParameters ?? {}).reduce((result: Record<string, any>, paramKey: string) => {
+      return { ...result, [paramKey]: referenceParameters?.[paramKey] };
+    }, {});
+    const parameters = { ...pickerParameters };
+
+    const operationParameters = getParameterValuesForDynamicInvoke(parameters, nodeInputs, idReplacements);
+
+    let dynamicState = { ...definition.extension.dynamicState };
+    if (selectedValue) {
+      dynamicState = { ...dynamicState, selectionState: { id: selectedValue.id } };
```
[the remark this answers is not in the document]
shuktika12163: Updated to use selectionState instead

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -320,6 +323,20 @@ export async function getFolderItems(
...
+    const { open, browse } = filePickerInfo;
+    const { connectorId } = operationInfo;
+    const connectionId = connectionReference?.connection.id as string;
+    const { operationId, parameters } = selectedValue ? browse : open;
+
+    const operationParameters = getParameterValuesForDynamicInvoke(parameters as DynamicParameters, nodeInputs, idReplacements);
+
+    let dynamicState = definition.extension.dynamicState;
+    if (selectedValue) {
+      dynamicState = { ...dynamicState, selectionState: selectedValue.selectionState };
```
preetriti1, line 336: When we implement this extension for LA connectors, this code will be removed because you are not relying on the selectedItemValuePath property of the reference parameters which tells which property to extract from the selectionState and that should be sent in parameters today. Your service should be reading that parameter and modify your  api call if these need to be sent in dynamicState property.

If you do not want to change it now, it will be breaking change for you then.

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
shuktika12163: Thanks Priti! We are okay with taking the breaking change in if it's okay to leave this like this for now. We'll fast-follow next week to use/pass selectedItemValuePath correctly instead.

preetriti1 (approved):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
shuktika12163: Created this PBI for fast-follow next week:
[Product Backlog Item 21887309](https://msazure.visualstudio.com/One/_workitems/edit/21887309): [Fast-follow][Folder picker] Use selectedItemValuePath from referenceParameters to get selection state

Issue #2203, opened 2023-05-02
Logic App Designer not setting AuthenticationType for Managed Identity connection


beedyg:
### Describe the Bug with repro steps

1. Create new Logic App
2. Enable Managed Identity
3. Use template "When a resource event occurs"
4. Complete details of event (observed for Microsoft.Storage.BlobCreated)
5. Use Managed Identity as connection
6. Save

In Preview the UI presents a save successful message. However in network tab the API returns:
{"error":{"code":"WorkflowManagedIdentityConfigurationInvalid","message":"The workflow connection parameter 'azureeventgrid' is not valid. The API connection 'azureeventgrid' is configured to support managed identity but the connection parameter is either missing 'authentication' property in connection properties or authentication type is not 'ManagedServiceIdentity'."}}
And the changes were not saved

In GA the error message is displayed in the UI.

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using Preview Designer or GA Designer

Generally Available

### Browser

Edge

preetriti1:
This work is related to supporting managed identity in connectors which only support single auth, but multiple connection variations.

PR #2285, opened 2023-05-09
feat(designer): Legacy Managed Identity connection support

rllyy97:
## Main Changes
- Connectors with legacy managed identity connections are now supported
  - This is styled the same as our existing multi-auth even though it uses different logic
  - This lets us switch between OAuth, Service Principal, and legacy managed identity on connectors that support it
  - For every connector I've checked, if it has legacy MI, it has OAuth and SP

Fixes https://github.com/Azure/LogicAppsUX/issues/2203

image
image
image


libs/utils/src/lib/helpers/connections.ts
```diff
@@ -151,6 +151,13 @@ export function connectorContainsAllServicePrinicipalConnectionParameters(
   );
 }
 
+export function usesLegacyManagedIdentity(alternativeParameters?: Record<string, ConnectionParameter>): boolean {
+  return (
+    alternativeParameters?.['authentication']?.uiDefinition?.schema?.['x-ms-editor-options']?.supportedAuthTypes?.[0] ==
```
preetriti1, line 156: Shouldn't there be 3 equal signs?

preetriti1 (approved):

PR #2292, opened 2023-05-09
fix(designer): Support Type Filtering in Token Picker

### LINKED ISSUE #2003: Consumption_Token window is hiding in above option taskbar.

1.Create a LA in consumption and http trigger.
2.Add an action and try to add some token : Token window is open.

**Actual:** Token window is overlapping at above.
image

**Expected:** Token window is in fit view.

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

Chrome

### Additional context

**Environmental Details:**

**URL:** [https://aka.ms/emastaging-flowstip1](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Faka.ms%2Femastaging-flowstip1&data=05%7C01%7Cv-govubilish%40microsoft.com%7C5dc1b66779e54ebfb20b08db3bf92e3f%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638169713681866036%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=Ox8FGZYcViENAadUWbkyOt05VGv%2FAvR7SYpEG%2BGHVDI%3D&reserved=0)

**Version :** 2.30410.1.3

Eric-B-Wu:
Filters Tokens in Token Picker based on type of the Editor. Should work for both simple and complex arrays
Also Fixes #2003


libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -128,7 +128,7 @@ const ParameterSection = ({
   expressionGroup: TokenGroup[];
 }) => {
   const dispatch = useDispatch<AppDispatch>();
-  const [sectionExpanded, setSectionExpanded] = useState<boolean>(false);
+  const [sectionExpanded, setSectionExpanded] = useState<boolean>(true);
```
preetriti1, line 131: Any reason to change this?

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -274,19 +274,42 @@ const ParameterSection = ({
...
+            token.outputInfo.functionName === constants.FUNCTION_NAME.VARIABLES ||
+            token.outputInfo.functionName === constants.FUNCTION_NAME.PARAMETERS ||
+            token.outputInfo.arrayDetails ||
+            token.key === constants.UNTIL_CURRENT_ITERATION_INDEX_KEY ||
+            token.key === constants.FOREACH_CURRENT_ITEM_KEY
+          );
+        }
+        return supportedTypes.some((supportedType) => equals(supportedType, token.type));
+      }),
+    }));
```
rllyy97, line 305: Might not be needed, but if these have potential to take a while on large token counts we might want to wrap them in useMemos

rllyy97 (approved):

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -274,19 +274,42 @@ const ParameterSection = ({
...
+      ? getPropertyValue(constants.TOKENS, parameterType) || [constants.SWAGGER.TYPE.ANY, parameterType]
+      : null;
+
+    const filteredTokenGroup = tokenGroup.map((group) => ({
+      ...group,
+      tokens: group.tokens.filter((token: OutputToken) => {
+        if (isCodeEditor) {
+          return !(
+            token.outputInfo.functionName === constants.FUNCTION_NAME.VARIABLES ||
+            token.outputInfo.functionName === constants.FUNCTION_NAME.PARAMETERS ||
```
preetriti1, line 297: Please use token.outputInfo.tokenType to check for variables or parameters, we have different type of tokens already stamped in the object, do not use functionName directly, it is supposed to be only used during serialization if needed

preetriti1 (changes requested):

libs/designer-ui/src/lib/settings/settings.less
```diff
@@ -89,7 +89,6 @@
 
     .msla-setting-section-setting {
       margin-left: 46px;
-      display: flex;
```
preetriti1, line 92: I think this is so commonly used can you check if recurrence parameters [since they are in different section] in a trigger are rendered correctly and the settings panel as well?

libs/designer-ui/src/lib/table/index.tsx
```diff
@@ -87,6 +89,8 @@ export const TableEditor: React.FC<TableEditorProps> = ({
             labelId={labelId}
             keyTitle={titles[0]}
             valueTitle={titles[1]}
+            editorKeyType={types[0]}
```
preetriti1, line 92: DictionaryEditor always has key type as string and shouldn't these be set in the editorOptions?

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -42,6 +42,7 @@ export interface SettingTokenFieldProps extends SettingProps {
   expressionGroup?: TokenGroup[];
   isTrigger?: boolean;
   isCallback?: boolean;
+  editorValueType?: string;
```
preetriti1, line 45: Nobody is using this and we should not be passing this as props to settingtoken field, because editors internally should get their sub types from editorOptions

libs/designer-ui/src/lib/editor/string/index.tsx
```diff
@@ -16,6 +16,7 @@ export const StringEditor = ({
   initialValue,
   labelId,
   clearEditorOnTokenInsertion,
+  editorValueType,
```
preetriti1, line 19: Can you call it valueType? editor prefix seems redundant for all editors.

libs/designer-ui/src/lib/editor/base/index.tsx
```diff
@@ -44,6 +44,7 @@ export type GetTokenPickerHandler = (
   editorId: string,
   labelId: string,
   tokenPickerMode?: TokenPickerMode,
+  editorType?: string,
```
preetriti1, line 47: Is this being used?

libs/designer-ui/src/lib/editor/base/index.tsx
```diff
@@ -67,6 +68,7 @@ export interface BaseEditorProps {
   showCallbackTokens?: boolean;
   labelId?: string;
   label?: string;
+  editorValueType?: string;
```
preetriti1, line 71: just valueType?

libs/designer-ui/src/lib/dictionary/index.tsx
```diff
@@ -21,6 +22,7 @@ export interface DictionaryEditorProps extends BaseEditorProps {
   initialItems?: DictionaryEditorItemProps[];
   keyTitle?: string;
   valueTitle?: string;
+  editorKeyType?: string;
```
preetriti1, line 25: Just key type.

libs/designer-ui/src/lib/code/index.tsx
```diff
@@ -118,7 +118,7 @@ export function CodeEditor({
         />
       ) : null}
       {getInTokenPicker()
-        ? getTokenPicker?.(editorId, callOutLabelId, undefined, closeTokenPicker, onClickTokenPicker, tokenClicked)
+        ? getTokenPicker?.(editorId, callOutLabelId, undefined, undefined, closeTokenPicker, onClickTokenPicker, tokenClicked)
```
preetriti1, line 121: Do you mind providing arg comments, on these undefined arguments

preetriti1 (changes requested):

libs/designer-ui/src/lib/table/index.tsx
Eric-B-Wu: I do default dictionary key type to be string, but doesn't table editor support different types for the key? Looking at the editorOptions, I can see that it has a types array... though given that the 2 cases which we use the table editor are both of type string for the key, I still think we should make this a dynamic type, right?

libs/designer-ui/src/lib/editor/base/index.tsx
Eric-B-Wu: yeah its what we pass from the editor to the tokenpicker

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
preetriti1: We never have it expanded in our GA, if they want it can be done in a separate PR.

libs/designer-ui/src/lib/editor/string/index.tsx
Eric-B-Wu: yeah sure, had did this because I didn't want to confuse it with valueType and keyType of dictionary editor, but I can make this change

libs/designer-ui/src/lib/settings/settings.less
Eric-B-Wu: this was actually something I did to fix how the recurrence parameters looked. It looks fine in the settings panel

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
hartra344: I think expanding it is fine if we're not defaulting it(Last I checked we weren't)

Agreed though it should be a separate PR

libs/designer-ui/src/lib/dictionary/plugins/SerializeExpandedDictionary.tsx
```diff
@@ -33,9 +33,9 @@ export const SerializeExpandedDictionary = ({ items, initialItem, index, type, s
         }
         $setSelection(null);
       });
+      setIsDelete(true);
```
preetriti1, line 36: What is this fixing?

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -274,19 +274,40 @@ const ParameterSection = ({
...
     tokenClickedCallback?: (token: ValueSegment) => void
   ): JSX.Element => {
-    const codeEditorFilteredTokens = tokenGroup.filter((group) => {
-      return group.id !== 'workflowparameters' && group.id !== 'variables';
-    });
+    const parameterType =
+      editorType ??
+      (nodeInputs.parameterGroups[group.id].parameters.find((param) => param.id === parameterId) ?? {})?.type ??
+      constants.SWAGGER.TYPE.ANY;
+    const supportedTypes: string[] = parameterType ? getPropertyValue(constants.TOKENS, parameterType) : null;
```
preetriti1, line 287: When is the parameterType null? line:283 makes sure it is never null right?

preetriti1 (approved):

PR #2295, opened 2023-05-09
fix(designer): Deletion of ForEach operation causes designer to crash

### LINKED ISSUE #2293: Deletion of ForEach operation causes designer to crash

1. Create a flow that has an operation before and after a "for each" operation.
2. Add an operation in the for each scope.
3. Delete the for each operation **through the operation side-pane**
4. Designer crashes.

**This only happens when deleting through the side-pane, deletion via right click + confirmation of delete prompt does not repro**

I believe this issue is happening due to deletion of the for each node before deletion of it's nested actions (I can't figure out when the nested items get deleted). While stepping through the code we can see that the nested item (compose in this case) will hit a null reference since compose still has a parentNodeId tied to the deleted forEach node.

```export const getAllParentsForNode = (nodeId: string, nodesMetadata: NodesMetadata): string[] => {
  let currentParent = nodesMetadata[nodeId]?.parentNodeId; // can have parentNodeId tied to deleted forEach node
  const result: string[] = [];

  while (currentParent) {
    result.push(currentParent);
    currentParent = nodesMetadata[currentParent].parentNodeId; // can hit null reference here
  }

  return result;
};
```

See relevant delete action below and the 'updateAllUpstreamNodes' function which leads to the null reference above

```
export const deleteOperation = createAsyncThunk(
  'deleteOperation',
  async (deletePayload: DeleteOperationPayload, { getState, dispatch }) => {
    batch(() => {
      const { nodeId } = deletePayload;

      dispatch(clearFocusNode());
      dispatch(clearPanel());

      dispatch(deleteNode(deletePayload));
      deleteOperationDetails(nodeId, dispatch);
      updateAllUpstreamNodes(getState() as RootState, dispatch);
    });
  }
);
```

Consumption (Portal)

Preview

### Workflow JSON

_No response_

### Screenshots or Videos

nestedGraphDeletionBug

### Browser

Chrome, etc

### Additional context

_No response_

maturpi:
fixes #2293

Operations that could act as a nested graph (if, forEach, scope, etc) have special delete handling when deleting the nodes through right click. This is because using right click on a card lets us known the context of the operation and uses it's own handleDelete function. When deleting through the side panel, it does a general delete and misses these special cases.

This PR fixes the issue by adding a conditional check on whether the selected node is one of these special cases and deletes it as such.

**Before**
nestedGraphDeletionBug2ndExample

**After**
nestedGraphDeletionBugFix


libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -305,43 +325,54 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
...
+        }}
+        trackEvent={handleTrackEvent}
+        onCommentChange={onCommentChange}
+        title={selectedNodeDisplayName}
+        onTitleChange={onTitleChange}
+      />
+      <DeleteNodeModal
+        nodeId={selectedNode}
+        // nodeIcon={iconUriResult.result}
+        // brandColor={brandColor}
```
preetriti1, line 369: remove comments

libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -192,6 +197,26 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
...
+      constants.NODE.TYPE.FOREACH,
+      constants.NODE.TYPE.SCOPE,
+      constants.NODE.TYPE.UNTIL,
+    ];
+    if (scopeNodeTypes.includes(operationInfo?.type.toLowerCase())) {
+      const scopeId = selectedNode.split('-#')[0];
+      dispatch(deleteGraphNode({ graphId: scopeId ?? '', graphNode }));
+    } else {
+      dispatch(deleteOperation({ nodeId: selectedNode, isTrigger: isTriggerNode }));
+    }
```
preetriti1, line 216: Please use the method isScopeNode to check for these types, we hsould not hard code it in the containers and also do not calculate scopeId like this, if that is hte graphId, then it should be graphnode.id

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -305,43 +325,54 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
...
+          togglePanel();
+        }}
+        trackEvent={handleTrackEvent}
+        onCommentChange={onCommentChange}
+        title={selectedNodeDisplayName}
+        onTitleChange={onTitleChange}
+      />
+      <DeleteNodeModal
+        nodeId={selectedNode}
+        // nodeIcon={iconUriResult.result}
```
ccastrotrejo, line 368: If these aren't needed (commented). There is no need to commit these lines.

ccastrotrejo (approved):

preetriti1:
Does this work for switch case?

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: These were taken from the DeleteNodeModal in ScopeCardNode. I can remove, didn't know if they were left there for context.

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: Sure - copy pasted this code from ScopeCardNode so wasn't sure if the commented out lines were left there for context.

maturpi:
> Does this work for switch case?
@preetriti1 yes it works for all of these whereas they weren't before:

      constants.NODE.TYPE.IF,
      constants.NODE.TYPE.SWITCH,
      constants.NODE.TYPE.FOREACH,
      constants.NODE.TYPE.SCOPE,
      constants.NODE.TYPE.UNTIL,

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: ACK on isScopeNode - however for the scopeId, this is how it is calculated in ScopeCardNode, is there a reason why it should be different here? @preetriti1

maturpi: Actually there's no IsScopeNode function as far as I can tell

libs/designer/src/lib/ui/panel/panelroot.tsx
rllyy97: They were left on the ScopeCardNode component because we were possibly going to have new modal designs that would use those props, but design never gave us anything so it was just leftover

libs/designer/src/lib/ui/panel/panelroot.tsx
preetriti1: You just need the nodeId, and that might be just outdated code and since the nodeIds do not contain # in their name it is just working since first segment is fetched, can you please just use selectedNode for your operation and see if something is failing (it should not but still try it out)

libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -192,6 +197,19 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
     dispatch(expandPanel());
   }, [dispatch]);
 
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const handleDeleteClick = () => setShowDeleteModal(true);
+  const handleDelete = () => {
+    // TODO: 12798935 Analytics (event logging)
+    if (isScopeOperation(operationInfo?.type)) {
```
preetriti1, line 204: For subgraphnodes i do see the delete method calling deleteSwitchCase as well, so if you explicitly delete case maybe not all graphs get deleted. @rllyy97 to confirm on this.

rllyy97:
> > Does this work for switch case?
> > @preetriti1 yes it works for all of these whereas they weren't before:
> 
> ```
>   constants.NODE.TYPE.IF,
>   constants.NODE.TYPE.SWITCH,
>   constants.NODE.TYPE.FOREACH,
>   constants.NODE.TYPE.SCOPE,
>   constants.NODE.TYPE.UNTIL,
> ```

This does not work on switch cases currently, it works for the switch but not the cases
image

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: I'll take a look to see if I can add handling for cases under a switch case.

libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -305,43 +318,52 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
...
+          togglePanel();
+        }}
+        trackEvent={handleTrackEvent}
+        onCommentChange={onCommentChange}
+        title={selectedNodeDisplayName}
+        onTitleChange={onTitleChange}
+      />
+      <DeleteNodeModal
+        nodeId={selectedNode}
+        nodeType={WORKFLOW_NODE_TYPES.GRAPH_NODE}
```
rllyy97, line 361: Make sure this type value is updated properly so the delete modal renders content with respect to the node

libs/designer/src/lib/ui/panel/panelroot.tsx
rllyy97: Yes Priti is correct, you'll call the `deleteGraphNode` function with the subgraph/case id and node object, then `deleteSwitchCase` with the subgraph/case id and the containing graph node id (the actual switch id)

maturpi: Fixed

maturpi: nestedGraphDeletionBugFixFinal

maturpi: Ignore bad quality of the GIF - had to make some edits to reduce size. There's no weird residue of the delete modal after accepting/declining the prompt lol

libs/designer/src/lib/ui/panel/panelroot.tsx
```diff
@@ -305,43 +319,52 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
...
+          togglePanel();
+        }}
+        trackEvent={handleTrackEvent}
+        onCommentChange={onCommentChange}
+        title={selectedNodeDisplayName}
+        onTitleChange={onTitleChange}
+      />
+      <DeleteNodeModal
+        nodeId={selectedNode}
+        nodeType={graphNode?.type || WORKFLOW_NODE_TYPES.GRAPH_NODE}
```
preetriti1, line 362: Do we need this fallback, fallback should not assume it is graphNode, ideally graphNode.type should always be populated so do not have this fallback for nodeType

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: graphNode wasn't populated on designer load and was throwing error due to undefined graphNode.type

maturpi: See line 327 where it implies the panel can be loaded without a node being selected:
```noNodeSelected={!selectedNode}```

Comment (preetriti1) [reply libs/designer/src/lib/ui/panel/panelroot.tsx:362]: ok, so do not render your deleteNodeModal when there is no graphNode, please do not fallback on a nodetype. Panel can be opened with node selection and deletenodemodal should not be present in such cases as well.

**Review by preetriti1 — CHANGES_REQUESTED** [review]

File: libs/designer/src/lib/ui/panel/panelroot.tsx [hunk 7]
```diff
@@ -42,8 +43,9 @@ import { testingTab } from './panelTabs/testingTab';
 import { RecommendationPanelContext } from './recommendation/recommendationPanelContext';
 import { WorkflowParametersPanel } from './workflowparameterspanel';
 import type { CommonPanelProps, MenuItemOption, PageActionTelemetryData } from '@microsoft/designer-ui';
+import { DeleteNodeModal } from '@microsoft/designer-ui';
 import { MenuItemType, PanelContainer, PanelHeaderControlType, PanelLocation, PanelScope, PanelSize } from '@microsoft/designer-ui';
```
rllyy97, line 47: Combine these imports

rllyy97 (approved):
I ran your branch locally and can confirm that it works as expected, just make sure to resolve the outstanding discussions

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: Done

libs/designer/src/lib/ui/panel/panelroot.tsx
maturpi: Done

preetriti1 (approved):

PR #2298, opened 2023-05-09
fix(designer): do not set Panel header horizontal padding when Panel is collapsed

jaimereyesmx:
Before:
image

Now:
image

### DIFF
--- a/libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
+++ b/libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
@@ -226,7 +226,7 @@ export const PanelHeader = ({
       </div>
       {!noNodeOnCardLevel ? (
         <>
-          <div className={'msla-panel-card-header'} style={isRight ? {} : { paddingLeft: horizontalPadding }}>
+          <div className={'msla-panel-card-header'} style={isRight || isCollapsed ? {} : { paddingLeft: horizontalPadding }}>
             {iconComponent}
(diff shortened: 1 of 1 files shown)


ek68794998:
LGTM. This meets our bug bar since the double-wide panel blocks the minimap controls.

libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
```diff
@@ -226,7 +226,7 @@ export const PanelHeader = ({
       </div>
       {!noNodeOnCardLevel ? (
         <>
-          <div className={'msla-panel-card-header'} style={isRight ? {} : { paddingLeft: horizontalPadding }}>
+          <div className={'msla-panel-card-header'} style={isRight || isCollapsed ? {} : { paddingLeft: horizontalPadding }}>
```
preetriti1, line 229: @rllyy97 - Will this remove any kind of padding in our portal which is needed?

jaimereyesmx: Only the padding when panel is on the left that I added last week. No impact to other padding.

rllyy97 (approved):

libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
rllyy97: Just tested and looks fine for us @preetriti1

PR #2325, opened 2023-05-10
feat(designer): Add static results support for manifest operations

Aleks-gdb:
Pass the ManifestParser to be able to use the static results feature on manifest based operations. Also add all operations from PowerApps as supported for static results


libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
```diff
@@ -248,6 +248,10 @@ export const getOutputParametersFromManifest = (
   };
 };
 
+export const getParsedManifest = (manifest: OperationManifest): ManifestParser => {
```
hartra344, line 251: why make this a function? seems redundant unless you plan on adding more functionality.

libs/utils/src/lib/helpers/connections.ts
```diff
@@ -58,6 +58,19 @@ export const isSharedManagedConnector = (connectorId: string) => {
   return true;
 };
 
+export const isSharedManagedConnectorFromPApps = (connectorId: string) => {
+  // Note: connectorId format: /providers/Microsoft.PowerApps/apis/{connector}
+  const fields = connectorId.split('/');
```
hartra344, line 63: instead of addressing random numbers of arrays, can this be deconstructed?

Wu-Bowen (approved):

Eric-B-Wu (approved):

libs/utils/src/lib/helpers/connections.ts
Eric-B-Wu: Currently, this is also how we check for our other managed connectors so I'm going to approve and merge, but I agree that we should deconstruct or just use a regex to check the format of the connector

libs/utils/src/lib/helpers/connections.ts
```diff
@@ -58,6 +58,19 @@ export const isSharedManagedConnector = (connectorId: string) => {
   return true;
 };
 
+export const isSharedManagedConnectorFromPApps = (connectorId: string) => {
```
preetriti1, line 61: Ideally we shouldn't be coding for paths like this, we should expose some function to host which can tell if it is shared, the reason being, if we support another provider tomorrow then that also need be coded here instead of the host and this is not scalable.

preetriti1: We should have like a resourceService that can determine if something is custom, armresource or sharedmanaged connector

PR #2340, opened 2023-05-11
feat(Data Mapper): Changed SchemaDropDown to ComboBox to Enable Schema Search

JoaquimMalcampo:
Changed Dropdown in schema view to ComboBox to enable search functionality.

comboboxnormal

When there is a large number of items, options can be scrolled:
comboboxmanyitems


refortie:
Could we also get the search functionality to filter? https://react.fluentui.dev/?path=/docs/components-combobox--default#filtering

libs/data-mapper/src/lib/components/configPanel/AddOrUpdateSchemaView.tsx
```diff
@@ -208,13 +208,16 @@ export const AddOrUpdateSchemaView = ({
       )}
 
       {uploadType === UploadSchemaTypes.SelectFrom && (
-        <Dropdown
+        <VirtualizedComboBox
```
refortie, line 211: Could we also get the search functionality to filter? https://react.fluentui.dev/?path=/docs/components-combobox--default#filtering

refortie (changes requested):

refortie:
We might also need to expose a visible scroll bar. We've been getting a fair few reports about "missing scrolling" when really the scroll bar is invisible. Maybe see what the fluent ux guidelines say about that?

PR #2345, opened 2023-05-11
fix(Designer): Fixed crash with manifest not having connector property 

rllyy97:
## Main changes
Recent changes to icon uri and brand color fallbacks assumed operation manifests had a connector property to read from.
Power Automate has operations without that property, this PR allows for this.


hartra344 (approved):

ek68794998:
Thank you! LGTM :)

PR #2348, opened 2023-05-12
fix(designer): Treat empty string as though the parameter has no value

wifun:
This handles situations where workflow definition has parameters with empty string values:
image

The goal of this change is for `isDynamicDataReadyToLoad` to return false even if there are parameters with empty string values. With this change `parameterValidForDynamicCall` will return false for empty string.
Otherwise the error is misleading:
image


Eric-B-Wu (approved):

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -2178,7 +2178,7 @@ export function getParameterFromId(nodeInputs: NodeInputs, parameterId: string):
 export function parameterHasValue(parameter: ParameterInfo): boolean {
   const value = parameter.value;
 
-  if (!isNullOrUndefined(parameter.preservedValue)) {
+  if (!isNullOrUndefined(parameter.preservedValue) && parameter.preservedValue !== '') {
```
preetriti1, line 2181: You can also use !isUndefinedOrEmptyString parameter.preservedValue) the reason being i think null is a valid value.

preetriti1 (changes requested):

preetriti1 (approved):

Issue #2234, opened 2023-05-04
Support for Parameter Casting on Complex Editors


Eric-B-Wu:
### Describe the Bug with repro steps

No Repro

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Browser

all browser

PR #2362, opened 2023-05-12
feat(designer): Added Casting to Array Editor/ Fixed Complex Array Editor

Eric-B-Wu:
Complex Array Editor should now be based on itemSchema instead of one-dimensional schema, making nested arrays possible

Added Casting to Complex and Simple array

Casting of simple arrays:
simpleArrayCast

Casting of complex arrays:

compexArrayCast


libs/designer-ui/src/lib/arrayeditor/plugins/CollapsedArrayValidation.tsx
```diff
@@ -26,17 +25,17 @@ export const CollapsedArrayValidation = ({
...
   setItems,
   setCollapsedValue,
 }: CollapsedArrayValidationProps): JSX.Element => {
   const [editor] = useLexicalComposerContext();
-  const [errorMessage, setErrorMessage] = useState('');
 
   const onChange = (editorState: EditorState) => {
     editorState.read(() => {
-      if (itemSchema) {
-        serializeComplexArray(editor, itemSchema, setItems as (complexItems: ComplexArrayItems[]) => void, setIsValid, setErrorMessage);
```
preetriti1, line 39: So no more error message needed?

libs/designer-ui/src/lib/arrayeditor/util/util.ts
```diff
@@ -1,87 +1,109 @@
...
       }
     });
   } else {
+    const isArray = itemSchema.type === constants.SWAGGER.TYPE.ARRAY && itemSchema.items?.properties;
     flattenedSchema.push({
-      title: itemSchema.title,
-      type: itemSchema.type,
-      isRequired: isRequired ?? false,
+      key: itemSchema.key,
+      title: itemSchema.title ?? itemSchema.key.split('.').slice(-1)[0],
```
preetriti1, line 39: itemSchema.key.split('.').at(-1)

libs/designer-ui/src/lib/arrayeditor/plugins/CollapsedArrayValidation.tsx
Eric-B-Wu: This was a "feature" I had added with the previous complex array editor to identify where the issue was; however, it became more difficult to support passing in the whole itemSchema. Though I think we can add this as a potential feature in the future. Right now it tries to convert the array of complex Items accordingly to the item schema. If it succeeds, we can view the expanded array. If not, the button to switch is blocked and it'll just show the error "Please enter a valid array"

libs/designer-ui/src/lib/arrayeditor/util/util.ts
```diff
@@ -1,87 +1,109 @@
...
+      type: isArray ? constants.SWAGGER.TYPE.ARRAY : itemSchema.type,
```
preetriti1, line 40: Why isn't itemSchema.type set to array for nested array?

libs/designer-ui/src/lib/arrayeditor/util/util.ts
```diff
@@ -1,87 +1,109 @@
...
+      isRequired: !isArray && isRequired,
```
preetriti1, line 41: I am wondering if this is read correctly from schema

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -463,16 +461,23 @@ const convertStringToInputParameter = (
   };
 };
 
+export const toArrayViewModelSchema = (schema: any): { arrayType: ArrayType; itemSchema: any } => {
+  const itemSchema = parseArrayItemSchema(schema);
+  const arrayType = schema?.type === constants.SWAGGER.TYPE.OBJECT && schema.properties ? ArrayType.COMPLEX : ArrayType.SIMPLE;
+  return { arrayType, itemSchema };
```
preetriti1, line 467: Shouldn't you be having uncasted value added here? lets make sure we always add it regardless of deserialization/authoring scenario.

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -354,6 +359,8 @@ const ParameterSection = ({
           onValueChange: (newState: ChangeState) => onValueChange(id, newState),
           onComboboxMenuOpen: () => onComboboxMenuOpen(param),
           pickerCallbacks: getPickerCallbacks(param),
+          onCastParameter: (value: ValueSegment[], type?: string, format?: string, suppressCasting?: boolean) =>
+            castParameterToString(value, type, format, suppressCasting),
```
preetriti1, line 363: Why can't you directly pass onCastparameter: castParameterToString
The arguments look exactly the same

libs/designer/src/lib/core/utils/parameters/helper.ts
Eric-B-Wu: if the uncastedValue is unset, it defaults to initialValue. I can make it so that we set uncastedValue to undefined though

libs/designer-ui/src/lib/arrayeditor/util/util.ts
Eric-B-Wu: it should be, I think I had set this earlier just to enforce that arrays should be arrays hahaha. Will just set it to be       type: itemSchema.type,

libs/designer-ui/src/lib/arrayeditor/util/util.ts
Eric-B-Wu: There may be some edge cases where it doesn't... I've covered all the edge scenarios I could find, so I'm thinking that until someone finds one that it fails, to keep as so

hartra344 (approved):

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -463,16 +461,23 @@ const convertStringToInputParameter = (
   };
 };
 
+export const toArrayViewModelSchema = (schema: any): { arrayType: ArrayType; itemSchema: any; uncastedValue: undefined } => {
+  const itemSchema = parseArrayItemSchema(schema);
+  const arrayType = schema?.type === constants.SWAGGER.TYPE.OBJECT && schema.properties ? ArrayType.COMPLEX : ArrayType.SIMPLE;
+  return { arrayType, itemSchema, uncastedValue: undefined };
```
preetriti1, line 467: your uncasted value is parameterValue, why set as undefined Aren't you always passing uncasted value from setting token picker?

preetriti1 (changes requested):

preetriti1 (approved):

Eric-B-Wu:
Fixes #2234

PR #2391, opened 2023-05-15
fix(designer): Add unit tests for canConvertToComplexCondition function

jaimereyesmx:
Add unit tests for canConvertToComplexCondition function


preetriti1:
@jaimereyesmx  in your last PR you only made sure your old definitions load for the simple expression. Is there a plan to put this in a editor code? It would be better to have this as a new editor which can switch between simple and complex based on the input value. The reason being we only put the code in helper.ts common path in loading parameter details, which requires definition, root state apart from parameter value. Your editor can get the entire parameter value as input and editor options from manifest Metadata to determine if your action supports hybrid mode. This is to keep the current editors clean since simple and complex are two different editors. If you want to update complex editor to support this case then simple has to be moved as one of the mode there and I think Eric separated it out.

The ask is to make the change in editor code and not here since you have all the info from parameter value. We should avoid making changes while creating the view model, view model has data which can only be inferred from manifest and parameter value and requires other data processing.

Issue #2385, opened 2023-05-15
Disappearing Exception Message in Action Shape


WarrenKinley:
### Describe the Bug with repro steps

Creating an exception in action shape

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/95749950/cfd38d9a-7ba8-4469-8788-0b14632a169d)

### Browser

Edge

Issue #2383, opened 2023-05-15
Full Action Shape Description cannot be Viewed


WarrenKinley:
### Describe the Bug with repro steps

1. In the new Logic App std designer, click on "Add an Action"
2. Choose the Azure Blob Storage
3. See more, some of the actions descritpions are hidden

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/95749950/cfbf09bc-5704-4fdd-8e22-2abca0a7773c)

### Browser

Edge

WarrenKinley:
Sorry not Consumption but Logic App Std preview designer in the portal

PR #2393, opened 2023-05-16
fix(designer): Cleanup of Dictionary Editor

### LINKED ISSUE #2290: Cleanup of Dictionary Editor

We can cleanup a lot of the dictionary serialization code by re-rendering the component based on a key, instead of clearing the editor and inserting the new text. This will then let us serialize on Editor Blur instead of editor change

Standard (Portal)

Preview

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

All

### Additional context

_No response_

Eric-B-Wu:
Fixes #2290

There were a couple of changes to dictionary editor:
1. Instead of onchange to serialize expanded dictionary editor, we now do it onblur
2. Instead of clearing the editor and repopulating the editor value when deleting a row in the expanded dictionary editor, we now just update the id causing a rerender of the component

By doing 2, we also fixed a bug where after deleting a row and only putting a token in the editor, the value was not serializing.
Before:

before

After

after

Also fixed a couple of items being cutoff styling bugs:
#2385 
errorMsg

#2383

token


libs/designer-ui/src/lib/dictionary/expandeddictionary.tsx
```diff
@@ -75,36 +87,22 @@ export const ExpandedDictionary = ({
...
-                <SerializeExpandedDictionary
-                  items={items}
-                  initialItem={item.key}
-                  index={index}
-                  type={ExpandedDictionaryEditorType.KEY}
-                  setItems={setItems}
-                />
-              </BaseEditor>
+                editorBlur={(newState: ChangeState) => handleBlur(newState, index, ExpandedDictionaryEditorType.KEY)}
+              ></StringEditor>
```
rllyy97, line 91: Super tiny nit, can we make these close with `/>` instead of a separate closing tag

rllyy97 (approved):

preetriti1 (approved):

PR #2401, opened 2023-05-17
feat(designer): Allow operations to specify unknown parameters to not be rendered or serialized

wifun:
Unfortunately there are some scenarios where the workflowDefinition to be loaded can contain parameters that do not exist in the manifest (and should not exist).

We want to have a way for the workflowDefinition to indicate that we do not want to show these parameters or serialize them.

current behavior:
image

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/helper.ts
+++ b/libs/designer/src/lib/core/utils/parameters/helper.ts
@@ -1131,7 +1131,7 @@ export function updateParameterWithValues(
                     isUnknown: true,
                   };
 
-                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ false));
+                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ true));
                 }
               }
             }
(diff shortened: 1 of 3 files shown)


wifun:
The plan is for PAuto's workflowDefinitions to always hide and not serialize unknown parameters.
I'm not 100% sure if this will break something, please let me know if there are any specific scenarios I should validate

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1131,7 +1131,7 @@ export function updateParameterWithValues(
                     isUnknown: true,
                   };
 
-                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ false));
+                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ true));
```
wifun, line 1134: From reading the code I'm assuming `invisible` should true since `createInvisibleParameter` is true

preetriti1: Please dont update this unless this is fixing some scenario.

wifun: Sure I didn't know of any existing scenarios to test it with, but the code reads like it's doing the wrong thing.

Undid the change

hartra344 (approved):

PR #2406, opened 2023-05-17
fix(designer): Moving Token Picker EntryPoint to Left

Eric-B-Wu:
Moved TokenPicker Entry Point to the left side for all editors to keep consistency. 

Added a type/description to Dynamic Content

Capture

Before merging, have PA signoff on this


hartra344 (approved):

libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickeroption.tsx
```diff
@@ -181,8 +182,13 @@ export const TokenPickerOptions = ({
                     <div className="msla-token-picker-section-option-text">
                       <div className="msla-token-picker-option-inner">
                         <div className="msla-token-picker-option-title">{token.title}</div>
-                        <div className="msla-token-picker-option-description">{token.description}</div>
+                        <div className="msla-token-picker-option-description" title={token.description}>
+                          {token.description}
+                        </div>
                       </div>
+                      {token.type && token.outputInfo?.type !== TokenType.FX ? (
```
preetriti1, line 189: Why do we want to show type to user?  User would know what is the parameter type they are entering on the dynamic content, these are implementation or developer logic/term right?
Also type any they wouldn't even understand. Was there any compelling reason to introduce this concept

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -496,8 +496,24 @@ const toSimpleQueryBuilderViewModel = (input: any): { isOldFormat: boolean; item
...
+    operand1 = loadParameterValue(
+      convertStringToInputParameter(
+        operand1String,
+        /* removeQuotesFromExpression */ true,
+        /*  trimExpression */ true,
+        /*convertIfContainsExpression */ true
+      )
+    );
+    console.log(operand1);
+    console.log(convertStringToInputParameter(operand2String, true, true, true));
```
preetriti1, line 508: Remove console logs

libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickeroption.tsx
preetriti1: Why dont we have showing type in the last PR for this feature and complete others like show more and stuff. I had put some questions in the original task for this.

PR #2409, opened 2023-05-17
fix(designer): Special case array actions being converted to Object on initializing, then breaking Runtime on Save

Eric-B-Wu:
Before
before

After
after

@preetriti1, can you confirm if this breaks anything in terms of updateInputsValueForSpecialCases function? I'm assuming that we don't want arrays to be included as they'll always be converted into object form


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -2306,7 +2306,7 @@ function swapInputsValueIfNeeded(inputsValue: any, manifest: OperationManifest)
  * Example: Batch trigger
  */
 function updateInputsValueForSpecialCases(inputsValue: any, allInputs: InputParameter[], customSwagger?: SwaggerParser) {
-  if (!inputsValue || typeof inputsValue !== 'object') {
+  if (!inputsValue || typeof inputsValue !== 'object' || Array.isArray(inputsValue)) {
```
preetriti1, line 2309: Why is the inputs Value getting changed here? It should only be executed or changed when we have property with serialization and deserialization set...it should not impact any other inputs..

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/helper.ts
Eric-B-Wu: This isn't happening because of the serialization/deserialization, but rather because we're using clone() which clones into an object... The issue is this doesn't work with arrays

preetriti1: I see, so can we fix it using copy instead of clone? Because i do not want to remove array from this equation when we support it.

preetriti1 (approved):

PR #2435, opened 2023-05-18
fix(designer): Fixed connections passing empty non-required parameters

rllyy97:
Our parameter "required" check for multi-auth connections was evaluating true on "false" strings, so we were getting undefined passed to the backend which was causing our connection creation errors
I made us a utils method that ensures the value be a boolean `true` or a "true" string

### DIFF
--- a/libs/utils/src/lib/helpers/functions.ts
+++ b/libs/utils/src/lib/helpers/functions.ts
@@ -937,3 +937,12 @@ export function findAncestorElement(
 export function hasInvalidChars(str: string, invalidChars: string[]): boolean {
   return invalidChars.some((invalidChar) => includes(str, invalidChar));
 }
+
+/**
+ * Returns true if true or "true"
+ * @arg {any} value - A value to compare.
+ * @return {boolean} - True if true or "true"
+ */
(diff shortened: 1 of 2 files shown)


preetriti1:
The value is never a boolean, it is always a string of value 'true' or 'false' that is how the api hub connector contract is.

ccastrotrejo (approved):

libs/utils/src/lib/helpers/functions.ts
```diff
@@ -937,3 +937,12 @@ export function findAncestorElement(
...
   return invalidChars.some((invalidChar) => includes(str, invalidChar));
 }
+
+/**
+ * Returns true if true or "true"
+ * @arg {any} value - A value to compare.
+ * @return {boolean} - True if true or "true"
+ */
+export function isTrue(value: any): boolean {
+  return value === true || value === 'true';
```
preetriti1, line 947: you should check based on typeof then... to keep this method generic, because i could say 0 or 1 also to be boolean when it is integer

preetriti1 (approved):

PR #2442, opened 2023-05-18
fix(designer): Shows Tokenpicker on Triggers

Eric-B-Wu:


hartra344 (approved):

libs/designer-ui/src/lib/editor/base/index.tsx
```diff
@@ -65,8 +65,6 @@ export interface BaseEditorProps {
   BasePlugins?: BasePlugins;
   initialValue: ValueSegment[];
   children?: React.ReactNode;
-  isTrigger?: boolean;
-  showCallbackTokens?: boolean;
```
preetriti1, line 69: What does showcallbacktokens set for?

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -351,8 +348,6 @@ const ParameterSection = ({
           editor,
           editorOptions,
           tokenEditor: true,
-          isTrigger,
-          isCallback: nodeType?.toLowerCase() === constants.NODE.TYPE.HTTP_WEBHOOK,
```
preetriti1, line 355: By removing web hook check would we still show the listcallbackurl token?

libs/designer-ui/src/lib/editor/base/index.tsx
Eric-B-Wu: previously because http callback was a trigger, we weren't showing the token picker, so this was a flag to let the editor know that if this was a HTTP callback to show the tokenpicker regardless of a trigger. Now this isn't needed cause we'll always show the picker

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
Eric-B-Wu: I have verified that we still do

preetriti1 (approved):

PR #2454, opened 2023-05-19
fix(designer): Workflow crashing when getting outputs of schema with multiple types

Eric-B-Wu:
In old designer we parse out the outputs with multiple types, but we weren't doing so in new designer. We expect type to be of type string, but when users input an array in something like a parse JSON, we then error out

Before:
before

After: 
after


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -2802,7 +2802,9 @@ function getOutputsByType(allOutputs: OutputInfo[], type = constants.SWAGGER.TYP
     return allOutputs;
   }
 
-  return allOutputs.filter((output) => equals(type, output.type));
+  return allOutputs.filter((output) => {
+    return Array.isArray(output.type) ? false : equals(type, output.type);
```
preetriti1, line 2806: nit: Simplify the expression as !Array.isArray(output.type) && equals(type, output.type)

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -301,7 +301,9 @@ const ParameterSection = ({
             token.key === constants.FOREACH_CURRENT_ITEM_KEY
           );
         }
-        return supportedTypes.some((supportedType) => equals(supportedType, token.type));
+        return supportedTypes.some((supportedType) => {
+          return Array.isArray(token.type) ? false : equals(supportedType, token.type);
```
preetriti1, line 305: nit: Simplify the expression as !Array.isArray(token.type) && equals(type, token.type)

preetriti1 (approved):

PR #2463, opened 2023-05-19
fix(designer): segment values are not showing properly

Eric-B-Wu:
before:

before

after:
after

### DIFF
--- a/libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
+++ b/libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
@@ -43,7 +43,7 @@ export const parseHtmlSegments = (value: ValueSegment[], tokensEnabled?: boolean
           data: segment,
           brandColor,
           icon: icon,
-          value: value ?? segmentValue,
+          value: segmentValue ?? value,
         });
         tokensEnabled && paragraph.append(token);
       } else {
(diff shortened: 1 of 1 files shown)


libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
```diff
@@ -43,7 +43,7 @@ export const parseHtmlSegments = (value: ValueSegment[], tokensEnabled?: boolean
           data: segment,
           brandColor,
           icon: icon,
-          value: value ?? segmentValue,
+          value: segmentValue ?? value,
         });
```
preetriti1, line 47: Didn't you already have this change made in some PR

Eric-B-Wu: I reversed this actually because it was showing the wrong value ... but I'm not able to repro the wrong value any more... so I'm just going to revert it

Eric-B-Wu: > Didn't you already have this change made in some PR

I'm unable to find this change in the change log, maybe it never got merged?

Eric-B-Wu: oh actually I found why, its in this PR https://github.com/Azure/LogicAppsUX/pull/2406/files, which hasn't checked in

PR #2480, opened 2023-05-22
fix(designer): Fix inputs/outputs and pager in nested for each loops

### LINKED ISSUE #2470: Logic App (Standard) VS Code debugger does not handle nested For Each loops correctly
### Describe the Bug

After a successful run the inner For Each loop always shows the number of iterations of the outer loop. It should be showing the number of iterations of that iteration of the inner loop. Additionally, any action inside the inner loop only appears to only show the details of the first iteration of the outer loop or, more likely, no input or output details even if the action was successful.

image

However, the workflow did run successfully and looking through the underlying storage tables shows that all loops were iterated correctly.

image

### Plan Type

Standard

1. Create the workflow
2. Run it in the debugger
3. Use the Overview to open the run history.

### Workflow JSON

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "Initialize_array": {
                "type": "InitializeVariable",
                "inputs": {
                    "variables": [
                        {
                            "name": "Data",
                            "type": "array",
                            "value": [
                                1,
                                2,
                                3,
                                4,
                                5,
                                6,
                                7,
                                8,
                                9,
                                10
                            ]
                        }
                    ]
                },
                "runAfter": {}
            },
            "For_each_value": {
                "type": "Foreach",
                "foreach": "@variables('Data')",
                "actions": {
                    "For_each_inner_item": {
                        "type": "Foreach",
                        "foreach": "@outputs('Compose_inner_array')",
                        "actions": {
                            "Compose_inner_value": {
                                "type": "Compose",
                                "inputs": "@items('For_each_inner_item')"
                            }
                        },
                        "runAfter": {
                            "Compose_inner_array": [
                                "SUCCEEDED"
                            ]
                        }
                    },
                    "Compose_inner_array": {
                        "type": "Compose",
                        "inputs": "@range(0, items('For_each_value'))"
                    }
                },
                "runAfter": {
                    "Initialize_array": [
                        "SUCCEEDED"
                    ]
                }
            }
        },
        "contentVersion": "1.0.0.0",
        "outputs": {},
        "triggers": {
            "When_a_HTTP_request_is_received": {
                "type": "Request",
                "kind": "Http"
            }
        }
    },
    "kind": "Stateful"
}
```

### Screenshots or Videos

_No response_

### Additional context

_No response_

AB#22449174

ccastrotrejo:
#### Main Code Changes
- Adding useEffect on pager component and handling the API call to handle scenario in which outer loop max number is bigger than the inner loop actual number
- Fix inputs and outputs display when either of them are numbers and arrays

Fixes #2470 

### Implementation
- General behavior

https://github.com/Azure/LogicAppsUX/assets/102700317/f50be221-0b5d-46e9-8ad1-5bc76f9016c3

- Pager behavior

https://github.com/Azure/LogicAppsUX/assets/102700317/d7d56310-1a1a-4d06-b7ec-5e1d5aee1bfc


libs/services/designer-client-services/src/lib/standard/run.ts
```diff
@@ -239,11 +241,14 @@ export class StandardRunService implements IRunService {
    * @returns {BoundParameters} List of parametes.
    */
   parseActionLink(response: Record<string, any>, isInput: boolean): BoundParameters {
-    if (!response) {
+    if (isNullOrUndefined(response)) {
       return response;
     }
 
-    const dictionaryResponse = isString(response) ? { [isInput ? 'Inputs' : 'Outputs']: response } : response;
+    const dictionaryResponse =
```
preetriti1, line 248: Array should not be handled as object?

preetriti1: What about isBoolean

ccastrotrejo: You are right, I will add the isBoolean. I handed the array as this because because it wasn't showing as an object, so I put it in an inputs/outputs object.  You can see the example below

- Before
image

- After
image

preetriti1 (approved):

PR #2524, opened 2023-05-24
fix(designer): Catch exception for when cors is not established for monitoring view

ccastrotrejo:
### Main Code Changes
- Add catch clause when getting the inputs/outputs

### What is missing
- Implementation in portal side of the updateCors function

### DIFF
--- a/libs/services/designer-client-services/src/lib/standard/run.ts
+++ b/libs/services/designer-client-services/src/lib/standard/run.ts
@@ -19,6 +19,7 @@ export interface RunServiceOptions {
   apiVersion: string;
   baseUrl: string;
   httpClient: IHttpClient;
+  updateCors?: () => void;
   accessToken?: string;
   workflowName: string;
   isDev?: boolean;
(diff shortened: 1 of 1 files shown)


libs/services/designer-client-services/src/lib/standard/run.ts
```diff
@@ -226,13 +228,22 @@ export class StandardRunService implements IRunService {
...
-      inputs = await this.getContent(inputsLink);
+    try {
+      if (outputsLink && outputsLink.uri) {
+        outputs = await this.getContent(outputsLink);
+      }
+      if (inputsLink && inputsLink.uri) {
+        inputs = await this.getContent(inputsLink);
+      }
+    } catch (e: any) {
+      if (e.message.includes('Failed to fetch') && !isNullOrUndefined(updateCors)) {
```
ccastrotrejo, line 239: If anyone knows a better way to catch a specific denied by CORS api call rather than just by the api message would be much appreciate it

hartra344 (approved):

libs/services/designer-client-services/src/lib/standard/run.ts
hartra344: Not really a CORS failure essentially means the call never actually happens at all

PR #2525, opened 2023-05-24
feat(designer): Add utility for extracting UI from tokens ad hoc

MStmaul:
Takes in a workflow definition and basic services to return back Token UI data. 

Needed for our case, where we store only token value data on the backend, so have to build the token UI from that outside the context of the designer.

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
@@ -148,7 +148,7 @@ export const initializeOperationMetadata = async (
   );
 };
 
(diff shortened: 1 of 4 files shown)


preetriti1 (changes requested):
Please setup sometime with our team for a design discussion before making these changes. We do not want to keep dispatches as optional else it will break the initialization process, if you need to add capability for your scenario we should discuss on what can be exposed and how it needs to be exposed for design.

MStmaul:
Messaged offline.

MStmaul:
Going to discuss further when GA release cools down... For now doing this in our repo.

hartra344:
I'm going to close this for now so we can discuss correct way we want to handle this for productization.

Issue #2539, opened 2023-05-24
Filter Array Expression works weird on the new designer in Logic Apps


nidubey1:
### Describe the Bug with repro steps

1. Go to the Logic Apps Designer and add a Filter Array expression and add the From as the array from previous action and From Query as expression: "length(split(item(), '|')?[0])"as below:

![image](https://github.com/Azure/LogicAppsUX/assets/85997053/f5988ea0-058b-470e-a007-c995d3ee3acc)

![image](https://github.com/Azure/LogicAppsUX/assets/85997053/b5ddb281-8e16-486d-87c5-1c614bbdbce3)

![image](https://github.com/Azure/LogicAppsUX/assets/85997053/797d8be5-827b-436b-b6d7-380afdd74a5a)

2. Next you save the workflow, it gets saved or even if you select some other action and then click on Filter Array action, it shows invalid parameters in filter array expression:

<img width="754" alt="image" src="https://github.com/Azure/LogicAppsUX/assets/85997053/c84c3bff-c591-40c1-a4a3-e007342b75e0">

3. Though the code view shows the correct expression:

<img width="707" alt="image" src="https://github.com/Azure/LogicAppsUX/assets/85997053/334894d4-c238-4366-b089-0f52ab50de17">

4. Also if we go to the fallback URL(https://aka.ms/lafallback), again everything looks fine:

<img width="767" alt="image" src="https://github.com/Azure/LogicAppsUX/assets/85997053/0d427953-ef9c-475e-b4c5-69fdf8cc0600">

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Browser

Edge

Eric-B-Wu:
Hello @nidubey1 thanks for reporting this issue, this is happening due to how we are parsing the editor, will try to get a fix in either today or tomorrow for it.

nidubey1:
Hi @Eric-B-Wu , thanks for the update, please do let me know once it is deployed.

Eric-B-Wu:
Hi @nidubey1  we have deployed onto all stations with the fix

PR #2543, opened 2023-05-24
fix(designer): make it so we split on outer most comma in simple query builder

Eric-B-Wu:
before: 

before

after: 
after
#2539


hartra344 (approved):

preetriti1 (commented):
Why are we not using the segments created in parameterValue to understand the expression. There you will have correctly what is the function and ots arguments.. it looks like string parsing with certain assumption on the functions being used. We have expression parser already creating functions and arguments

Issue #2633, opened 2023-06-01
New Designer regression for creating built-in connection with parameters dependency


daviburg:
### Describe the Bug with repro steps

'Old designer honors the connection parameters dependencies, e.g. when Authentication Type is Basic then fields SAP Username and Password are shown:

![image](https://github.com/Azure/LogicAppsUX/assets/12040431/1201ecf0-c4dc-4b7a-9185-52fca28f7bf2)

This does not work in new Designer both portal and VS Code:

![image](https://github.com/Azure/LogicAppsUX/assets/12040431/7b697e11-e581-4bd4-9972-d324f6640205)

![image](https://github.com/Azure/LogicAppsUX/assets/12040431/bf988cd9-9a08-448e-96fc-f33cfe052207)

![image](https://github.com/Azure/LogicAppsUX/assets/12040431/66bcdf63-dfff-4f05-92e6-8e412942d975)

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Edge

hartra344:
@rllyy97 can you place take a look at this tomorrow with urgency? Might be hotfix material.

rllyy97:
Just from a quick investigation I'm only seeing this behavior on the SAP in-app connector, all the other connectors I've tried so far are conditionally showing the correct parameters with respect to the auth type selected

rllyy97:
The issue is that the SAP connector is using dependent parameters for the conditionality instead of actual multi-auth, and the dependent parameter keys have a case mismatch, ex: `authenticationType` vs `AuthenticationType`

PR #2635, opened 2023-06-01
fix(Designer): Fixed issue with case-insensitive dependency parameters

rllyy97:
## Main Changes
- On the SAP connector there were connection parameters with constraints for the auth type parameter as a dependent parameter, but there was a key casing mismatch between the two
  - Ex: `AuthenticationType` on the dependency definition vs `authenticationType` on the actual parameter key
  - The in-app SAP connector was the only connector I've been able to see this issue on
- I just switched our dependency logic to be case-insensitive when checking for the dependent parameter
  - Made this a helper function in our utils package for future use

Fixes https://github.com/Azure/LogicAppsUX/issues/2633

### DIFF
--- a/libs/designer-ui/src/lib/createConnection/index.tsx
+++ b/libs/designer-ui/src/lib/createConnection/index.tsx
@@ -187,9 +187,7 @@ export const CreateConnection = (props: CreateConnectionProps): JSX.Element => {
       if (legacyManagedIdentitySelected) return false; // TODO: Riley - Only show the managed identity parameters (which is none for now)
       if (constraints?.hidden === 'true' || constraints?.hideInUI === 'true') return false;
       const dependentParam = constraints?.dependentParameter;
-      if (dependentParam?.parameter) {
-        if (getPropertyValue(parameterValues, dependentParam.parameter) !== dependentParam.value) return false;
-      }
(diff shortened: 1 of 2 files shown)


libs/designer-ui/src/lib/createConnection/index.tsx
```diff
@@ -185,8 +186,9 @@ export const CreateConnection = (props: CreateConnectionProps): JSX.Element => {
         return isServicePrinicipalConnectionParameter(key) && isServicePrincipalParameterVisible(key, parameter);
       if (legacyManagedIdentitySelected) return false; // TODO: Riley - Only show the managed identity parameters (which is none for now)
       if (constraints?.hidden === 'true' || constraints?.hideInUI === 'true') return false;
-      const dependencyParam = constraints?.dependentParameter;
-      if (dependencyParam && parameterValues[dependencyParam.parameter] !== dependencyParam.value) return false;
+      const dependentParam = constraints?.dependentParameter;
+      const currentDependencyValue = getCaseInsensitiveProperty(parameterValues, dependentParam?.parameter);
```
Eric-B-Wu, line 190: we have a function in functions.ts called "getPropertyValue", think we can use that here instead

Eric-B-Wu (approved):

PR #2652, opened 2023-06-02
chore: Production release action auto-creates hotfix branches on minor releases

rllyy97:
# // WORK IN PROGRESS


.github/workflows/production-build.yml
```diff
@@ -136,3 +136,40 @@ jobs:
...
+        with:
+          fallback: 0.0.0
+      - name: Create Branch
+        uses: peterjgrainger/action-create-branch@v2.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.AUTOMATION_PAT }}
+        with:
+          branch: 'hotfix/v${{ steps.previoustag.outputs.tag }}'
+
+  delete-old-hotfixes:
```
preetriti1, line 161: We should only delete old hotfix branches if the release is more than a month old, to keep room for hotfixes if needed or any reverts needed

hartra344: I think a 2 month buffer that this uses should be plenty

rllyy97:
Closing, will revisit later

Issue #2455, opened 2023-05-19
Control graph actions do not allow dragging nodes into/out of their graphs


rllyy97:
![image](https://github.com/Azure/LogicAppsUX/assets/25409734/dee2c2ad-353e-4cf3-9632-bcc1c55029cd)

## What is the issue?
If you have a control graph action (Scope, For Each, Until, Condition, Switch), trying to drag an operation into or out of the graph is prevented.

## Why is this occurring?
This was an intended change that we implemented to prevent some token scope issues.
Dragging nodes in and out of a graph _is_ supported, but disabled temporarily until we can find a proper solution to our token scope issues.
 
## When can I expect this to be fixed?
We are still working on designing a proper solution to this node-moving issue.
We do not have an ETA yet, but will update when we start implementing a solution.

Progress will be updated on this ticket as new information becomes available.

PR #2658, opened 2023-06-05
fix(designer): Allow drag and drop of nodes in/out of scopes

Eric-B-Wu:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes/features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug Fix

* **What is the current behavior?** (You can also link to an open issue here)
Fixes https://github.com/Azure/LogicAppsUX/issues/2455

* **What is the new behavior (if this is a feature change)?**
Able to drag nodes in/out of scope as long as it doesn't move a node that requires a dependent variable below its original output

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

* **Please Include Screenshots or Videos of the intended change**:
dragndrop


rllyy97:
Just adding to this, this only adds protection against dragging nodes above their token dependencies, this does not have any logic over dragging token sources _below_ their dependents

Eric-B-Wu:
> Just adding to this, this only adds protection against dragging nodes above their token dependencies, this does not have any logic over dragging token sources _below_ their dependents

yep, I've added a TODO, but we can create a separate issue for tracking as well

libs/designer/src/lib/core/state/tokens/tokenSelectors.ts
```diff
@@ -0,0 +1,22 @@
...
+  return useSelector(
+    createSelector(getTokenState, (state: TokensState) => {
+      const upstreamNodes = new Set();
+      const nodeTokens = state.outputTokens[id ?? ''];
+      // TODO: Support variables
+      const upstreamNodeIds = nodeTokens?.upstreamNodeIds ?? [];
+
+      upstreamNodeIds.forEach((upstreamNodeId) => {
+        upstreamNodes.add(upstreamNodeId);
+      });
```
rllyy97, line 18: If you're just trying to remove duplicates, instead of making the set obj and building it with `.add` you could just do this
```suggestion
      const nodeTokens = state.outputTokens[id ?? ''];
      // TODO: Support variables
      return [...(new Set(nodeTokens?.upstreamNodeIds ?? []))];
```
We could honestly just make it a `RemoveDuplicates` utils function for arrays

rllyy97 (approved):

libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
```diff
@@ -180,25 +180,50 @@ export const useNodeGraphId = (nodeId: string): string => {
   );
 };
 
-export const useGetAllAncestors = (nodeId: string) => {
+// BFS search for nodeId
+const getChildrenOfNodeId = (childrenNodes: Set<string>, rootNode?: WorkflowNode, nodeId?: string) => {
```
preetriti1, line 184: Why is the nodeId optional when the method explicitly asks to get the children of that?

libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
```diff
@@ -180,25 +180,50 @@ export const useNodeGraphId = (nodeId: string): string => {
...
+  return undefined;
+};
+
+// Adds all childrenIds
+const getAllChildren = (currNode: WorkflowNode, childrenNodes: Set<string>) => {
+  if (currNode.children) {
+    for (const child of currNode.children) {
+      getAllChildren(child, childrenNodes);
+    }
+  } else if (currNode.type === WORKFLOW_NODE_TYPES.OPERATION_NODE) {
```
preetriti1, line 213: How will this work for case nodes?
So if switch is a children of some scope action, wont the children of switch be all subgraphnodes then how do we get for actions inside case nodes?

Eric-B-Wu: This function is just returning the parent node if it's an operational node, and recursing to find all operation nodes within a subgraph node. Because switch node cases are based on the children, it does still calculate to get the operation nodes within a switch case

PR #2695, opened 2023-06-07
fix(designer): Expose get connector for runtime url

bnarayanared:
Exposing the **getConnector** function to access the connector **RuntimeUrl** from Ppux for Hybrid Triggers:
image


.github/pull_request_template.md
```diff
@@ -1,18 +1,15 @@
-* **Please check if the PR fulfills these requirements**
-- [ ] The commit message follows our guidelines
```
preetriti1, line 2: Why is your editor making these changes in these files

libs/designer/src/lib/core/index.ts
```diff
@@ -21,3 +21,4 @@ export { resetWorkflowState } from './state/global';
 export { validateParameter } from './utils/parameters/helper';
 export { getOutputTokenSections, getExpressionTokenSections } from './utils/tokens';
 export { updateParameterValidation } from './state/operation/operationMetadataSlice';
+export { getConnector } from './queries/operation';
```
preetriti1, line 24: Why do you need to expose this from teh designer, eventually all the functions would be exported.
If needed we can expose the keys if you think that is what is missing to query the queryCache which is already exposed from designer object to the host.

I have mentioned this earlier also in a discussion, the query client is already present when you initialize the designer
import { getReactQueryClient } from '@microsoft/logic-apps-designer';
const queryClient = getReactQueryClient();

Add your fetch here, once you have the result you can store it in your host in your cache like we do for one connetor swagger. So just adding a fetch call to read back from your serviec is not a big code change in the host, that way you do not have to rely on designer packages.

See code here - https://msazure.visualstudio.com/One/_git/AzureUX-LogicAppsPortal?path=/src/Extension/Client/React/Services/ApiManagement.ts&version=GBmaster&line=92&lineEnd=97&lineStartColumn=9&lineEndColumn=1&lineStyle=plain&_a=contents

It is better to take control in host for such queries rather make changes in designer for a new package, we do the same in LA Portal for one or two queries. The only thing i can see which might changte are the cacheKey which we can definately export and use the same both in designer and host.

preetriti1 (changes requested):
Also, i think you should fix something in your editor since it has updated so many unneeded files

libs/designer/src/lib/core/index.ts
bnarayanared: Understood; Thanks for the input. Upon debugging it looks like the cacheKey also matches. No changes required to the cacheKey

hartra344:
Closing this since it looks like a better solution was found

PR #2710, opened 2023-06-08
fix(Designer): Connection request errors now show in UI

rllyy97:
## Main Changes
When getting operation connections, if the request failed we didn't show any errors in the UI before, only in the console.
With this PR we show the errors using our existing invalid connection logic.

image

---

I also adjusted some of our connection loading logic.
Currently, when the connection is unable to be found, each time we select the node it will try to refetch.
This made the operation card appear to reload itself each time when it was just the connection loading.
This isn't really needed since the only thing that waits for the connection loading state is the connection display component.
In this PR I moved the connection loading UI state to the Connection Display parameter in the panel instead of on the card itself.
(This also makes the actions appear to load just a bit faster)

image


apps/designer-standalone/src/app/AzureLogicAppsDesigner/DesignerCommandBar.tsx
```diff
@@ -49,14 +52,16 @@ export const DesignerCommandBar = ({
     updateCallbackUrl(designerState, DesignerStore.dispatch);
   });
 
-  const allOperationErrors = useSelector((state: RootState) => {
+  const designerIsDirty = useIsDesignerDirty();
```
ccastrotrejo, line 55: Do we need this as well for vscode?

rllyy97: All of these command bar changes are just for standalone editor, but yes the save button should be disabled when designer isn't dirty

ccastrotrejo: Ok, I can update vscode side

hartra344 (approved):

PR #2713, opened 2023-06-08
feat(vscode): Update logic to disable save button in designer

ccastrotrejo:
### Main code changes
- Update logic/decision to disable save button in vscode designer to follow portal behavior
- Disable sabe button when designer isn't dirty


apps/vs-code-designer-react/src/app/DesignerCommandBar/index.tsx
```diff
@@ -85,7 +86,10 @@ export const DesignerCommandBar: React.FC<DesignerCommandBarProps> = ({ isRefres
     disableGrey: [{ color: 'rgb(121, 119, 117)' }, iconClass],
   });
 
-  const allOperationErrors = useSelector((state: RootState) => {
+  const designerIsDirty = useIsDesignerDirty();
+
+  const allOperationErrors = useSelector((state: RootState) => Object.entries(state.operations.errors) ?? []);
```
rllyy97, line 91: This is where we get the errors from invalid connections, this shouldn't block saving as per Travis' recent message

rllyy97 (changes requested):
We do need the dirty state change, but the connection errors shouldn't be blocking, remove that part and I'll rereview

rllyy97 (approved):

PR #2718, opened 2023-06-09
feat(designer): Add 'search' function for conditional param dropdown

ek68794998:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature.

* **What is the current behavior?** (You can also link to an open issue here)
No ability to search lengthy parameter dropdowns.

image

* **What is the new behavior (if this is a feature change)?**
A search box has been added at the top of the conditional parameters dropdown if there are 4 or more available parameters.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No, unless they rely on element counts/indices with CSS or automation.

* **Please Include Screenshots or Videos of the intended change**:
param-search


libs/designer/src/lib/ui/settings/parameters.tsx
```diff
@@ -0,0 +1,93 @@
+import { updateParameterConditionalVisibility } from '../../core/state/operation/operationMetadataSlice';
+import type { Settings } from './settingsection';
```
preetriti1, line 2: Please dont call it parameters, let's keep it in settings concept for now and add like searchable setting or just a normal utility inside some other folder in designer-ui...
When we do refactoring if we do then we can separate the concepts

preetriti1 (changes requested):
We want this to be present for settings as well, so please remove parameter usage

libs/designer/src/lib/ui/settings/parameters.tsx
ek68794998: Sure. You're talking about just the file name + component name, right?

ek68794998: Actually I see what you're driving at. I can make this more broadly usable.

ek68794998: Okay, I've separated the component logic best I can into the `designer-ui` library.

hartra344 (approved):

Issue #2714, opened 2023-06-08
Unexpected period in action statuses tooltips


daviburg:
### Describe the Bug with repro steps

Hoover over the status of any completed action in a flow run. The tooltip will start with a period. See screenshot below. Probably a typo in the UX code.

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Screenshots or Videos

<img width="266" alt="MicrosoftTeams-image" src="https://github.com/Azure/LogicAppsUX/assets/12040431/565d2667-cbcd-4713-9597-186b07d68e6d">

### Browser

Chrome

AB#24222190

PR #2719, opened 2023-06-09
fix(Designer): Monitoring - StatusPill component `duration` in tooltip 

rllyy97:
## Main Changes
Our `StatusPill` component was not correctly showing the full duration string in the tooltip.
This adds some extra logic to calculate the full duration string from run data.
Fixes https://github.com/Azure/LogicAppsUX/issues/2714

### Old
image

### New
image

### DIFF
--- a/libs/designer-ui/src/lib/utils/utils.ts
+++ b/libs/designer-ui/src/lib/utils/utils.ts
@@ -115,6 +115,19 @@ export function getDurationString(milliseconds: number, abbreviated = true): str
 
 /**
  * Returns a string with a duration, possibly abbreviated, e.g., 15s or 15 second(s)
+ * @arg {string} startTime - The start time of the duration
+ * @arg {string} endTime - The end time of the duration
+ * @arg {boolean} [abbreviated=true] - True if the string should be abbreviated, e.g., "s" instead of "second(s)".
+ */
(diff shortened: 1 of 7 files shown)


ccastrotrejo (approved):

libs/designer-ui/src/lib/utils/utils.ts
```diff
@@ -113,6 +113,19 @@ export function getDurationString(milliseconds: number, abbreviated = true): str
...
 }
 
+/**
+ * Returns a string with a duration, possibly abbreviated, e.g., 15s or 15 second(s)
+ * @arg {string} startTime - The start time of the duration
+ * @arg {string} endTime - The end time of the duration
+ * @arg {boolean} [abbreviated=true] - True if the string should be abbreviated, e.g., "s" instead of "second(s)".
+ */
+export function getDurationStringFromTimes(startTime: string, endTime: string, abbreviated = true): string {
+  const start = new Date(startTime);
```
preetriti1, line 123: Does this handle undefined? For running status it won't have endtime

rllyy97: This function does not, in the component if start or end are undefined we don't call this to get the duration string

PR #2729, opened 2023-06-12
feat(vscode): Func Core Tools - winget

YogiJogiFresh:
**Only tested on Windows - Have not been tested on Linux & Mac**

Feature - Allows the install. update, and uninstall of Function Core Tools with winget.

Current install, update, and uninstall uses brew and npm packagemanagers. The function extension command does not support winget. 
We want support for winget because npm (node) is an optional dependency for Logic Apps vs code extension.

For install - Removed the option to pick the version to install. Always latest.
function core tools can only be update/uninstalled by packagemanager that was used to install it.
getPackageManager - the packagemanager is saved to the global setting platformPackageManager.
The Setting PackageManager is the used as the preferred when using the commands.
Logic Apps Extension Function Core Tools Commands
image

Multiple versions of Function Core Tools (winget & npm)
image

image


apps/vs-code-designer/src/app/utils/funcCoreTools/getFuncPackageManagers.ts
```diff
@@ -2,36 +2,73 @@
...
+
+    // fall through to check npm on windows, mac and linux
+    try {
+      await executeCommand(undefined, undefined, 'npm', '-v');
+      result.push(PackageManager.npm);
+    } catch (error) {
+      // an error indicates no npm
+    }
   }
   return result;
```
mireedmsft, line 73: nit: I would log the result (as well as where we push to that result) as well so we have telemetry if this fails for some reason.

YogiJogiFresh: I think in this case - it's ok not to log as it is checking if npm is installed and adding it to the array if it does.
If there is not npm, there are no "consequences".
"Fall through" to check npm because it is common pakagemanager between Operating Systems.
The nodejs dependency is also supposed to be optional for logic app extension but it is the preferred packagemanager for azure funtions core tools

apps/vs-code-designer/src/app/commands/funcCoreTools/installOrUpdateFuncCoreTools.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getGlobalSetting } from '../../utils/vsCodeConfig/settings';
+import type { IActionContext } from '@microsoft/vscode-azext-utils';
+import { commands } from 'vscode';
+
+export async function installOrUpdateFuncCoreTools(context: IActionContext, packageManagers: PackageManager[]): Promise<void> {
+  ext.outputChannel.show();
+  context.telemetry.properties.installOrUpdateFuncCoreTools = 'true';
+
+  let packageManager = packageManagers ? packageManagers[0] : null;
+  if (packageManager == undefined || packageManager == null) {
```
ccastrotrejo, line 18: NIT: I feel that you can do if(!packageManager). Either null or undefined are falsy statements.

apps/vs-code-designer/src/app/commands/funcCoreTools/installOrUpdateFuncCoreTools.ts
```diff
@@ -0,0 +1,35 @@
...
+    context.telemetry.properties.packageManager = packageManager;
+  }
+
+  switch (packageManager) {
+    case PackageManager.winget:
+      await executeCommand(ext.outputChannel, undefined, 'winget', 'install', '--silent', `${wingetFuncPackageName}`);
+      break;
+    case PackageManager.brew:
+    case PackageManager.npm:
+      await commands.executeCommand(extensionCommand.azureFunctionsInstallOrUpdateFuncCoreTools);
```
ccastrotrejo, line 30: 👏🏽

apps/vs-code-designer/src/app/commands/funcCoreTools/uninstallFuncCoreTools.ts
```diff
@@ -2,10 +2,52 @@
...
+  } else if (packageManagers && packageManagers.length > 1 && context.telemetry.properties.multiFunc) {
+    const placeHolder: string = localize('multipleInstalls', 'Multiple installs of the func cli detected. Select the one to uninstall');
+    const picks: IAzureQuickPickItem<PackageManager>[] = packageManagers.map((pm) => {
+      return { label: localize('uninstall', 'Uninstall {0} package', pm), data: pm };
+    });
+    packageManager = (await context.ui.showQuickPick(picks, { placeHolder, stepName: 'multipleFuncInstalls' })).data;
+  }
+
+  switch (packageManager) {
+    case PackageManager.npm:
```
ccastrotrejo, line 38: Can we not call the azure functions command for the npm and brew? The same way you did for the install

apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsIsLatest.ts
```diff
@@ -49,7 +50,7 @@ export async function validateFuncCoreToolsIsLatest(): Promise<void> {
           const result: MessageItem = await context.ui.showWarningMessage(message, selectUninstall, DialogResponses.dontWarnAgain);
 
           if (result === selectUninstall) {
-            await executeOnFunctions(uninstallFuncCoreTools, context, packageManagers);
+            await executeOnFunctions(uninstallFuncCoreTools, context, context, packageManagers);
```
ccastrotrejo, line 53: It seems that with the update of uninstallFuncCoreTools file, we don't need to wrap it with the executeOnFunctions function, therefore there will no need to add an additional context.

apps/vs-code-designer/src/app/commands/funcCoreTools/installOrUpdateFuncCoreTools.ts
ccastrotrejo: If we are executing in functions extension command lets try to wrap it in the executeOnFunctions function.

apps/vs-code-designer/src/app/utils/funcCoreTools/getFuncPackageManagers.ts
```diff
@@ -2,36 +2,73 @@
...
+      const response = await executeCommand(undefined, undefined, 'winget', 'list', wingetFuncPackageName);
+      // winget response includes headers if package is installed
+      if (response.length > 2) {
+        result.push(PackageManager.winget);
+      }
+    }
+
+    if (process.platform == Platform.mac) {
+      if (
+        isBrewPackageInstalled(getFuncCoreToolsBrewPackageName(FuncVersion.v4)) ||
```
ccastrotrejo, line 33: Does the hasBrew(isFuncInstalled) doesn't do this?

ccastrotrejo (changes requested):
Leaved some comments

apps/vs-code-designer/src/app/commands/funcCoreTools/uninstallFuncCoreTools.ts
YogiJogiFresh: winget is not a supported package manager for func core tools extension, which is why I had to also copy the AzureQuickPickItem. I can't use the extension command because then AzureQuickPickItem prompt will show twice and be redundant.
Scenario: 
2 azure function core tools version installed (installed with winget and npm).
image
image

apps/vs-code-designer/src/app/utils/funcCoreTools/getFuncPackageManagers.ts
YogiJogiFresh: true

PR #2733, opened 2023-06-13
fix(designer): dictionary and array editor collapsed views erroring

### LINKED ISSUE #2699: Dictionary editor shows invalid json when content in value field contains new lines

1. Add any action with dictionary editor.
2. In the value field start entering some tokens and text with new line.

Actual: It shows invalid parameters although it can be stringified

Standard (Portal)

Generally Available

Yes

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

Any

### Additional context

Incident - https://portal.microsofticm.com/imp/v3/incidents/details/395701603/home

AB#24208544

Eric-B-Wu:

bugfix

Adding a new line in dictionary and array editor would result in validation errors

1. New lines no longer cause validation errors in json objects
2. Array editor doesn't crash designer when serializing in collapsed view

bugfix

bugfix2

Fixes #2699


libs/designer-ui/src/lib/arrayeditor/collapsedarray.tsx
```diff
@@ -23,7 +23,6 @@ export interface CollapsedArrayProps {
 
 export const CollapsedArray = ({
   labelProps,
-  isValid = true,
```
Eric-B-Wu, line 26: here we were doing validation checks within the editor as well as the parameter validation to enable/disable the toggle button. Removed this so there would never be the case of 2 error messages showing.

PR #2734, opened 2023-06-13
fix(Designer): Consumption - Adding swagger function actions

rllyy97:
## Main Changes
- Added support for swagger function actions
  - This has been added as a separate operation from the normal function actions
  - The swagger action selection and de/serialization steps are very different, instead of over-complicating the existing logic that we use for all actions, this seemed like the cleaner solution
  - The only difference between Legacy and Preview will be during addition you will need to select the swagger or default function operation (instead of there being two tabs in function selection)
  - This maintains similar functionality to App service and Http+Swagger where you can select the endpoint after adding the operation
- Added search bar for azure resource selection component
- Added `SwaggerClient` as a Designer option along with the existing `SchemaClient` and `ValuesClient`
  - These swagger function actions do not currently store the swagger definition url in the operation, so they have to be retrieved on initilization/addition
  - Instead of hardcoding this behavior I've opted to make this new client so that it can be manifest-based and reused elsewhere
  - This was also done in anticipation of a rework for the consumption APIM operations, they ran into a similar issue and are currently hard-coded but I would like to convert them to use this client as well
---
Sibling Portal PR: https://msazure.visualstudio.com/One/_git/AzureUX-LogicAppsPortal/pullrequest/8269276

---
image
image
image


libs/services/designer-client-services/src/lib/consumption/manifests/swaggerFunctions.ts
```diff
@@ -0,0 +1,157 @@
...
+      //     },
+      //   },
+      //   parameters: {
+      //     type: 'object',
+      //     functionAppId: {
+      //       parameterReference: 'functionApp.id',
+      //       required: true,
+      //     },
+      //   },
+      // },
```
rllyy97, line 150: This is the format we use for the other schema/values clients, I initially tried to match them 1:1 but we use this data pre-parsing the parameters, so the info we normally have for the other clients we don't have yet (in getCustomSwaggerIfNeeded), wasn't really sure how to proceed so I'm just passing the whole definition instead of specific parameters

apps/designer-standalone/src/app/AzureLogicAppsDesigner/Utilities/Workflow.ts
```diff
@@ -37,7 +37,7 @@ export class WorkflowUtility {
       };
     }
 
-    const functionConnectorId = '/connectionProviders/azureFunctionOperation';
```
preetriti1, line 40: Isnt the azure function connector this one for standard?

libs/designer-ui/src/lib/azureResourcePicker/index.tsx
```diff
@@ -39,20 +49,46 @@ export const AzureResourcePicker = (props: AzureResourcePickerProps) => {
...
+  const searchFilteredResources = useMemo(() => {
+    return resources.filter((resource) => searchResourceNames.includes(resource.name));
+  }, [resources, searchResourceNames]);
+
   return (
     <div style={{ width: '100%' }}>
       <Label className="label" required>
         {titleText}
       </Label>
+      <SearchBox placeholder={searchText} onChange={(e, newValue) => setSearchTerm(newValue ?? null)} />
```
preetriti1, line 91: this is great, we should also have a load more kind of UI, since sometimes the resources can be big list

libs/designer-ui/src/lib/azureResourcePicker/index.tsx
```diff
@@ -39,20 +49,46 @@ export const AzureResourcePicker = (props: AzureResourcePickerProps) => {
...
+      ),
+    [resourceNames]
+  );
+  const searchResourceNames = useMemo(() => {
+    if (!searchTerm) return resourceNames;
+    return fuseObject.search(searchTerm).map((result) => result.item.id);
+  }, [resourceNames, fuseObject, searchTerm]);
+
+  const searchFilteredResources = useMemo(() => {
+    return resources.filter((resource) => searchResourceNames.includes(resource.name));
```
preetriti1, line 83: will this do a case insensitive search?

libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
```diff
@@ -433,15 +439,26 @@ export const updateAllUpstreamNodes = (state: RootState, dispatch: Dispatch): vo
 };
 
 export const getCustomSwaggerIfNeeded = async (
+  operationId: string,
```
preetriti1, line 442: INstead of passing this on, we can have a metadata inside customSwagger of manifest to specify whether this is apim or function, so the function getDynamicSwaggerUrl can do it based on that.
We have something similar for connection creation inside manifest which says the conection is apim or function

libs/designer/src/lib/core/utils/connectors/connections.ts
```diff
@@ -83,7 +83,9 @@ export function getAssistedConnectionProps(connector: Connector, manifest?: Oper
       description: 'Label for function app selection',
     });
 
-    const getColumns = (functionApp: any) => [functionApp?.name, functionApp?.properties?.resourceGroup, functionApp?.location];
+    const getColumns = (functionApp: any) => [getObjectName(functionApp), functionApp?.properties?.resourceGroup, functionApp?.location];
+
+    const getSubResourceName = (azureFunction: any) => getObjectName(azureFunction).split('/').slice(1).join('/');
```
preetriti1, line 88: can you give an example here?
are we just getting the function name from function app? 

And also we should be handling for null refs here

libs/services/designer-client-services/src/lib/base/connector.ts
```diff
@@ -77,10 +79,10 @@ export abstract class BaseConnectorService implements IConnectorService {
     const configuration = await getConfiguration(connectionId ?? '');
 
     if (this._isClientSupportedOperation(connectorId, operationId)) {
-      if (!this.options.valuesClient[dynamicOperation]) {
+      if (!this.options.valuesClient?.[dynamicOperation]) {
```
preetriti1, line 82: Can you make the schema and values client optional then?

libs/services/designer-client-services/src/lib/base/function.ts
```diff
@@ -53,4 +54,119 @@ export class BaseFunctionService implements IFunctionService {
     });
     return keysResponse?.default ?? 'NotFound';
   }
+
+  public async fetchApiDefinitionUrl(functionAppId: string) {
+    const response = await this.options.httpClient.get<any>({
+      uri: `https://management.azure.com/${functionAppId}/config/web`,
```
preetriti1, line 60: Do not hard code the uri, we should get it from baseurl

libs/services/designer-client-services/src/lib/base/function.ts
```diff
@@ -53,4 +54,119 @@ export class BaseFunctionService implements IFunctionService {
...
+      queryParameters: { 'api-version': this.options.apiVersion },
+    });
+    if (!response?.properties?.apiDefinition?.url) {
+      throw new Error('ApiDefinitionUrl not found');
+    }
+    return response?.properties?.apiDefinition?.url ?? '';
```
preetriti1, line 66: lets not default to empty string

libs/services/designer-client-services/src/lib/base/operationmanifest.ts
```diff
@@ -112,7 +112,7 @@ const xslttransform = 'xslttransform';
 const datamapper = 'datamapper';
 
 export const apiManagementConnectorId = '/connectionProviders/apiManagementOperation';
-export const azureFunctionConnectorId = '/connectionProviders/azureFunctionOperation';
```
preetriti1, line 115: I am not sure if backend is using this connectorId, please double check on what standard service is returning

libs/services/designer-client-services/src/lib/connector.ts
```diff
@@ -83,6 +83,14 @@ export interface IConnectorService {
     nodeMetadata: any
   ): Promise<OpenAPIV2.SchemaObject>;
 
+  /**
+   * Gets the dynamic swagger url in manifest based operations.
+   * @arg {string} operationId - The operation id.'
+   * @arg {any} defintion - The swagger definition.
+   * @return {Promise<string>}
+   */
+  getDynamicSwaggerUrl(connectorId: string, operationId: string, dynamicOperationId: string, defintion: string): Promise<string>;
```
preetriti1, line 92: Can't we implement this in azurefunction and azureapim service and based on the manifest property we can call those services... since we already have separate services for them

libs/services/designer-client-services/src/lib/consumption/manifests/swaggerFunctions.ts
```diff
@@ -0,0 +1,157 @@
...
+    description: 'Show Azure Swagger Functions in my subscription',
+
+    environmentBadge: coreBadge,
+
+    inputs: {
+      type: 'object',
+      properties: {
+        functionApp: {
+          type: 'object',
+          'x-ms-visibility': 'hideInUI',
```
preetriti1, line 35: Do we have to set this at object level also?

libs/utils/src/lib/helpers/functions.ts
```diff
@@ -937,3 +937,12 @@ export function findAncestorElement(
...
   return invalidChars.some((invalidChar) => includes(str, invalidChar));
 }
+
+/**
+ * Returns the closest thing to a "name" on an object.
+ * @arg {any} obj - An object to get the name of.
+ * @return {string} - The name of the object.
+ */
+export function getObjectName(obj: any): string {
+  return obj?.displayName ?? obj?.properties?.displayName ?? obj?.name ?? obj?.properties?.name ?? obj?.id ?? '';
```
preetriti1, line 947: Why do we want to show a resource display name when selecting a resource.... the Name column specifically means resource name, if we want to show a friendly name it should be a separate column..
And dont call it object since it looks like we are looking for objectKey like json object key

libs/utils/src/lib/models/operationmanifest.ts
```diff
@@ -193,7 +193,8 @@ export interface OperationManifestProperties {
   };
 
   customSwagger?: {
-    location: string[];
+    location?: string[];
+    operationId?: string;
```
preetriti1, line 197: We can add here -
location?: ...
service: {
  name: ENUM [ApiManagement, AzureFunction]
  operationId: getSwaggerUrl [string;]
 }

preetriti1 (changes requested):

libs/designer-ui/src/lib/azureResourcePicker/index.tsx
rllyy97: The <List/> fluent component is virtualized, so it automatically does render paging under the hood, no real need for an explicit "load more"

libs/designer/src/lib/core/utils/connectors/connections.ts
rllyy97: This can be reverted, it was for back when I was showing both swagger functions in this component, the swagger names had a different format

libs/services/designer-client-services/src/lib/base/connector.ts
rllyy97: Yes I can do that

libs/services/designer-client-services/src/lib/base/function.ts
rllyy97: Should we throw an error if it's not present?

libs/services/designer-client-services/src/lib/consumption/manifests/swaggerFunctions.ts
rllyy97: I'm not sure I understand what you mean `at object level`

preetriti1: You have hideinui set for both object and and its leaf nodes, is it needed at the object level is my question

libs/designer/src/lib/core/utils/connectors/connections.ts
rllyy97: Update on this, the function name value is `{appName}/{functionName}`, the extra split and slice was just to remove the app name since some are pretty long and it's already going to be known since they had to click on it

CHANGELOG.md
```diff
@@ -1,98 +1,89 @@
...
 
+## [2.26.0](https://github.com/Azure/LogicAppsUX/compare/v2.25.0...v2.26.0) (2023-06-08)
 
 ### Features
 
-* **Data Mapper:** Functions will remain during navigation ([#2665](https://github.com/Azure/LogicAppsUX/issues/2665)) ([d9ef24c](https://github.com/Azure/LogicAppsUX/commit/d9ef24c9e66defae2ee15a72914dbfaef052721c))
-* **Designer:** Browse View - Hiding connectors with no valid actions/triggers ([#2654](https://github.com/Azure/LogicAppsUX/issues/2654)) ([2aa81af](https://github.com/Azure/LogicAppsUX/commit/2aa81afd82b5e14617bdd36a79668c48d87725d4))
-* **designer:** Initial setup work of the chatbot ([#2691](https://github.com/Azure/LogicAppsUX/issues/2691)) ([aaf33c0](https://github.com/Azure/LogicAppsUX/commit/aaf33c0972a61ac173887b72534faa719aafea9a))
-
+- **Data Mapper:** Functions will remain during navigation ([#2665](https://github.com/Azure/LogicAppsUX/issues/2665)) ([d9ef24c](https://github.com/Azure/LogicAppsUX/commit/d9ef24c9e66defae2ee15a72914dbfaef052721c))
```
preetriti1, line 7: The changelog is getting updated to remove * to -, will this cause any change in how it is viewed in vscode? Can we try to not include this in the PR

libs/services/designer-client-services/src/lib/base/function.ts
```diff
@@ -53,4 +54,120 @@ export class BaseFunctionService implements IFunctionService {
     });
     return keysResponse?.default ?? 'NotFound';
   }
+
+  public async fetchSwaggerUrl(functionAppId: string) {
```
preetriti1, line 58: Can we just expose fetchFuntionAppSwagger and let the swaggerparser be returned in customSwagger so we dont have to do two separate calls in teh caller?

libs/services/designer-client-services/src/lib/function.ts
```diff
@@ -4,6 +4,9 @@ export interface IFunctionService {
   fetchFunctionApps(): Promise<any>;
   fetchFunctionAppsFunctions(functionAppId: string): Promise<any>;
   fetchFunctionKey(functionId: string): Promise<any>;
+  fetchSwaggerUrl(functionAppId: string): Promise<string>;
```
preetriti1, line 7: See if we can just expose fetchFunctionSwagger from here

libs/utils/src/lib/helpers/functions.ts
```diff
@@ -937,3 +937,12 @@ export function findAncestorElement(
 export function hasInvalidChars(str: string, invalidChars: string[]): boolean {
   return invalidChars.some((invalidChar) => includes(str, invalidChar));
 }
+
+/**
+ * Returns the closest thing to a "name" on an object.
```
preetriti1, line 942: nit: on an azure resource

preetriti1 (changes requested):

CHANGELOG.md
rllyy97: Yeah I can revert this, this keeps happening automatically in my PRs and I'm really not sure why

Eric-B-Wu: I think this is a prettier setting, both do the same thing in markdown of creating an unordered list, but I think we should change our release automation to make it so that it uses - instead of * so it aligns with prettier (or we can change or prettier setting)

libs/services/designer-client-services/src/lib/base/function.ts
```diff
@@ -53,4 +54,120 @@ export class BaseFunctionService implements IFunctionService {
...
+    const paths = swagger.api.paths[operation.path];
+    const rawOperation = paths[operation.method];
+    const schema = { type: 'object', properties: {} as any, required: [] as string[] };
+
+    if (isInput) {
+      const baseUrl = swagger.api.host
+        ? swagger.api.schemes?.length
+          ? `${swagger.api.schemes.at(-1)}://${swagger.api.host}`
+          : `http://${swagger.api.host}`
+        : swagger?.api?.basePath ?? 'NotFound';
```
preetriti1, line 99: this is not including the basepath

libs/services/designer-client-services/src/lib/base/operationmanifest.ts
```diff
@@ -411,6 +412,12 @@ export function getBuiltInOperationInfo(definition: any, isTrigger: boolean): Op
         operationId: isTrigger ? appservicetrigger : appservice,
       };
 
+    case function_:
+      return {
+        connectorId: azureFunctionConnectorId,
+        operationId: definition?.inputs?.uri ? azureswaggerfunction : azurefunction,
+      };
+
```
preetriti1, line 420: Move this to consumption/operationmanifest.ts

libs/services/designer-client-services/src/lib/consumption/operations/functions.ts
```diff
@@ -15,19 +15,21 @@ export const functionGroup = {
   },
 };
 
+const api = {
+  id: '/connectionProviders/azureFunctionOperation',
```
preetriti1, line 19: Get these from operationgroup.ts in this folder so we have only one place to define the connector details. See how it is done for xslt and other IA operations

preetriti1 (changes requested):

libs/services/designer-client-services/src/lib/base/function.ts
rllyy97: What do you mean by including the basepath

libs/services/designer-client-services/src/lib/base/operationmanifest.ts
```diff
@@ -411,6 +411,12 @@ export function getBuiltInOperationInfo(definition: any, isTrigger: boolean): Op
         operationId: isTrigger ? appservicetrigger : appservice,
       };
 
+    case function_:
```
preetriti1, line 414: Revert changes in this file, if there is kind or any other complicated check for connectorid and operationId it is in this switch case else it is in the dictionary below

libs/services/designer-client-services/src/lib/base/function.ts
preetriti1: this chunk of code is not getting this - 

schemes + api.host + api.basePath

YOu can either have baseUrl as the above or just have baseUrl as schemes + api.host and in the below comment have 
default: api.basePath ? baseurl + api.basepath + operation.path : baseurl + operation.path
See appservices.ts

preetriti1 (approved):

PR #2745, opened 2023-06-14
fix(Designer): Add missing quotes around reference key for hybrid triggers

bnarayanared:
**Fix 1:**
The connection reference format for hybrid trigger seems to be missing the quotes around the reference key:
image

**Fix2:**
Currently on the v3 designer, for a manifest based operation, we default the splitOn value to be added as `"@triggerOutputs()['<property_path>']",` defaulted here:

https://github.com/Azure/LogicAppsUX/blob/c2e7189c55629b61d0f2608cdf5a3d4c75af1f1c/libs/designer/src/lib/core/actions/bjsworkflow/settings.ts#L525

 

But for hybridTriggers, it would need to be `"@triggerBody()['<property_path>']",`
image

AB#24273174

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -644,7 +644,7 @@ const serializeHost = (
       return {
         host: {
           connection: {
-            name: "@parameters('$connections')[" + referenceKey + "]['connectionId']",
+            name: "@parameters('$connections')['" + referenceKey + "']['connectionId']",
           },
         },
       };
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -644,7 +644,7 @@ const serializeHost = (
       return {
         host: {
           connection: {
-            name: "@parameters('$connections')['" + referenceKey + "']['connectionId']",
+            name: `@parameters('$connections')['${referenceKey}']['connectionId']`,
           },
         },
       };
--- a/libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
@@ -521,7 +521,10 @@ const getSplitOnValue = (
         // TODO (3727460) - Consume top level required fields when available here.
         const { alias, propertyName, required } = getSplitOnArrayAliasMetadata(manifest.properties.outputs, /* propertyRequired */ true);
         const propertyPath = alias || propertyName;
-        if (propertyPath) {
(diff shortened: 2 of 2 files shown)


preetriti1:
@bnarayanared - The second one is not true, if your outputs are correctly set then the splitON should be triggerOutputs()['body']['rows']
So please fix the outputs in the manifest.

libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
```diff
@@ -521,7 +522,10 @@ const getSplitOnValue = (
         // TODO (3727460) - Consume top level required fields when available here.
         const { alias, propertyName, required } = getSplitOnArrayAliasMetadata(manifest.properties.outputs, /* propertyRequired */ true);
         const propertyPath = alias || propertyName;
-        if (propertyPath) {
+        if (propertyPath && (manifest.properties.connectionReference?.referenceKeyFormat !== undefined && manifest.properties.connectionReference?.referenceKeyFormat === ConnectionReferenceKeyFormat.HybridTrigger)) {
+          return `@${Constants.TRIGGER_BODY_OUTPUT}[${convertToStringLiteral(propertyPath)}]`;
+        }
+        else if (propertyPath) {
```
preetriti1, line 528: revert and make sure you have correct outputs in manifest rather updating like this for a specific trigger.

libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -644,7 +644,7 @@ const serializeHost = (
       return {
         host: {
           connection: {
-            name: "@parameters('$connections')[" + referenceKey + "]['connectionId']",
+            name: "@parameters('$connections')['" + referenceKey + "']['connectionId']",
```
preetriti1, line 647: This could be written as 
name: `@parameters('$connections')['${referenceKey}']['connectionId']`

Just a headsUp, this might change pretty soon since LA and PA will be using same connectionformat for open api and this might not look exactly same for all hosts.

preetriti1 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
bnarayanared: Could you suggest what needs to be the correct output format for adding the SplitOn:
image

bnarayanared:
> @bnarayanared - The second one is not true, if your outputs are correctly set then the splitON should be triggerOutputs()['body']['rows'] So please fix the outputs in the manifest.

I have attached the output manifest I'm using but this doesn't set the SplitOn to be triggerOutputs()['body']['rows']

libs/designer/src/lib/core/actions/bjsworkflow/settings.ts
preetriti1: Why are you adding lines 117-123? Body property is already marked as dynamic properties, so initially you won't have splitonset because outputs do not contain an array to set as spliton options.

Once you have dynamic properties call made, please debug to check what does the outputs in updateoutputsandtokens contain.
Please debug the code to see where it is going wrong

PR #2759, opened 2023-06-15
feat(chatbot): Chatbot input and user message component

wangjcynthia:

Adds user input and user message components for the Logic Apps Copilot chatbot
Chatbot panel is empty
Chatbot panel now shows text input that allows user to type out a message. When 'Enter' key or the 'Send' button is pressed, the message appears in a chat bubble in the chat.
Does not affect any of our production designers, just introduces it in standalone

https://github.com/Azure/LogicAppsUX/assets/64856290/8dd1ddc2-fb3b-4fdb-9677-635faa609add


hartra344 (approved):

libs/chatbot/src/lib/ui/chatbot.tsx
```diff
@@ -1,23 +1,217 @@
...
+              setInputQuery(newValue ?? '');
+            }}
+          ></ChatInput>
+        </div>
+      </div>
     </Panel>
   );
 };
+
+const useStyles = makeStyles({
```
Eric-B-Wu, line 138: We should avoid using component styling - it causes our the CSS to be loaded in runtime, which can cause performance issues: https://github.com/microsoft/fluentui/wiki/Component-Styling

libs/chatbot/src/lib/ui/styles.less
```diff
@@ -0,0 +1 @@
+@import '../../../../designer-ui/src/styles.less';
```
Eric-B-Wu, line 1: we shouldn't be importing the designer-ui less files, we should instead be importing any designer-ui components

libs/designer-ui/src/lib/chatbot/chatBubble.tsx
```diff
@@ -0,0 +1,181 @@
...
+          </div>
+        </div>
+      )}
+    </div>
+  );
+};
+
+const USER_MESSAGE_CLASS = 'is-user-message';
+
+const footerButtonStyles: IButtonStyles = {
```
Eric-B-Wu, line 94: move from styled component to less file

libs/designer-ui/src/lib/chatbot/chatResources.ts
```diff
@@ -0,0 +1,163 @@
+export interface ChatResources {
```
Eric-B-Wu, line 1: are we planning to be using all these props? We should avoid putting logic we're currently not using, cause this makes it harder to find during cleanup

libs/designer-ui/src/lib/chatbot/thumbsReactionButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+      toggle={true}
+      checked={isVoted}
+    />
+  );
+};
+
+const getIconButtonStyles = (isVoted?: boolean) => {
+  const theme = getTheme();
+  return {
+    root: { color: theme.palette.neutralPrimary, backgroundColor: 'transparent !important' },
```
Eric-B-Wu, line 34: Should avoid using !important

libs/designer-ui/src/lib/styles.less
```diff
@@ -68,7 +68,7 @@
 @import './azureResourcePicker/azureResourcePicker.less';
 @import './staticResult/staticResult.less';
 @import './dynamicallyaddedparameter/dynamicallyaddedparameter.less';
-
+@import (reference) './themes.less';
```
Eric-B-Wu, line 71: do we need this?

wangjcynthia: oops sorry forgot to remove that, I was using it before but not anymore

libs/designer-ui/src/lib/chatbot/chatResources.ts
wangjcynthia: yeah i'll remove the ones i'm not using right now, and then add them in as needed

libs/chatbot/src/lib/ui/chatbot.tsx
```diff
@@ -1,23 +1,217 @@
-import { Panel, PanelType } from '@fluentui/react';
-import { PanelLocation } from '@microsoft/designer-ui';
-import { useState } from 'react';
+import './styles.less';
```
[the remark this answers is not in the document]
wangjcynthia: i moved the styling into the less file now, does it still not need to be imported? i don't see the styles applied otherwise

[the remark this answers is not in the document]
Eric-B-Wu: what we should be doing is using a parent less file in chatbot folder and importing the less file there, that way we separate the less files and the components

libs/chatbot/src/lib/ui/chatbot.tsx
```diff
@@ -1,23 +1,133 @@
...
+      layerProps={{ styles: { root: { zIndex: 0, display: 'flex' } } }}
     >
-      <p>Content goes here.</p>
+      <div className={'chatbot-container'}>
+        <div className={'chatbot-header'}>
+          {/*TODO: Add icon for header*/}
+          <div className={'chatbot-header-title'}>{intlText.headerTitle}</div>
+          <div className={'chatbot-header-mode-pill'}>{intlText.pill}</div>
+          <IconButton
+            title={'Close'}
```
Eric-B-Wu, line 88: can we localize this?

libs/chatbot/src/lib/ui/chatbot.tsx
```diff
@@ -1,23 +1,133 @@
...
       isOpen={!collapsed}
-      customWidth={'300px'}
+      customWidth={'340px'}
       hasCloseButton={false}
       isBlocking={false}
-      layerProps={{ styles: { root: { zIndex: 0 } } }}
+      layerProps={{ styles: { root: { zIndex: 0, display: 'flex' } } }}
     >
-      <p>Content goes here.</p>
+      <div className={'chatbot-container'}>
```
Eric-B-Wu, line 82: doesn't have to this PR, but just wanted to let you know for class names we tend to prefix it with msla-, which makes it so we can track which are our custom components.

Eric-B-Wu (approved):

PR #2765, opened 2023-06-16
fix(designer): StandardRunService to properly include access token in headers

wifun:

The `accessToken` option for StandardRunService didn't work because the "Headers" object (evaluates to empty object) from the fetch API is not compatible with the IHttpClient interface (expecting a Record<string, string>.

Using a constant is also problematic if the accessToken were to expire.

Instead changing the option to accept a function to fetch the access token.

Yes changing accessToken option from string to fn is breaking

### DIFF
--- a/apps/vs-code-react/src/app/overview/app.tsx
+++ b/apps/vs-code-react/src/app/overview/app.tsx
@@ -59,7 +59,7 @@ const OverviewApp: React.FC<AppProps> = ({ workflowProperties, apiVersion, baseU
     return new StandardRunService({
       baseUrl,
       apiVersion,
-      accessToken,
+      getAccessToken: () => Promise.resolve(accessToken),
       workflowName: workflowProperties.name,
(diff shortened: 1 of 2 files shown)


hartra344 (changes requested):
Generally this should be done in the HttpClint. Not an individual service.

apps/vs-code-react/src/app/overview/app.tsx
```diff
@@ -59,7 +59,7 @@ const OverviewApp: React.FC<AppProps> = ({ workflowProperties, apiVersion, baseU
     return new StandardRunService({
       baseUrl,
       apiVersion,
-      accessToken,
+      getAccessToken: () => Promise.resolve(accessToken),
```
preetriti1, line 62: This should only be set when accessToken is not undefined

preetriti1 (changes requested):

Issue #2760, opened 2023-04-12
IBM host files action : Running trigger failed by showing server failure message.


vgouth:
### Describe the Bug with repro steps

1.Create a new workflow app using VS Code.
2.Create a stateful workflow
3.Add a HTTP trigger to the workflow and save the workflow
4.Add a "Generate Host File Contents" (from the IBM Host File SP) Action after the trigger
5.The Connection Parameters can (should) be left unchanged - or, if you like: 37 and unchecked.
6.For the first 2 parameters of the Action, enter: redefines.hidx (for the HIDX Name), "CUSTOMER_FILE_ROW" (for the Schema Name).
7.Open the attachment named InputJson.txt, copy the contents. In VS Code - the third Parameter Rows, has a button on the right enabling you to fill out all Input as a single value, click this. Delete the contents of the text box ("[ null ]" or similar) and then paste in the copied Input.Json contents.
8.
Add a "Parse Host File Contents" Action, and fill out the first 2 Parameters as above (step 6). For the "Binary Contents" select the Dynamic content "Binary Contents" (which is the output of the previous step).
9.Add a response action as well - Http Response - for the Body, use the Dynamic content "Rows" (the output from the previous step).
10.Click on "Artifacts/Maps", right click - reveal in file explorer. Download the attachment "Redefines.hidx" into that directory.
11.Press F5/Run the Workflow
12. Get the trigger URL by going to the "overview" page for the workflow.
13. **Expected Result :** Run trigger should succeed.
**Actual Result :** Run trigger getting failed by showing **server failure message**

### Workflow JSON

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "Parse_Host_File_Contents": {
                "type": "ServiceProvider",
                "inputs": {
                    "parameters": {
                        "hidx": "redefines.hidx",
                        "schema": "CUSTOMER_FILE_ROW",
                        "contents": "@body('Generate_Host_File_Contents')?['contents']"
                    },
                    "serviceProviderConfiguration": {
                        "connectionName": "hostfile",
                        "operationId": "parseFileContents",
                        "serviceProviderId": "/serviceProviders/hostfile"
                    }
                },
                "runAfter": {
                    "Generate_Host_File_Contents": [
                        "SUCCEEDED"
                    ]
                }
            },
            "Response": {
                "type": "Response",
[...]

PR #2776, opened 2023-06-19
fix(desinger): Simple Array Editor support for non-string types in expanded mode

Eric-B-Wu:

bug fix
it would try to convert non-string types to a string and once serialized it would default to collapsed view with quotes.
https://github.com/Azure/LogicAppsUX/issues/2760
Should only add quotes when the type is of string, and should now parse properly
no
simpleArrayEditorNonString


libs/designer-ui/src/lib/arrayeditor/index.tsx
```diff
@@ -85,8 +85,9 @@ export const ArrayEditor: React.FC<ArrayEditorProps> = ({
   useEffect(() => {
     arrayType === ArrayType.COMPLEX
       ? initializeComplexArrayItems(initialValue, itemSchema, setItems, setIsValid, setCollapsed)
-      : initializeSimpleArrayItems(initialValue, setItems, setIsValid, setCollapsed);
-  }, [arrayType, initialValue, itemSchema]);
+      : initializeSimpleArrayItems(initialValue, itemSchema.type, setItems, setIsValid, setCollapsed);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
preetriti1, line 90: Why no dependencies?

Eric-B-Wu: this was being called onBlur, so if the user typed an invalid value type in the expanded array, it would be converted to collapsedArray automatically. Instead it should now only do it on mount

preetriti1 (approved):

PR #2780, opened 2023-06-19
fix(designer): Prevent panel from closing if `intl` is updated

ek68794998:

Bug fix.

If there is a situation where the designer is re-rendered but nothing has changed, the panel will collapse since #2722. This is because `intl` gets updated by `IntlProvider` and re-runs the `useEffect` in `panelroot.tsx` which causes the panel to collapse.

Designer re-renders with no effective changes will no longer close the panel. By memoizing `onError`, `IntlProvider` does not update `intl` when a re-render occurs.

No.

Can provide if needed, but hopefully the above is descriptive enough.


wifun (approved):

libs/designer/src/lib/core/DesignerProvider.tsx
```diff
@@ -39,6 +40,15 @@ export const DesignerProvider = ({ locale = 'en', options, children }: DesignerP
   const azTheme = !isDarkMode ? AzureThemeLight : AzureThemeDark;
   const webTheme = !isDarkMode ? webLightTheme : webDarkTheme;
   const themeName = useMemo(() => (!isDarkMode ? 'light' : 'dark'), [isDarkMode]);
+  const onError = useMemo<OnIntlErrorFn>(
```
hartra344, line 43: can we make this 
```suggestion
  const onError = useCallback<OnIntlErrorFn>(
```

useCallback is similar to useMemo but is made specifically for this use case.

ek68794998: You're right, I totally forgot that function existed. Will do.

preetriti1 (approved):

hartra344 (approved):

Issue #2762, opened 2023-06-15
Parameters should show loading when parameters are still being loaded


preetriti1:
![Image](https://github.com/Azure/LogicAppsUX/assets/48265693/4374479b-c52a-4251-9ebc-3ca6acb2488c)

It incorrectly shows there are no parameters when parameters are still loading or in error state.

PR #2785, opened 2023-06-19
fix(Designer): Added conditions for Parameters panel loading state

rllyy97:
## Main Changes
- Added a section for loadStatus flags in operation metadata redux
  - Mainly for the deserialization `initializeNodes` status, to check if we ran the initialization for parameter panel loading
  - Made it a separate section on that slice so that we can add to it later if the need arises
- Also added a check for dynamic inputs load status in state but this doesn't appear to be hooked up yet
- Hides the `no parameters` message on dynamic inputs error state (could be that the operation has parameters that just didn't load)

Fixes https://github.com/Azure/LogicAppsUX/issues/2762


libs/designer/src/lib/core/state/operation/operationMetadataSlice.ts
```diff
@@ -119,6 +124,9 @@ const initialState: OperationMetadataState = {
   staticResults: {},
   repetitionInfos: {},
   errors: {},
+  loadStatus: {
+    nodesInitialized: false,
```
hartra344, line 128: A little orthogonal to this PR itself but can we add some logging to time how long after load starts to this switching to true? I think that would be useful data to have.

hartra344 (approved):

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
```diff
@@ -90,7 +98,7 @@ export const ParametersTab = () => {
           {errorInfo.message}
         </MessageBar>
       ) : null}
-      {!hasParametersToAuthor(inputs?.parameterGroups ?? {}) ? (
+      {!errorInfo && !hasParametersToAuthor(inputs?.parameterGroups ?? {}) ? (
```
preetriti1, line 101: We should not be checking for warning errors, for example dynamic outputs load error should not impact inputs loading

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/panelTabs/parametersTab/index.tsx
rllyy97: This block doesn't impact the actual inputs, it's just the message bar that appears when there are no inputs, but I get what you're saying and I can adjust to be more specific in the errors that trigger it.  I mainly was trying to catch when we have errors around getting dynamic inputs and possibly being empty as an error state instead of intentionally

hartra344 (approved):

preetriti1 (approved):

PR #2786, opened 2023-06-19
fix(designer): Parameters with visibility "internal" should hide in UI on flow load

shuktika12163:
Parameters marked as "internal" are not part of "availableParameters" but are part of the remaining parameters. They should be part of available parameters since they are defined in manifest

Bug fix

Parameter marked as "internal" are visible in UI on flow load

Parameter marked as "internal" will not be visible in UI on flow load

I tested by hardcoding visibility of "value" parameter to "internal" for Initialize variable action.
With the fix, the parameter doesn't show on flow load:
image

Without the fix, the parameter shows on flow load:
image


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1193,7 +1193,7 @@ export function updateParameterWithValues(
                     isUnknown: true,
                   };
 
-                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ false));
+                  parameters.push(transformInputParameter(restInputParameter, propertyValue, /* invisible */ true));
```
preetriti1, line 1196: Any manifest defined parameter should never reach here, there must be somewhere before this code..

shuktika12163: Updated SchemaProcessorOptions to include internal instead. That function is only called for getInputParametersFromManifest where the issue was happening earlier.

preetriti1 (approved):

PR #2787, opened 2023-06-19
fix(Designer): Add connectionReferenceKeyFormat for hybrid triggers for deserialization

bnarayanared:
1. For deserialization of workflow, adding the HybridTrigger case to extract the connectionReferenceKey
2. For adding the SplitOn value, adding the hybridTrigger connectionReferenceKey type to get the AliasPathParsing

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
@@ -408,6 +408,9 @@ function getConnectionReferenceKeyForManifest(referenceFormat: string, operation
 
     case ConnectionReferenceKeyFormat.OpenApi:
       return getOpenApiConnectionReferenceKey((operationDefinition as LogicAppsV2.OpenApiOperationAction).inputs);
+
+    case ConnectionReferenceKeyFormat.HybridTrigger:
+      return getHybridTriggerConnectionReferenceKey((operationDefinition as LogicAppsV2.HybridTriggerOperationAction).inputs);
     default:
       throw Error('No known connection reference key type');
   }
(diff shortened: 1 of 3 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -440,3 +443,19 @@ export function getLegacyConnectionReferenceKey(operationDefinition: any): strin
   }
   return referenceKey;
 }
+
+function getHybridTriggerConnectionReferenceKey(operationDefinition: LogicAppsV2.HybridTriggerConnectionInfo): string {
+  let connectionName: string;
+  if (typeof operationDefinition.host.connection.name === 'string') {
```
preetriti1, line 449: do you guys have two formats for a hybrid trigger?

libs/utils/src/lib/models/logicAppsV2.ts
```diff
@@ -100,6 +100,14 @@ export interface OpenApiOperationInputs extends RetryableActionInputs {
   parameters?: any;
 }
 
+export interface HybridTriggerOperationAction extends Trigger {
+  inputs: HybridTriggerConnectionInfo;
```
preetriti1, line 104: Why do you call it action and trigger when this is just trigger.. and this does not seem correct you have other properties also inside inputs no? like the schema

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -120,7 +120,7 @@ export const getUpdatedManifestForSplitOn = (manifest: OperationManifest, splitO
       throw new AssertionException(AssertionErrorCode.INVALID_SPLITON, invalidSplitOn);
     }
 
-    const isAliasPathParsingEnabled = manifest.properties.connectionReference?.referenceKeyFormat === ConnectionReferenceKeyFormat.OpenApi;
+    const isAliasPathParsingEnabled = (manifest.properties.connectionReference?.referenceKeyFormat === ConnectionReferenceKeyFormat.OpenApi || manifest.properties.connectionReference?.referenceKeyFormat === ConnectionReferenceKeyFormat.HybridTrigger) ? true : false;
```
preetriti1, line 123: why the need for ternary operator when both are boolean evaluations??????

preetriti1 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
bnarayanared: Removed the string check; only one connection type

libs/utils/src/lib/models/logicAppsV2.ts
bnarayanared: Added the other missing properties

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: Removed

preetriti1 (approved):

PR #2791, opened 2023-06-20
fix(designer): Disable Change Connection when in readonly

Eric-B-Wu:

bug fix
able to change connection
change connection button is now disabled

readonlyDisable


rllyy97:
Do we want to instead hide the "Change Connection" button altogether on read only views?  This should only be visible in monitoring view which is a whole separate page from default, so it's not like the button will be enabled again until you navigate away

Eric-B-Wu:
> Do we want to instead hide the "Change Connection" button altogether on read only views? This should only be visible in monitoring view which is a whole separate page from default, so it's not like the button will be enabled again until you navigate away

I think the connection doesn't even show in monitoring view, this was mostly just something for PA because according to them: "We have a specific scenario where we prepare to load a new/updated copy of a workflow, at this time the designer is readonly." so didn't know if we wanted to hide it for them

rllyy97:
> I think the change connection doesn't even show in monitoring view, this was mostly just something for PA because according to them: "We have a specific scenario where we prepare to load a new/updated copy of a workflow, at this time the designer is readonly." so didn't know if we wanted to hide it for them

Oh true I forgot we don't show the same parameters tab

rllyy97 (approved):

Issue #2788, opened 2023-06-19
Intermittently unable to expand expression editor


BrianCollet:
### Describe the Bug with repro steps

1. Go to workflow designer
2. Select set variable action (not sure if occurs with other actions, but happened with this action in my experience)
3. Click in the value input box
4. Select insert expression
5. The expression editor opens (although not fully) and trying to expand it does not work and also unable to type anything into it
6. Console log errors pasted below

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/50160870/96d2b0e6-0cf2-4f96-b26d-c0009cdf268e)

### Browser

Edge, Chrome

### Additional context

Edge console logs:
6a57e33c-111c-4f67-b621-ecfb3314c387:1 Extension: HubsExtension
XF04yqkBGY3O.js:341 Version: 11.210.6.1 (v11.210.0.1#88c2b04310.230614-0057) Signed
Session: 0b699dcc7b4f4c4ab53f35430c5b155b
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
 Extension: Microsoft_Azure_PIMCommon
6a57e33c-111c-4f67-b621-ecfb3314c387:1 Extension: WebsitesExtension
2A7lcmKw7zxp.js:28 2023-06-19T12:50:28.363Z [info][WorkflowAppOverviewBlade][tabs][notification-count] - Logged '0' notifications
2A7lcmKw7zxp.js:28 2023-06-19T12:52:21.173Z [info][WorkflowAppOverviewBlade][tabs][notification-count] - Logged '0' notifications
[...]

PR #2792, opened 2023-06-20
fix(designer): Load monaco from npm instead of cdn

Eric-B-Wu:

bug fix
monaco is loaded from cdn. So if you don't have internet access or cdn is blocked, monaco errors and does not load
https://github.com/Azure/LogicAppsUX/issues/2788
monaco is initialized from npm package
no
before: 
before

after: 
after


Eric-B-Wu:
this will also make it so that monaco editor doesn't have a delay when opening the first time because it doesn't need to reach out to CDN

rllyy97 (approved):

Eric-B-Wu:
importing the whole monaco package is breaking our build and thus without a fix for this we cannot complete this PR

Issue #2795, opened 2023-06-20
Existing workflow can be saved after removing trigger, reloading page causes workflow to never show


wifun:
### Describe the Bug with repro steps

1. go to an existing workflow
2. delete the trigger
3. save
4. reload the page

![image](https://github.com/Azure/LogicAppsUX/assets/55114634/6258dd7d-d788-4f85-9d65-e13fbcee9f30)

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using Preview Designer or GA Designer

Preview

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Chrome

AB#24325128

rllyy97:
This is just appearing in staging / our next release, will work on getting this hotfixed today

PR #2804, opened 2023-06-21
fix(Designer): Workflows without triggers now properly deserialize

rllyy97:
## Main Changes
- With our recent fix to serializing workflows without triggers, we were not handling the matching deserialization instance
  - Workflows would appear to just be blank with no option to add anything
- We now add the placeholder trigger node on deserialization if there is no trigger
  - Also added some catches so we don't over-initialize the placeholder node

Fixes https://github.com/Azure/LogicAppsUX/issues/2795

---
### One action with no triggers deserializes as this now:
image

### DIFF
--- a/libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
+++ b/libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
+import constants from '../../../common/constants';
 import { UnsupportedException, UnsupportedExceptionCode } from '../../../common/exceptions/unsupported';
 import type { Operations, NodesMetadata } from '../../state/workflow/workflowInterfaces';
 import { createWorkflowNode, createWorkflowEdge } from '../../utils/graph';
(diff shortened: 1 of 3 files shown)


hartra344 (approved):

libs/designer/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
```diff
@@ -51,6 +52,16 @@ export const Deserialize = (
       ...addTriggerInstanceMetaData(runInstance),
     };
     allActionNames.push(tID);
+  } else {
+    // Workflow has no trigger, create a placeholder trigger node to reference during initialization
+    const tID = constants.NODE.TYPE.PLACEHOLDER_TRIGGER;
+    triggerNode = createWorkflowNode(tID, WORKFLOW_NODE_TYPES.PLACEHOLDER_NODE);
```
preetriti1, line 58: I thought we already had that or maybe it was only for totally empty designer which has no actions as well.

rllyy97: Completely empty designer is handled fine, this is just when there's other actions that rely on there being a trigger to reference in the graph

Issue #2686, opened 2023-06-06
Duplicate calls being made for swagger and manifest when another call is either in progress or data present in query cache


preetriti1:
Currently we query the cache multiple times to fetch manifest or swagger for a connectorId and operationId.
It seems either the query cache is evicting the item or getting reset which is causing multiple calls for the above which might cause perf bottlenck

AB#24195114

PR #2806, opened 2023-06-21
fix(Designer): Removed duplicate service calls

rllyy97:
## Main Changes
Currently we make several of our network requests multiple times due to a misconfiguration of some of our query store code.
(Each place we defined a manifest hook it would make a new request)
We also had a second query client wrapping our intl provider (does not seem to be needed) that was splitting our cache in two and allowing duplicate requests
These query store configs have been adjusted to properly cache and retrieve data, reducing duplicate requests by a large amount.

---
### Old: 4 manifest requests per operations (then infinitely more as the invalid queries refetch)
image

### New: One manifest request
image

Fixes https://github.com/Azure/LogicAppsUX/issues/2686


hartra344 (approved):

apps/vs-code-react/src/app/export/export.tsx
```diff
@@ -6,23 +6,14 @@ import './export.less';
...
 import { Outlet, useOutletContext } from 'react-router-dom';
 
-const queryClient = new QueryClient({
-  defaultOptions: {
-    queries: {
-      refetchInterval: false,
-      refetchIntervalInBackground: false,
-      refetchOnWindowFocus: false,
-      refetchOnReconnect: false,
-      refetchOnMount: true,
```
ccastrotrejo, line 22: In this one, the only difference from the `getReactQueryClient` is that the `refetchOnMount` its true here and over there its false. I do remember we faced some issues when we set it up to false

ccastrotrejo (changes requested):

apps/vs-code-react/src/app/export/export.tsx
rllyy97: I guess it's not actually needed over here to use the designer query client, I'll go ahead and revert this

ccastrotrejo: Sure, I guess in the future if we needed I can take a further look on how to properly make it work it without the refetchOnMount

PR #2813, opened 2023-06-21
fix(vscode): Updated Designer Host.Json Config files 

samikay101:
Removed trailing comma from host.json designer file, small nit.


samikay101:
Trailing comma was part of javascript file, this PR is no longer needed.

PR #2816, opened 2023-06-22
fix(designer): Disable selection of invalid connections for selectConnectionsTab

wifun:

Disable selection of invalid connections for selectConnectionsTab

image

image

No

AB#24380640


libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -73,20 +75,40 @@ const updateNodeConnectionAndProperties = async (
...
-    getAllVariables(newState.tokens.variables),
-    dispatch,
-    getState,
-    newState.workflow.newlyAddedOperations[nodeId] ? undefined : newState.workflow.operations[nodeId]
-  );
+  const operationInfo = newState.operations.operationInfo[nodeId];
+  const reference = getConnectionReference(newState.connections, nodeId);
+  const isValidConnection = await isConnectionReferenceValid(operationInfo, reference);
+
+  if (isValidConnection) {
```
preetriti1, line 82: What is the need of checking connection validity here? Connections should always be valid when this is called because connections UI should block selecting invalid connections.
Also connection object should know the validity of the connection so please do not repeat the calls when it is not needed

preetriti1 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
wifun: Reverted this proposal, now the PR is to block selecting invalid connections

Eric-B-Wu (approved):

shuktika12163:
Curious why we are disabling selection of invalid connection. I thought selecting the invalid connection is the only way to fix connections currently. @preetriti1 Does LAUX currently fix their connection by selecting the invalid connection? Will it be a breaking change for you? If I recall correctly, clicking on connection makes it go through setupConnectionIfNeeded function, which can be used to fix the connection if needed. By making invalid connection disabled, does user have any way to fix a specific invalid connection?

libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
shuktika12163: Curious why we are disabling selection of invalid connection. I thought selecting the invalid connection is the only way to fix connections currently. @preetriti1 Does LAUX currently fix their connection by selecting the invalid connection? Will it be a breaking change for you? If I recall correctly, clicking on connection makes it go through setupConnectionIfNeeded function, which can be used to fix the connection if needed. By making invalid connection disabled, does user have any way to fix a specific invalid connection?

wifun: From offline discussion, LAUX designer does not have support to fix connections currently. Desired behavior is to prevent selection of invalid connections

preetriti1 (approved):

PR #2856, opened 2023-06-27
fix(PowerAutomate): Re-initialize connectorId on fetching operation manifest if it has changed

shuktika12163:

We have some connectors that build on top of other connectors, eg. AI Builder connector is actually a CDS connector and overrides icon/display information to appear different. To make AIB actions show in a different group from CDS ones in search service, the connectorId is set to "aibuilder" when search service fetches AIB actions. On adding these actions to the flow, the re-fetched operation manifest gets CDS connectorId to fetch dynamic data and to use connections appropriately. However, operationInfo is only initialized once based on the connectorId fetched from search service, so our dynamic calls fail for the AIB actions (since it retrieves aibuilder instead of CDS connectorId). This change re-initializes operationInfo on fetching manifest if connectorId has changed.

Fetched connectorId for some connectors is different during search and during operation use (add/edit/save). ConnectorId is only initialized once through initializeOperationInfo based on value fetched from search service operation. Due to this, dynamic calls fail since the connectorId needs to be different (but isn't) on fetching operation manifest and initializing operation.

OperationInfo is re-initialized after fetching operation manifest if connectorId has changed. This stores the correct connectorId from operationmanifest in the redux store instead of keeping the one from search service operation. Dynamic calls and operation initialization succeeds.


libs/designer/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -113,6 +113,10 @@ const initializeOperationDetails = async (
     isConnectionRequired = isConnectionRequiredForOperation(manifest);
     connector = manifest.properties?.connector;
 
+    if (connector && connector.id !== operationInfo.connectorId) {
+      dispatch(initializeOperationInfo({ id: nodeId, ...operationInfo, connectorId: connector.id }));
```
shuktika12163, line 117: For certain connectors, connectorId changes based on whether it's being retrieved for search service or if it's retrieved as part of operation manifest. We want the connectorId to be re-initialized to the correct one once it's fetched from operation manifest and is different from the search service one.

preetriti1: This is a hack, we cannot have code like this spilled in places just to support some scenario. This looks like a bug in manifest where it does not have correct connectorId which is used to actually fetch the manifest...
You see a manifest is fetched from {connectorId, operationId} tuple.

Please revisit this in the service either backend or search service, it should be a first class connector either from apihub / rp service / hard coded in your client [like we have so many built ins]
If there are PA scenarios which have such connectors, some system needs to be built to support it and not like this in designer but in client service or in backend.

preetriti1 (changes requested):

libs/designer/src/lib/core/actions/bjsworkflow/add.ts
shuktika12163: Thanks Priti. We have this connector as built in hardcoded in RP but it builds on top of another one and sends different connector id to display data correctly based on when it's being fetched. So, the connector id changes. I agree, this usage of changing connectorId does not seem right so we should look for a way to solve it on our end. I just wanted to run this solution by your team in case it made sense.

I'm thinking another option is to initializeOperationInfo in our host service when we first load the manifest with expanded connector info. Do you think it's okay to change this to just export that function from LAUX instead, or are there be concerns about it too?

preetriti1: I think you should have a different operation like apiOperation, manifest and different connectorID, different connection parametera.. and that way in the host when the call comes for getting dynamic data comes you will know which actual connector to call and similarly for connection. 
Many builtins may need specific client codes to handle their cases.

I do have concerns in exposing store update methods to host because it exposes so many entry points in manipulating the data out of what's needed and this can be misused because it is doing things which out of the designed updates... this method is not something which is actually updating something based on user interaction so please don't do it this way and try to create implementation in the calls happening rather trying to fix it in one line.

shuktika12163: Thank you Priti! Discussed with the team today and we'll try to push the connector team to make it a generic connector and we'll hack the call on our end in the meantime.

shuktika12163:
As per latest discussion, the changes will be made in host service and in the connector.

Issue #2898, opened 2023-07-03
"ContainerGroup Containers Item" part of "Create or update a container group" action is misleading/broken 


AndreiBarbu95:
### Describe the Bug with repro steps

I mentioned below how I reproduced it. Maybe there is an easier way.
1 - Create an Azure Container Registry and have an image there
2 - Create an Azure Container Instance with the image from ACR
3 - Create a Standard Logic App with a Workflow and at the Designer level use "When a HTTP request is received" with the following body:

```
{
    "properties": {
        "acrName": {
            "type": "string"
        },
        "eventName": {
            "type": "string"
        },
        "id": {
            "type": "string"
        },
        "image": {
            "type": "string"
        },
        "request": {
            "properties": {
                "headers": {
                    "type": "object"
                },
                "method": {
                    "type": "string"
                },
                "uri": {
                    "type": "string"
                }
            },
            "type": "object"
        },
        "status": {
            "type": "string"
        },
        "timestamp": {
            "type": "string"
        }
    },
    "type": "object"
}
```

Then create action "Create or update a container group" and fill the details for the ACI. Filling "ContainerGroup Containers Item" is very difficult and I think properties is not placed correctly.

4 - At ACR level create a webhook referencing HTTP POST URL from "When a HTTP request is received", action push, scope the image used at ACI level.

5 - Perform a Ping from the ACR webhook.

6 - Check the run at Worflow level under Run history and if the provided JSON under "ContainerGroup Containers Item" is not fine you will get an error.

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using Preview Designer or GA Designer

I am not sure. There is a new Designer but I would assume it's GA.

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

No

### Workflow JSON

```json
#Below is the default JSON provided for "ContainerGroup Containers Item"
[
    {
        "name": "",
        "": {
            "image": "",
            "command": "",
            "instanceView": {
                "restartCount": "",
                "currentState": {
                    "state": "",
                    "startTime": "",
                    "exitCode": "",
                    "finishTime": "",
                    "detailStatus": ""
                },
                "previousState": {
                    "state": "",
                    "startTime": "",
                    "exitCode": "",
                    "finishTime": "",
                    "detailStatus": ""
                }
            },
            "resources": {
                "requests": {
                    "memoryInGB": "",
                    "cpu": "",
                    "gpu": {
                        "count": "",
                        "sku": ""
[...]

PR #2979, opened 2023-07-17
fix(designer): Complex Array Editor - Support for nested items

Eric-B-Wu:

Bug fix
We weren't serializing/deserializing correctly, so we would only be able to edit in collapsed array editor
https://github.com/Azure/LogicAppsUX/issues/2898
we are able to see expanded editor now

[test1.zip](https://github.com/Azure/LogicAppsUX/files/12072860/test1.zip) - video too large


libs/designer-ui/src/lib/editor/base/utils/editorToSegement.ts
```diff
@@ -31,6 +31,8 @@ const getChildrenNodesToSegments = (node: ElementNode, segments: ValueSegment[],
 };
 
 export const convertStringToSegments = (value: string, tokensEnabled?: boolean, nodeMap?: Map<string, ValueSegment>): ValueSegment[] => {
+  if (!value) return [];
+  if (typeof value !== 'string') return [{ id: guid(), type: ValueSegmentType.LITERAL, value: JSON.stringify(value) }];
```
rllyy97, line 35: If the `value` param is typed as string do we need to do the JSON.stringify?  If we're not sure what value is we might want to do a try catch or something in case it errors

rllyy97 (approved):

libs/designer-ui/src/lib/editor/base/utils/editorToSegement.ts
hartra344: I think it puts proper quotes around it if it wasn't already there:
CleanShot 2023-07-17 at 12 16 08@2x

hartra344: Not sure if it's needed but it does make a real difference in output

Eric-B-Wu: I had tested with some sample values and it seems to be working without it, so I went ahead and removed it

PR #2988, opened 2023-07-18
feat(Designer): Added node selection functionality options

rllyy97:
## Main Changes
- Added option to suppress default functionality for selecting nodes in designer
  - In EiPaaS we are selecting nodes but opening a completely different panel, and not show the parameters panel like normal
- Added option to give an extra callback to run on selecting nodes in designer
  - This might be helpful for EiPaas work to open their new panel


libs/designer/src/lib/ui/CustomNodes/OperationCardNode.tsx
```diff
@@ -74,6 +79,9 @@ const DefaultNode = ({ targetPosition = Position.Top, sourcePosition = Position.
   const isSecureInputsOutputs = useSecureInputsOutputs(id);
   const { status: statusRun, error: errorRun, code: codeRun, repetitionCount } = runData ?? {};
 
+  const suppressDefaultNodeSelect = useSuppressDefaultNodeSelectFunctionality();
+  const nodeSelectCallbackOverride = useAddedNodeSelectCallback();
```
ccastrotrejo, line 83: NIT: I don't know if `useAddedNodeSelectCallback` is the right name. I am just picky about the `Added` part

libs/designer/src/lib/core/DesignerProvider.tsx
```diff
@@ -22,15 +22,7 @@ const OptionsStateSet = ({ options, children }: any) => {
...
-    dispatch(
-      initDesignerOptions({
-        readOnly: options.readOnly,
-        isMonitoringView: options.isMonitoringView,
-        isDarkMode: options.isDarkMode,
-        useLegacyWorkflowParameters: options.useLegacyWorkflowParameters,
-        isXrmConnectionReferenceMode: options.isXrmConnectionReferenceMode,
-      })
-    );
+    dispatch(initDesignerOptions(options));
```
ccastrotrejo, line 25: Do we need to update the designerOptionsSlice for the suppressDefaultNodeSelectFunctionality and addedNodeSelectCallback?

ccastrotrejo (changes requested):

libs/designer/src/lib/core/DesignerProvider.tsx
rllyy97: Switching it over to this implementation makes it so we don't have to specify each option anymore

libs/designer/src/lib/ui/CustomNodes/OperationCardNode.tsx
rllyy97: Yeah I wasn't really sure how to phrase both of these, do you have any suggestions?

libs/designer/src/lib/core/DesignerProvider.tsx
ccastrotrejo: Yep, thats great, but I wonder if on the execution of `initDesignerOptions` we need to specify this new properties.
image

libs/designer/src/lib/ui/CustomNodes/OperationCardNode.tsx
rllyy97: Discussed offline, changing to `nodeSelectAdditionalCallback`

ccastrotrejo (approved):

PR #2996, opened 2023-07-19
feat(designer): Add EditorService to allow custom components

cdgolmar-msft:
# Changes

This PR implement changes discussed in #2871.

Feature: introduce a new service `IEditorService` to allow package consumers to replace editors with a custom component.
This customization mechanism will be consumed in Power Automate to migrate some custom experience with operation parameters.

When editor service is implemented, it can decide if a parameter should use a custom editor by providing the React component to use instead of the default one. The custom component can still render the default editor if needed.

There is no behavior change if editor service is not implemented, or if it returns no customization.

Added tests for introduces functions, and `SettingTokenField` component.

Added an example of custom editor in local `designer-standalone` app: a checkbox allows enabling custom editors. Then the "Initialize variable" and "Increment variable" "value" parameter input is replaced by another component (see screenshots).

Next steps:
- Similar change for editors in universal connection parameter
- Integration in Power Automate (and maybe fixes or adjustments based on needs/feedback)

# Checklist

# Screenshots

_Example of custom editor for "Increment variable" "value" parameter_

Screenshot 2023-07-19 171514

Screenshot 2023-07-20 104456

laux-custom-editor

AB#24591195


hartra344:
@eric-b-wu can you review this one?

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -315,3 +328,38 @@ const TokenField = ({
       );
   }
 };
+
+export type CustomTokenFieldProps = TokenFieldProps & { editorOptions: ICustomEditorOptions };
+
+export const CustomTokenField = (props: CustomTokenFieldProps) => {
+  invariant(isCustomEditor(props), 'only custom editor can be rendered by CustomTokenField');
```
Eric-B-Wu, line 335: instead of throwing an error here, let's just use our default editors

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -315,3 +328,38 @@ const TokenField = ({
...
+function useRenderDefaultEditor(tokenFieldProps: Omit<TokenFieldProps, 'editor' | 'editor' | 'value' | 'onValueChange'>) {
+  return useCallback(
+    (props: IRenderDefaultEditorParams) => {
+      return (
+        <TokenField
+          {...tokenFieldProps}
+          editor={props.editor}
+          editorOptions={props.editorOptions}
+          value={props.value as ValueSegment[]}
+          onValueChange={props.onValueChange as ChangeHandler}
```
Eric-B-Wu, line 359: instead of converting onValueChange to ChangeHandler, we should change it of type ChangeHandler without having to update the viewmodel

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -315,3 +328,38 @@ const TokenField = ({
...
   }
 };
+
+export type CustomTokenFieldProps = TokenFieldProps & { editorOptions: ICustomEditorOptions };
+
+export const CustomTokenField = (props: CustomTokenFieldProps) => {
+  invariant(isCustomEditor(props), 'only custom editor can be rendered by CustomTokenField');
+  const { EditorComponent, editor, editorOptions = {} }: ICustomEditorOptions = props.editorOptions;
+
+  const renderDefaultEditor = useRenderDefaultEditor(props);
```
Eric-B-Wu, line 338: instead of using the default editor, and omitting props, we can just import the UI editor component in our custom render. This logic shouldn't be within the SettingTokenField file

libs/designer/src/lib/core/state/designerOptions/designerOptionsSlice.ts
```diff
@@ -73,6 +75,11 @@ export const initializeServices = createAsyncThunk(
       InitRunService(runService);
     }
 
+    console.log(' InitEditorService', !!editorService);
```
Eric-B-Wu, line 78: remove

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
```diff
@@ -413,11 +415,26 @@ const ParameterSection = ({
...
 };
 
-const getEditorAndOptions = (
+export const getEditorAndOptions = (
+  operationInfo: OperationInfo,
   parameter: ParameterInfo,
   upstreamNodeIds: string[],
   variables: Record<string, VariableDeclaration[]>
 ): { editor?: string; editorOptions?: any } => {
+  const customEditor = EditorService()?.getEditor({
```
Eric-B-Wu, line 424: let's move this filtering logic outside into our into getEditor, and let's make it so that getEditor returns { editor?: string; editorOptions?: any }, so we won't need toEditorAndOptions

libs/services/designer-client-services/src/lib/editor.ts
```diff
@@ -0,0 +1,109 @@
+// import type { CastHandler, ChangeHandler, ValueSegment } from '@microsoft/designer-ui'
```
Eric-B-Wu, line 1: remove

libs/services/designer-client-services/src/lib/editor.ts
```diff
@@ -0,0 +1,109 @@
...
+import type * as React from 'react';
```
Eric-B-Wu, line 2: also we don't need to import react

libs/services/designer-client-services/src/lib/editor.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export const EditorService = (): IEditorService | undefined => {
+  return service;
+};
+
+// Overridden custom editors are identified by this name internally.
+const customEditorName = 'internal-custom-editor';
+
+export const isCustomEditor = (props: {
```
Eric-B-Wu, line 93: a cleaner way we can write
```
export const isCustomEditor = ({ editor, editorOptions }: { editor?: string; editorOptions?: ICustomEditorOptions }) => {
  return equals(editor, customEditorName) && typeof editorOptions === 'object' && typeof editorOptions?.EditorComponent === 'function';
};
```

Eric-B-Wu (changes requested):

libs/services/designer-client-services/src/lib/editor.ts
```diff
@@ -0,0 +1,109 @@
...
+  operationId: string;
+  parameterKey: string;
+  parameterName: string;
+  required: boolean;
+  schema?: any;
+  editor?: string;
+  editorOptions?: Record<string, any>;
+}
+
+export type IEditorValueSegmentChangeEvent = { value: ValueSegment[] };
```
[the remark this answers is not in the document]
Eric-B-Wu: I think we should be able to export it from the @microsoft/logic-apps-designer; however, the main thing I was getting at is that onChange should be of type ChangeHandler 
```
ChangeHandler = ({  value: ValueSegment[], viewModel?: any}) => void;
```
such that users would be also able to change the viewModel

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
Eric-B-Wu: Instead of destructuring parameter and operationInfo, we can just pass them the whole object. What I'm getting at is that we shouldn't be creating new interfaces that abstract the Parameter Information, and if there is more information that you want to be returned from the customer editor you can extend it there (for now I know there will be circular dependencies, but for now can we just create a copy of the interfaces -- a future PR I'll extract these models to the utils that can be imported). Also I think most of this logic should be done in libs\designer\src\lib\core\utils\parameters\helper.ts at: 
```
getParameterEditorProps
```
We shouldn't be calling our services in our UI component

Eric-B-Wu (approved):

Issue #2701, opened 2023-06-08
Standard_Custom Http Timeout Setting Per Request Action: Unable to debug the workflow, getting the below error with PT1S and PT24H request timeout.


vgouth:
### Describe the Bug with repro steps

1.Create a standard logic app
2.Create an empty stateless workflow
3.Add a request trigger and HTTP request action.
4.The request url to point to
([https://httptimeoutendpoint.azurewebsites.net:443/api/longresponse/triggers/manual/invoke?api-version=2020-05-01-preview&sp=%2ftriggers%2fmanual%2frun&sv=1.0&sig=BiCOpndJgI4xnNMW0n1MkN0NgRPgcH0NIK-h8NStZrI](https://httptimeoutendpoint.azurewebsites.net/api/longresponse/triggers/manual/invoke?api-version=2020-05-01-preview&sp=%2ftriggers%2fmanual%2frun&sv=1.0&sig=BiCOpndJgI4xnNMW0n1MkN0NgRPgcH0NIK-h8NStZrI))
5.Go to settings tab and set Request options to PT1S or PT24H
6.Disable the Asynchronous pattern.
7.Save and run the workflow.
8.**Expected Result:** Debugging should run without any errors.
**Actual Result:** Unable to debug the workflow, getting the below error with PT1S and PT24H request timeout.

**Note:** While trying to open the overview page getting the 404 error.

### What type of Logic App Is this happening in?

Standard (VSCode)

### Are you using Preview Designer or GA Designer

Preview

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

No

### Workflow JSON

```json
PT1S workflow JSON

{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "HTTP": {
                "type": "Http",
                "inputs": {
                    "uri": "https://httptimeoutendpoint.azurewebsites.net:443/api/longresponse/triggers/manual/invoke?api-version=2020-05-01-preview&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=BiCOpndJgI4xnNMW0n1MkN0NgRPgcH0NIK-h8NStZrI",
                    "method": "POST"
                },
                "runAfter": {},
                "operationOptions": "DisableAsyncPattern",
                "runtimeConfiguration": {
                    "contentTransfer": {
                        "transferMode": "Chunked"
                    },
[...]

Issue #2922, opened 2023-07-06
Use HTTP Action in stateless workflow, Logic Apps Standard designer causes error.


nobuko-yamada:
### Describe the Bug with repro steps

Put any trigger and put HTTP Action in stateless workflow

Cannot save with the error below.
![image](https://github.com/Azure/LogicAppsUX/assets/115044612/5aab49d1-9ef2-4727-91c3-0dd66e801ef4)

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "HTTP": {
                "inputs": {
                    "method": "GET",
                    "uri": "https://jpazinteg.github.io/"
                },
                "runAfter": {},
                "runtimeConfiguration": {
                    "contentTransfer": {
                        "transferMode": "Chunked"
                    }
                },
                "type": "Http"
            },
            "HTTP_1": {
                "inputs": {
                    "method": "GET",
                    "uri": "https://jpazinteg.github.io/"
                },
                "runAfter": {
                    "HTTP": [
                        "SUCCEEDED"
                    ]
                },
                "runtimeConfiguration": {
                    "contentTransfer": {
                        "transferMode": "Chunked"
                    }
                },
                "type": "Http"
            }
        },
        "contentVersion": "1.0.0.0",
        "outputs": {},
        "triggers": {
            "When_a_HTTP_request_is_received": {
                "kind": "Http",
                "type": "Request"
            }
        }
    },
    "kind": "Stateless"
}
```

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/115044612/43cfe010-fe2e-4dee-aae6-56a6f31d16fd)

### Browser

Edge and Chrome

AB#24464482

PR #2997, opened 2023-07-19
fix(designer): Default to off on content transfer chunking in stateless workflows

Eric-B-Wu:

Bugfix: https://github.com/Azure/LogicAppsUX/issues/2922
Also adds a stateful/stateless into our workflow store
When adding an action that has chunking, ex HTTP, it automatically set the setting as on, which will error when saving
When adding an action that has chunking, it will be off by default
Will require a change in Portal

AB#24591376


Eric-B-Wu:
also i wasn't sure if there was a use case where we would need this setting in a stateless workflow; so for now I've just kept it like old designer and didn't turn off the isSupported flag

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/2701 as well

hartra344 (approved):

PR #3030, opened 2023-07-25
fix(data mapper): functions missing on load schema

DanielleCogs:

bug fix
because of recent improvement, functions are not being shown on correct target nodes
fixes the above
no
N/A- app should be functioning as normal


__mocks__/mapDefinitions/FullTranscriptMapDefinition.ts
```diff
@@ -1,144 +1,16 @@
-﻿export const fullTranscriptMapDefinitionString = `
```
refortie, line 1: This change might be an accident?

libs/data-mapper/src/lib/models/Schema.ts
```diff
@@ -64,7 +64,7 @@ export interface SchemaNodeExtended extends SchemaNode {
   pathToRoot: PathItem[];
   parentKey: string | undefined;
   // Used only in ArrayItem arrays, not regular arrays
-  arrayItemIndex: number | undefined;
+  arrayItemIndex?: number | undefined;
```
refortie, line 67: Why make this optional?

libs/data-mapper/src/lib/utils/Function.Utils.ts
```diff
@@ -134,3 +135,25 @@ export const functionsForLocation = (functions: FunctionDictionary, targetKey: s
   Object.fromEntries(
     Object.entries(functions).filter(([_key, value]) => value.functionLocations.some((location) => location.key === targetKey))
   );
+
+export const getFunctionLocationsForAllFunctions = (
+  dataMapConnections: ConnectionDictionary,
+  flattenedTargetSchema: SchemaNodeDictionary
+): FunctionDictionary => {
+  const functionNodes: FunctionDictionary = {};
```
refortie, line 143: Nit: add a little white space to this to break it into logical sections?

refortie (changes requested):
Remove the non-DM changes?

libs/data-mapper/src/lib/models/Schema.ts
DanielleCogs: it was giving me errors when I was adding the mocks that I pulled directly from the store. wouldn't it only apply to certain nodes anyways?

__mocks__/mapDefinitions/FullTranscriptMapDefinition.ts
DanielleCogs: nope I was using it for testing, fixed now

libs/data-mapper/src/lib/utils/Function.Utils.ts
DanielleCogs: done

refortie (approved):

PR #3065, opened 2023-08-01
feat(data mapper): XPath function now works (when backend makes metadata change)

DanielleCogs:

- feature allowing XPath function to work

none
- new textbox allowing for longer inputs in panel
- allow for functions that have only custom input, none from other nodes
- no

Screenshot 2023-07-31 at 5 31 30 PM


libs/data-mapper/src/lib/components/inputTypes/InputTextbox.tsx
```diff
@@ -0,0 +1,91 @@
...
+          setConnectionInput({
+            targetNode: functionNode,
+            targetNodeReactFlowKey: selectedItemKey,
+            inputIndex: 0,
+            input: formattedValue,
+          })
+        );
+      }
+    };
+    const timeOutId = setTimeout(() => onCustomTextBoxChange(inputText), 500);
```
refortie, line 53: Is this just to debounce? If so take a look at https://github.com/react-hookz/web if you haven't. Might be some options there. We use some of the features from this elsewhere

libs/data-mapper/src/lib/components/propertiesPane/tabs/FunctionNodePropertiesTab.tsx
```diff
@@ -202,16 +204,25 @@ export const FunctionNodePropertiesTab = ({ functionData }: FunctionNodeProperti
...
-                        currentNode={functionData}
-                        placeholder={input.placeHolder}
-                        inputName={getInputName(inputConnection)}
-                        inputValue={getInputValue(inputConnection)}
-                        inputIndex={index}
-                        inputAllowsCustomValues={input.allowCustomInput}
-                      />
-                    </Tooltip>
+                    {input.displayAttribute === InputFormat.TextBox && (
+                      <InputTextbox
```
refortie, line 208: Should we have an info bubble or something to put the tool tip into?

libs/data-mapper/src/lib/models/Function.ts
```diff
@@ -512,5 +513,45 @@ export const functionMock: FunctionData[] = [
...
+        allowedTypes: [NormalizedDataType.String],
+        displayAttribute: DisplayAttribute.TextBox,
+        isOptional: false,
+        allowCustomInput: true,
+        placeHolder: 'expression',
+      },
+    ],
+    displayName: 'Execute XPath',
+    category: FunctionCategory.Utility,
+    // "iconFileName": "dm_category_customcode.svg",
```
refortie, line 533: These can be removed

libs/data-mapper/src/lib/models/Schema.ts
```diff
@@ -53,6 +53,11 @@ export enum NormalizedDataType {
   String = 'String',
 }
 
+export enum InputFormat {
+  TextBox = 'TextBox',
+  FilePicker = 'FilePicker',
```
refortie, line 58: I assume this does nothing for now?

libs/data-mapper/src/lib/models/Schema.ts
```diff
@@ -53,6 +53,11 @@ export enum NormalizedDataType {
   String = 'String',
 }
 
+export enum InputFormat {
```
refortie, line 56: Should we also have `Dropdown` as an option so that we are explicit about which input? Or just always default to dropdown?

libs/data-mapper/src/lib/models/Function.ts
```diff
@@ -26,6 +26,7 @@ export interface FunctionInput {
   allowedTypes: NormalizedDataType[];
   isOptional: boolean;
   allowCustomInput: boolean;
+  displayAttribute?: DisplayAttribute;
```
refortie, line 29: Kind of an awkward name? Why not something simple like "inputType"?

refortie (approved):

libs/data-mapper/src/lib/models/Function.ts
```diff
@@ -512,5 +513,45 @@ export const functionMock: FunctionData[] = [
...
+  {
+    key: 'XPath',
+    maxNumberOfInputs: 1,
+    functionName: 'xpath',
+    outputValueType: NormalizedDataType.Any,
+    inputs: [
+      {
+        name: 'XPATH expression',
+        allowedTypes: [NormalizedDataType.String],
+        displayAttribute: DisplayAttribute.TextBox,
```
refortie, line 525: Might want to add in some tests that test things like. Multiple input and other like that to make sure we are serializing the data in and out correctly

libs/data-mapper/src/lib/models/Schema.ts
DanielleCogs: Default is dropdown. Discussed this with Sanjiv yesterday! And if we want we can add more in the future

libs/data-mapper/src/lib/models/Schema.ts
DanielleCogs: Yeah so this will work as soon as backend pushes these changes, but not yet.

libs/data-mapper/src/lib/components/propertiesPane/tabs/FunctionNodePropertiesTab.tsx
DanielleCogs: https://github.com/Azure/LogicAppsUX/issues/3073 made a separate workitem for this!

libs/data-mapper/src/lib/models/Function.ts
DanielleCogs: I want it to be distinct from "allowedTypes" but I'll find something else to call it

refortie: That works

libs/data-mapper/src/lib/components/inputTypes/InputTextbox.tsx
DanielleCogs: done

libs/data-mapper/src/lib/models/Function.ts
DanielleCogs: I'll do this on the next PR, I need to do some similar work with XML files anyways

libs/data-mapper/src/lib/models/Function.ts
DanielleCogs: done

PR #3093, opened 2023-08-04
feat(designer): Allow opertations to specify a "display" operation group

speray:

This PR allows operations to specify a "display" operation group. An operation can ask to be listed under a different operation group during discoverability. When selected it will keep working as previously. This is only for marketing/logical grouping.
This is change is generic but will be used initially for AI Builder operations (see below).

After being added, saved or restored, the operation is behaving as belonging to its original group. For the display in the designer though, today the current behavior of the server is to specify the original group id, but the designer brandcolor and icon uri, so I haven't added anything in this repo to support it, but it may be logical to instead also use a similar mechanism there. Let me know if I should do this.

All AI Builder operations are operations of the dataverse connector. Still in the current flow designer in power automate, they are displayed as they would be a separate connector. See:

image

This is achieved by listed operations as having the operations claiming to belong to the operation group "providers/Microsoft.ProcessSimple/operationGroups/aibuilder". When the operation is selected by the user, a new call is made to retrieve the manifest and this time the server answers that it belongs to the api group "/providers/Microsoft.PowerApps/apis/shared_commondataserviceforapps". This isn't working the newest version of the designer (this repo), as it keeps somewhere a relationship link, and we were asked to update the behavior.

Without this PR and with the operation being listed as members of dataverse, it would look like this:
image
image

After the PR, they would be grouped as expected:
image
image

No.

See above

AB#24757956


libs/utils/src/lib/models/operationDiscoveryResults.ts
```diff
@@ -30,6 +30,7 @@ export interface SomeKindOfAzureOperationDiscovery {
   capabilities?: string[];
   isNotification?: boolean;
   api: OperationApi;
+  displayApi?: OperationApi;
```
speray, line 33: This can also be just an id, since the rest isn't used so far, but could be in the future. Would it better to just have a "displayApiId" ?

speray:
CC @shuktika12163 @edgarguoms

speray:

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
```diff
@@ -36,7 +36,12 @@ export const SearchResultsGrid: React.FC<PropsWithChildren<SearchResultsGridProp
   const intl = useIntl();
 
   const apiIds = useMemo(
-    () => Array.from(new Set(operationSearchResults.filter((r) => r !== undefined).map((res) => res.properties?.api?.id))),
+    () =>
+      Array.from(
+        new Set(
+          operationSearchResults.filter((r) => r !== undefined).map((res) => res.properties?.displayApi?.id || res.properties?.api?.id)
```
hartra344, line 42:
```suggestion
          operationSearchResults.filter((r) => r !== undefined).map((res) => res.properties?.displayApi?.id ?? res.properties?.api?.id)
```

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
```diff
@@ -60,9 +65,12 @@ export const SearchResultsGrid: React.FC<PropsWithChildren<SearchResultsGridProp
...
     (apiId: string | undefined, _index: number | undefined) => {
       if (!apiId) return;
-      const operations = operationSearchResults.filter((res) => res?.properties.api.id === apiId);
+      const operations = operationSearchResults.filter((res) => {
+        const displayApi = res?.properties.displayApi || res?.properties.api;
+        return displayApi?.id === apiId;
+      });
       if (operations.length === 0) return null;
-      const api = operations[0].properties.api;
+      const api = operations[0].properties.displayApi || operations[0].properties.api;
```
hartra344, line 73:
```suggestion
      const api = operations[0].properties.displayApi ?? operations[0].properties.api;
```

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
```diff
@@ -129,8 +137,8 @@ export const OperationActionDataFromOperation = (operation: DiscoveryOperation<D
   id: operation.id,
   title: operation.properties.summary,
   description: operation.properties.description,
-  brandColor: operation.properties.api.brandColor,
-  iconUri: operation.properties.api.iconUri,
+  brandColor: (operation.properties as BuiltInOperation).brandColor || operation.properties.api.brandColor,
```
hartra344, line 140: why is this different?

hartra344 (changes requested):

libs/designer-ui/src/lib/panel/__test__/searchResult.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+    renderer.render(<SearchResultsGrid {...minimal} groupByConnector={true} operationSearchResults={[operation, aliasedOperation]} />);
+
+    const grid = renderer.getRenderOutput();
+    const list = findByName<IListProps<string>>('List', grid);
+    expectValidReactElement(list);
+    expect(list.props.items).toHaveLength(2);
+    expect(list.props.items).toContain('api_id');
+    expect(list.props.items).toContain('displayApi_id');
+
+    const cell = list.props.onRenderCell!('displayApi_id');
```
hartra344, line 41:
```suggestion
    const cell = list.props.onRenderCell?.('displayApi_id');
```

Please fix linter errors as well, for non null assertion I suggestion option unwrapping like the suggestion above

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
hartra344: why mark as resolved? I don't see it fixed?

libs/designer-ui/src/lib/panel/__test__/searchResult.spec.tsx
speray: Ok thx, this didn't show up when running locally npm run lint, I'll update.

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
speray: Good question. The goal of the feature was to offer operation "aliasing" meaning, although an operation comes from ConnectorA, we'd like for marketing reasons to show it as ConnectorB. This piece of code is used to determine which iconUri and brandColor should be used, so I could use the same logic as above and check displayApi, but I actually saw that an operation itself can declare its own brandcolor and iconUri, so it felt natural to respect it.
Since this is something which was present in previous data model, it was limiting the surface of impact of the displayApi, and I would use displayApi only for grouping, but if you feel I should rather check displayApi here let me know, I'll update.

hartra344: ok last note then, as with he others, please use null coalescing operator rather than the || operator

```suggestion
  brandColor: (operation.properties as BuiltInOperation).brandColor ?? operation.properties.api.brandColor,
```

libs/designer-ui/src/lib/panel/recommendationpanel/searchResult.tsx
```diff
@@ -129,8 +137,8 @@ export const OperationActionDataFromOperation = (operation: DiscoveryOperation<D
   id: operation.id,
   title: operation.properties.summary,
   description: operation.properties.description,
-  brandColor: operation.properties.api.brandColor,
-  iconUri: operation.properties.api.iconUri,
+  brandColor: (operation.properties as BuiltInOperation).brandColor ?? operation.properties.api.brandColor,
+  iconUri: (operation.properties as BuiltInOperation).iconUri ?? operation.properties.api.iconUri,
```
preetriti1, line 141: Couldn't you have the iconuri and brandcolor inside api overridden?

PR #3120, opened 2023-08-11
fix(Designer): Fixed creation of phantom "cases" node input object - cherrypick release 2.55

rllyy97:
## Main Changes
Previously we have always deserialized case nodes with an extra "cases" phantom operation using parameter data meant for the case operations.
Now we only deserialize to the case operation data like intended.

We will be hotfixing this into our current release as it is a regression.
This was found in our current release due to some extra pre-save validation we added.

Cherry pick of this PR https://github.com/Azure/LogicAppsUX/pull/3119

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
@@ -304,24 +304,6 @@ const processChildGraphAndItsInputs = (
             });
           }
         }
-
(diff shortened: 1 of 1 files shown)


randiratnayake:
@rllyy97 When was this change meant to land in the live version? This is currently blocking some development work.

Eric-B-Wu:
> @rllyy97 When was this change meant to land in the live version? This is currently blocking some development work.

There were some regions that were not able to complete on Friday. We have this fix deployed to all regions by EOD today 8/14

randiratnayake:
Thank you @Eric-B-Wu. Confirming this is now fixed in AU region.

Issue #3107, opened 2023-08-08
Logic App Standard | Request trigger provide wrong callback url when have relative path


tonytang-microsoft-com:
### Describe the Bug with repro steps

Create a workflow with request trigger. See call back URL in the designer. It has the wrong encoded callback URL.

expected:
https://draclogicapp.azurewebsites.net:443/api/Test/triggers/When_a_HTTP_request_is_received/invoke?api-version=2022-05-01&sp=%2Ftriggers%2FWhen_a_HTTP_request_is_recei<stripped>
actual:
[https://draclogicapp.azurewebsites.net/api/Test/triggers/When_a_HTTP_request_is_received/invoke/test?api-version=2022-05-01&sp=%252Ftriggers%252FWhen_a_HTTP_request_is_received<stripped>

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

ANY

AB#24789949

Drac-Zhang:
@tonytang-microsoft-com  
one thing need to correct, it is Logic App Standard

Drac-Zhang:
HI Team, do we have any update on this issue or a rough ETA? Thank you

PR #3136, opened 2023-08-15
fix(designer): decode callback URL 

Eric-B-Wu:

Bugfix 
https://github.com/Azure/LogicAppsUX/issues/3107
decodes the URL so its no longer an invalid URL with relative paths
no
callbackurl

### DIFF
--- a/apps/designer-standalone/src/app/AzureLogicAppsDesigner/Services/WorkflowAndArtifacts.tsx
+++ b/apps/designer-standalone/src/app/AzureLogicAppsDesigner/Services/WorkflowAndArtifacts.tsx
@@ -108,9 +108,8 @@ export const listCallbackUrl = async (
   }
 
   let callbackUri: URL;
-
   if (callbackInfo.relativePath) {
-    callbackUri = new URL(callbackInfo.basePath, validateResourceId(callbackInfo.relativePath));
+    callbackUri = new URL(`${callbackInfo.basePath}${validateResourceId(callbackInfo.relativePath)}`);
     Object.entries(callbackInfo.queries).forEach(([key, value]) =>
       callbackUri.searchParams.append(encodeURIComponent(key), encodeURIComponent((value as any) ?? ''))
     );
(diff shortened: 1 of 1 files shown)


rllyy97 (approved):

ccastrotrejo (approved):

apps/designer-standalone/src/app/AzureLogicAppsDesigner/Services/WorkflowAndArtifacts.tsx
```diff
@@ -120,7 +119,7 @@ export const listCallbackUrl = async (
 
   return {
     method: callbackInfo.method,
-    value: callbackUri.toString(),
+    value: decodeURI(callbackUri.toString()),
```
tonytang-microsoft-com, line 122: I'm curious why we need to decode it here? The fact that we want to decode it to get the correct URL means to me that it was double encoded before. Can we find out where it was double encoded?

apps/designer-standalone/src/app/AzureLogicAppsDesigner/Services/WorkflowAndArtifacts.tsx
```diff
@@ -108,9 +108,8 @@ export const listCallbackUrl = async (
   }
 
   let callbackUri: URL;
-
   if (callbackInfo.relativePath) {
-    callbackUri = new URL(callbackInfo.basePath, validateResourceId(callbackInfo.relativePath));
+    callbackUri = new URL(`${callbackInfo.basePath}${validateResourceId(callbackInfo.relativePath)}`);
     Object.entries(callbackInfo.queries).forEach(([key, value]) =>
       callbackUri.searchParams.append(encodeURIComponent(key), encodeURIComponent((value as any) ?? ''))
```
tonytang-microsoft-com, line 114: Looks like it's double encoded here. When adding the search parameters, we should not encode it. URL.toString() will emit the correct URL by encoding the parameter

PR #3148, opened 2023-08-16
fix(Designer): Fixing the foatingActionMenu root object for HybridTriggers

bnarayanared:
**Issue:** Manual trigger assumes the root object of type:
image

But for hybrid triggers, the default rootObject needs to be nested inside rows..items like **so**:
image

**Fix:** Adding a feature flag (`isManualTrigger`) that differentiates what root object needs to be added based on the flag.
Once the rootObject is fixed, the schema objects are merged with the other input parameters, if they exist, for the token picker to convert both input parameters and floatingActionMenu items into tokens.


libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -123,20 +123,45 @@ function getDescriptionForDynamicallyAddedParameterType(type: DynamicallyAddedPa
...
+ *               },
+ *             }
+ *           }
  * @param value - valueSegment provided to us by rest of designer parent components
+ * @param isManualTrigger - Flag that sets the rootObject to required schema type.
  * @param onChange - handler to update value when the user changes their input in one of the dynamic parameters
  * @returns - array of props to render DynamicallyAddedParameter editors with
  */
-export function deserialize(value: ValueSegment[]): DynamicallyAddedParameterProps[] {
+export function deserialize(value: ValueSegment[], isManualTrigger: boolean): DynamicallyAddedParameterProps[] {
```
hartra344, line 148:
```suggestion
export function deserialize(value: ValueSegment[], isManualTrigger: boolean = false): DynamicallyAddedParameterProps[] {
```

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -175,11 +201,25 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
       return { ...resultPropertiesObj, [schemaKey]: propertyValue };
     }, {});
 
-  const rootObject = {
-    type: 'object',
-    properties,
-    required: requiredArray,
-  };
+  let rootObject: any;
+  if (isManualTrigger) {
```
hartra344, line 205: shouldn't this be the other way around?

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -156,9 +181,10 @@ export function deserialize(value: ValueSegment[]): DynamicallyAddedParameterPro
 /**
  * See deserialize function above for sample.
  * @param props - array of props of all DynamicallyAddedParameter editors currently rendered
+ * @param isManualTrigger - Flag that sets the rootObject to required schema type.
  * @returns - ValueSegment array with one literal -- value for which is a JSON representation of the dynamically added parameters in the shape expected by FlowRP
  */
-export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment[] {
+export function serialize(props: DynamicallyAddedParameterProps[], isManualTrigger: boolean): ValueSegment[] {
```
hartra344, line 187:
```suggestion
export function serialize(props: DynamicallyAddedParameterProps[], isManualTrigger: boolean = false): ValueSegment[] {
```

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -190,12 +230,26 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
   ];
 }
 
-export function getEmptySchemaValueSegmentForInitialization(useStaticInputs: boolean) {
-  let rootObject: { type: string; properties: unknown; required: string[] } = {
-    type: 'object',
-    properties: {},
-    required: [],
-  };
+export function getEmptySchemaValueSegmentForInitialization(useStaticInputs: boolean, isManualTrigger: boolean) {
```
hartra344, line 233:
```suggestion
export function getEmptySchemaValueSegmentForInitialization(useStaticInputs: boolean, isManualTrigger: boolean = false) {
```

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -190,12 +230,26 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
...
+  let rootObject;
+  if (isManualTrigger) {
```
hartra344, line 235: other way around?

libs/designer-ui/src/lib/floatingactionmenu/index.tsx
```diff
@@ -24,8 +24,9 @@ export interface FloatingActionMenuItem {
 
 export interface FloatingActionMenuProps {
   supportedTypes: Array<string>;
-  useStaticInputs: boolean |  undefined;
+  useStaticInputs: boolean | undefined;
   initialValue: ValueSegment[];
+  isManualTrigger: boolean;
```
hartra344, line 29:
```suggestion
  isManualTrigger?: boolean;
```

hartra344 (changes requested):

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
bnarayanared: defaulting now to false, so it is now correct.

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
bnarayanared: Same comment as above, defaulting now to false so now it's fixed

libs/utils/src/lib/helpers/functions.ts
```diff
@@ -315,11 +317,40 @@
...
+ * @arg {any} The source location to be merged from.
+ */
+function deepMerge(target: any, source: any) {
+  for (const key in source) {
+    if (Object.prototype.hasOwnProperty.call(source, key)) {
+      if (source[key] instanceof Object && !(source[key] instanceof Function)) {
+        if (!target[key]) target[key] = {};
+        deepMerge(target[key], source[key]);
+      } else {
+        target[key] = source[key];
```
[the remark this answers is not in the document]
hartra344: @bnarayanared please correct this as well

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
```diff
@@ -454,6 +463,27 @@ export const getEditorAndOptions = (
...
+      )
+    ) {
+      isManualTrigger = true;
+    }
+    return {
+      editor: editor,
+      editorOptions: {
+        ...editorOptions,
+        ...{
+          isManualTrigger,
```
wifun, line 482: Why not create a new option under `x-ms-editor-options` for the manual trigger manifest so we don't need this custom logic.

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -123,20 +123,45 @@ function getDescriptionForDynamicallyAddedParameterType(type: DynamicallyAddedPa
...
     return [];
   }
   // ASSUMPTION: for manual trigger, we assume there is *only one* ValueSegment which contains the required data
+  // ASSUMPTION: for hybrid triggers, the floating action segmets are always nested within rows.items
   const rootObject = JSON.parse(value[0].value);
 
   const retval: DynamicallyAddedParameterProps[] = [];
-  for (const [schemaKey, propertiesUnknown] of Object.entries(rootObject?.properties)) {
+  let itemsProperties: any;
+  if (isManualTrigger) {
```
wifun, line 158: What is the appropriate "default" behavior for this editor? This reads like you are special casing manual trigger and otherwise assume the general usage will be like for hybrid triggers.

With the context I have, I feel it should be the other way around. Like hybrid trigger gets special treatment if a flag is set. I don't think the default should be using rootObject.properties.rows.items : rootObject.rows.items to deserialize

bnarayanared: yes, currently working on adding a flag for hybridTriggers and leaving the default route unchanged

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -175,11 +201,25 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
       return { ...resultPropertiesObj, [schemaKey]: propertyValue };
     }, {});
 
-  const rootObject = {
-    type: 'object',
-    properties,
-    required: requiredArray,
-  };
+  let rootObject: any;
```
wifun, line 204: We should have a type for this or create one, and prefer no using `any` since this is new code when we should have the most context.
Maybe `SwaggerSchema`?

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -190,12 +230,26 @@ export function serialize(props: DynamicallyAddedParameterProps[]): ValueSegment
...
+export function getEmptySchemaValueSegmentForInitialization(useStaticInputs: boolean, isManualTrigger = false) {
+  let rootObject;
+  if (isManualTrigger) {
+    rootObject = {
+      type: 'object',
+      properties: {},
+      required: [],
+    };
+  } else {
+    rootObject = {
```
wifun, line 242: Curious, can you use the `default` property in manifest instead of using `getEmptySchemaValueSegmentForInitialization`?

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
```diff
@@ -123,20 +123,45 @@ function getDescriptionForDynamicallyAddedParameterType(type: DynamicallyAddedPa
...
+ *               },
+ *             }
+ *           }
  * @param value - valueSegment provided to us by rest of designer parent components
+ * @param isManualTrigger - Flag that sets the rootObject to required schema type.
  * @param onChange - handler to update value when the user changes their input in one of the dynamic parameters
  * @returns - array of props to render DynamicallyAddedParameter editors with
  */
-export function deserialize(value: ValueSegment[]): DynamicallyAddedParameterProps[] {
+export function deserialize(value: ValueSegment[], isManualTrigger = false): DynamicallyAddedParameterProps[] {
```
wifun, line 148: the parameter is always passed in, there's no need to be optional so I think we should make it required (same with the other helpers)

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -341,7 +341,18 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
       }
 
       const currentSchemaValue = getObjectPropertyValue(updatedManifest.properties.outputs, outputLocation);
-      safeSetObjectPropertyValue(updatedManifest.properties.outputs, outputLocation, { ...currentSchemaValue, ...schemaToReplace });
+
+      // if schema contains static object returned from RP, merge the current schema value and new schema value
+      if (schemaToReplace && 'rows' in schemaToReplace) {
```
wifun, line 346: I personally don't think using a hardcoded property name is reliable to have custom logic for. But don't have a suggestion cause I don't know much about the scenario. Perhaps LAUX team can comment

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
bnarayanared: Thank you for pointing out this option, I added the new `x-ms-editor-options' of `'isRequestApiConnectionTrigger` for the hybridTriggers instead and removed the customization here. For manual triggers it can remain undefined

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
bnarayanared: Changed to type `Schema`

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
bnarayanared: What would be the `default` in manifest? I don't see a default set

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
bnarayanared: Made it required, with a default false

libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
wifun: I might be wrong here, I though manifest's capability to set a default value would help here but it may not be possible

wifun: I might be wrong here, I though manifest's capability to set a default value would help here but it may not be possible

hartra344 (approved):

PR #3167, opened 2023-08-21
fix(Designer): Fixed type import build issue

rllyy97:
## Main Changes

One of our import statements had an internal `type` specification, this is fine in the Designer repo but will cause build errors on trying to build portal.
I've moved the type import out to it's own statement.

I looked for an eslint rule for this and I couldn't find one.  We have one import rule that potentially would have fixed this issue but it doesn't affect it when running `eslint --fix`.


rllyy97:
Bug source commit was reverted  =>  https://github.com/Azure/LogicAppsUX/pull/3160

PR #3169, opened 2023-08-21
feat: adding custom panel location capability

halenstalians23:

- Allows capability of specifying panel location for each individual panel type (errors, workflow parameters, etc.) This is needed for porting the errors pane into PPUX, as the spec requires it to render on a different side of the designer than in LAUX. This change lays the foundation, existing behavior is unchanged.

- All panels display on right side of designer.

- Same as current behavior.

N/A

N/A

AB#24899016


libs/designer/src/lib/ui/panel/panelRoot.tsx
```diff
@@ -61,6 +61,13 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
     [currentPanelMode]
   );
 
+  const customPanelLocations: CustomPanelLocation[] = [];
+  for (const customPanelLocation of customPanelLocations) {
+    if (currentPanelMode === customPanelLocation.panelMode) {
+      commonPanelProps.panelLocation = customPanelLocation.panelLocation;
```
hartra344, line 67: please move this into a local variable instead of changing the props themselves.

halenstalians23:

libs/designer/src/lib/ui/panel/panelRoot.tsx
```diff
@@ -61,6 +61,13 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
     [currentPanelMode]
   );
 
+  const customPanelLocations: CustomPanelLocation[] = [];
+  for (const customPanelLocation of customPanelLocations) {
```
hartra344, line 65: you're initializing customPanelLocations to an empty array above so this will never get ran?

halenstalians23: I will be passing this in as a Designer prop, this was only temporary, will update in next commit

libs/designer/src/lib/ui/panel/panelRoot.tsx
halenstalians23: ok, putting in memo function in next commit

libs/designer/src/lib/ui/panel/panelRoot.tsx
```diff
@@ -45,16 +46,21 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
 
   const dismissPanel = useCallback(() => dispatch(clearPanel()), [dispatch]);
 
-  const commonPanelProps: CommonPanelProps = useMemo(
-    () => ({
+  const commonPanelProps: CommonPanelProps = useMemo(() => {
+    let customLocation = undefined;
```
hartra344, line 50:
```suggestion
    let customLocation = customPanelLocations?.find(x => currentPanelMode === x.panelMode);
```

then you can remove the entire loop below

PR #3173, opened 2023-08-21
feat(designer): Add no actions node for is read only mode

ccastrotrejo:

### Main Code Changes
- Add no actions component for when is read only mode and there are no actions
- There is no tooltip in the component
- Added this component to avoid scenario in EiPaaS for when choosing property Id and workflow is empty
image

### Screenshots
Screenshot 2023-08-21 at 12 03 39 PM
Screenshot 2023-08-21 at 12 03 47 PM

### DIFF
--- a/libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
+++ b/libs/designer-ui/src/lib/dynamicallyaddedparameter/helper.ts
@@ -4,6 +4,7 @@ import constants from '../constants';
 import type { ValueSegment } from '../editor';
 import { ValueSegmentType } from '../editor';
 import { getIntl } from '@microsoft/intl-logic-apps';
+import type { Schema } from '@microsoft/parsers-logic-apps';
 import { generateUniqueName, guid } from '@microsoft/utils-logic-apps';
 
 export type DynamicallyAddedParameterIcon = string;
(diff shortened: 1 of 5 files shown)


libs/designer-ui/src/lib/card/noActionCard/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <div className="msla-selection-box" />
+        <div className="panel-card-main">
+          <div className="panel-card-header" role="button">
+            <div className="panel-card-content-container">
+              <div className="panel-card-top-content panel-card-top-content--no-action">
+                <div className="panel-msla-title">{triggerTitle}</div>
+              </div>
+            </div>
+          </div>
+        </div>
```
rllyy97, line 30: Do we need all of these?  I feel like we can just get away with the very last one

libs/designer-ui/src/lib/card/card.less
```diff
@@ -1072,6 +1080,14 @@
         color: @ms-color-black;
       }
     }
+
+    &--no-action {
+      border: 1px dashed @ms-color-primaryBorder;
+
+      &:hover {
```
rllyy97, line 1087: We don't need a hover effect since this is an un-interactive component.  If this is inheriting a bunch of styles lets just remove all the intherited classes from the component and have a fresh class, there is only going to be a few styles on this component anyways.

rllyy97 (changes requested):

hartra344 (approved):

rllyy97 (approved):

PR #3180, opened 2023-08-22
fix(Designer): Ensure outputs merge only based on isRequestApiConnection FF

bnarayanared:
Enclosing outputs merge only if `isRequestHybridTrigger` flag is set.
Here are the sample outputs that were used to test the scenarios:

https://docs.google.com/document/d/1-gPlN3IxPn5OomHG-pQKx4ASaBbSwVhUIQCSt2fFwhM/edit?usp=sharing

### DIFF
--- a/libs/designer/src/lib/core/utils/outputs.ts
+++ b/libs/designer/src/lib/core/utils/outputs.ts
@@ -360,11 +360,11 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
           schemaValue = { ...currentSchemaValue, ...schemaToReplace };
           shouldMerge = true;
         } else {
-          schemaValue = currentSchemaValue;
-          shouldMerge = false;
+          continue;
         }
       } else {
-        continue;
+        schemaValue = { ...currentSchemaValue, ...schemaToReplace };
+        shouldMerge = false;
       }
       safeSetObjectPropertyValue(updatedManifest.properties.outputs, outputLocation, schemaValue, shouldMerge);
     }
--- a/libs/designer/src/lib/core/utils/outputs.ts
+++ b/libs/designer/src/lib/core/utils/outputs.ts
@@ -349,9 +349,8 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
 
       const currentSchemaValue = getObjectPropertyValue(updatedManifest.properties.outputs, outputLocation);
 
-      const inputSchema = updatedManifest.properties?.inputs?.properties?.schema;
       const isRequestApiConnectionTrigger =
-        'x-ms-editor-options' in inputSchema ? inputSchema['x-ms-editor-options'].isRequestApiConnectionTrigger : undefined;
+        !!updatedManifest.properties?.inputs?.properties?.schema?.['x-ms-editor-options']?.isRequestApiConnectionTrigger;
 
       let schemaValue: Schema;
       let shouldMerge: boolean;
@@ -365,8 +364,7 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
           shouldMerge = false;
         }
       } else {
-        schemaValue = { ...currentSchemaValue, ...schemaToReplace };
-        shouldMerge = false;
+        continue;
       }
       safeSetObjectPropertyValue(updatedManifest.properties.outputs, outputLocation, schemaValue, shouldMerge);
     }
--- a/libs/designer/src/lib/core/utils/outputs.ts
+++ b/libs/designer/src/lib/core/utils/outputs.ts
@@ -21,7 +21,14 @@ import { convertOutputsToTokens } from './tokens';
 import { OperationManifestService } from '@microsoft/designer-client-services-logic-apps';
 import { generateSchemaFromJsonString, ValueSegmentType } from '@microsoft/designer-ui';
 import { getIntl } from '@microsoft/intl-logic-apps';
-import type { Expression, ExpressionFunction, ExpressionLiteral, OutputParameter, OutputParameters } from '@microsoft/parsers-logic-apps';
+import type {
+  Expression,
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -342,17 +349,26 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
...
-            outputLocation,
-            { ...currentSchemaValue, ...schemaToReplace },
-            true
-          );
+      if (isRequestApiConnectionTrigger && schemaToReplace && 'rows' in schemaToReplace) {
+        if ('rows' in currentSchemaValue) {
+          schemaValue = { ...currentSchemaValue, ...schemaToReplace };
+          shouldMerge = true;
+        } else {
+          schemaValue = currentSchemaValue;
```
wifun, line 364: isn't this the same as the current value. what's the point of calling `safeSetObjectPropertyValue` after this

isn't this like doing
```
const b = a;
a = b;
```

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -342,17 +349,26 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
 
       const currentSchemaValue = getObjectPropertyValue(updatedManifest.properties.outputs, outputLocation);
 
+      const inputSchema = updatedManifest.properties?.inputs?.properties?.schema;
```
wifun, line 352: why not

```
const isRequestApiConnectionTrigger = !!updatedManifest.....?.schema>?.['x-ms-editor-options']?.isRequestApiConnectionTrigger
```

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -342,17 +349,26 @@ export const getUpdatedManifestForSchemaDependency = (manifest: OperationManifes
...
-          safeSetObjectPropertyValue(
-            updatedManifest.properties.outputs,
-            outputLocation,
-            { ...currentSchemaValue, ...schemaToReplace },
-            true
-          );
+      if (isRequestApiConnectionTrigger && schemaToReplace && 'rows' in schemaToReplace) {
+        if ('rows' in currentSchemaValue) {
+          schemaValue = { ...currentSchemaValue, ...schemaToReplace };
+          shouldMerge = true;
```
wifun, line 362: Suppose someone added parameters using the floating editor. How does deleting the parameter work if this is always merging with the existing schema.

libs/designer/src/lib/core/utils/outputs.ts
wifun, line 352: just calling out that this logic is essentially hardcoding and assuming that all hybrid triggers will use `schema` as the name for the floating editor parameter

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: Yeah, the else can be skipped which is why previously I didn't have an else statement. Added it incase it isn't very clear

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: yes, the properties of the floatingActionMenu go inside the `schema` object. It is specifically defined to add it in the `schema` of the inputs. For now that's by design anyway

libs/designer/src/lib/core/utils/outputs.ts
wifun: If this was intentional why not just do a `continue`

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: `schemaToReplace` in the case of a floatingActionMenu is the value that's added in the rootObject. Incase there is a delete, the rootObject is updated accordingly. Incase the rootObject gets empty, the `schemaToReplace` would be an empty rootObject (Incase of a hybridTrigger, the rootObject would contain the 'rows' object, for manualTrigger there won't be). The merge would be a. if the currentSchemaValue is empty, then nothing to merge, b. if the currentSchemValue has a value then merge with emptyRootObject

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: Done

libs/designer/src/lib/core/utils/outputs.ts
bnarayanared: `continue` makes more sense. Changed

hartra344 (approved):

libs/designer/src/lib/core/utils/outputs.ts
wifun: I don't understand what you are saying, mostly because I don't know the terminology being used.

I'm picturing the currentSchemaValue containing something like:
```
{
"type": "object",
"properties": {
    "ID": {
        "type": "integer",
        "format": "int64",
        "description": "File Identifier"
    },
    "fileName": {
        "type": "string",
        "description": "File Name"
    }
}
```
and schemaToReplace is (after delete):
```
{
"type": "object",
"properties": {
    "ID": {
        "type": "integer",
        "format": "int64",
        "description": "File Identifier"
    }
}
```

will the resulting merge be without `fileName`?

Is this scenario possible and is somehow handled, that's my concern here.

bnarayanared: This scenario after delete will be merged and final result will be the following:
```{
"type": "object",
"properties": {
    "ID": {
        "type": "integer",
        "format": "int64",
        "description": "File Identifier"
    }
}```

I have tested this scenario

wifun: ok

PR #3201, opened 2023-08-25
feat(designer): Adding search service 2 to allow for custom search results

almaasrawi:
This PR adds an optional search service 2 with the expected functionality of:
* **search operations:** to allow the host to define its own search results for the recommendation panel search.
* **get runtime filters:** to allow the host to define a different set of runtime filters.
* **filter connector:** to allow the host to filter the connector based on one of the runtime filters from the above step.

The change has been tested for backward compatibility, and it is working as expected. If the host has defined the new search service 2, then that would be used, otherwise, a default search service 2 would be used to match the previous behavior.

I have also tested the change in a host with the new service, and it does meet the requirements on that.


libs/services/designer-client-services/src/lib/search.ts
```diff
@@ -40,3 +53,131 @@ export const SearchService = (): ISearchService => {
 export interface SearchResult {
   searchOperations: DiscoveryOperation<DiscoveryResultTypes>[];
 }
+
+let service2: ISearchService2;
```
hartra344, line 57: can we call it somehting other than SearchService2? Or just make these optional parts of SearchService?

hartra344 (changes requested):

libs/services/designer-client-services/src/lib/search.ts
almaasrawi: Making it part of search service would make the optional part of it a bit harder / uncleaner to split. How about `IOptionalSearchService` ?

libs/designer/src/lib/ui/panel/recommendation/searchView.tsx
```diff
@@ -18,99 +17,27 @@ type SearchViewProps = {
...
-    []
-  );
-
   useEffect(() => {
     if (searchTerm !== '') setIsLoadingSearchResults(true);
   }, [searchTerm]);
 
   useDebouncedEffect(
     () => {
-      if (!allOperations) return;
```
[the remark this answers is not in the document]
almaasrawi: the type going in is required, so I believe it's best to match the type. I don't know the history for why the check exists here despite the type, but I added a check in the default implementation to return empty results

hartra344 (approved):

PR #3278, opened 2023-09-07
Fix(data mapper): able to add new connections to expanded cards

DanielleCogs:

- bug fix

unable to add a second connection to an expanded card if the function has unlimited inputs (caused by mutating immutable state)
fixed! and added a relevant test
no


libs/data-mapper/src/lib/core/services/dataMapperApiService/index.ts
```diff
@@ -7,7 +7,7 @@ let service: IDataMapperApiService;
 
 export const defaultDataMapperApiServiceOptions = {
   baseUrl: 'http://localhost',
-  port: '7071',
+  port: '8000',
```
refortie, line 10: Isn't the default 7071?

refortie (approved):

libs/data-mapper/src/lib/core/services/dataMapperApiService/index.ts
DanielleCogs: nope it isn't, had to change it

PR #3287, opened 2023-09-09
feat(Data Mapper): Adding messaging to make input types more clear

DanielleCogs:

feature
confusing to the user that they do not know the expected input type of a function
1. added text to display expected input type in the panel
2. if there is a type mismatch error, new suggestion to the user to select a "Conversion" function that will output the data as the proper type
- no

- Screenshot 2023-09-09 at 2 17 21 PM


libs/data-mapper/src/lib/components/inputTypes/InputDropdown.tsx
```diff
@@ -310,10 +331,27 @@ export const InputDropdown = (props: InputDropdownProps) => {
...
+  if (isFunctionData(currentNode)) {
+    const allowedInputTypes = currentNode.inputs[inputIndex].allowedTypes;
+    inputType = allowedInputTypes.toString();
+    if (
+      allowedInputTypes.length === 3 &&
+      allowedInputTypes.includes(NormalizedDataType.Number) &&
+      allowedInputTypes.includes(NormalizedDataType.Decimal) &&
+      allowedInputTypes.includes(NormalizedDataType.Integer)
+    ) {
+      inputType = 'Numerical';
```
refortie, line 348: Use your constant here?

libs/data-mapper/src/lib/components/inputTypes/InputDropdown.tsx
```diff
@@ -283,14 +286,32 @@ export const InputDropdown = (props: InputDropdownProps) => {
             }
           } else {
             let someTypesMatched = false;
+            let possibleConversionFunctions = '';
             currentNode.inputs[isUnboundedInput ? 0 : inputIndex].allowedTypes.forEach((type) => {
+              const conversion = functions.find((func) => func.category === FunctionCategory.Conversion && func.outputValueType === type);
```
refortie, line 291: Maybe this would work, maybe it would. Could have a 2 tiered message: where we suggest a function where the input matches the incoming type and the output matches the destination and if we don't find a conversion that matches the input, do what you have here.

Example: Decimal -> String
Search the conversion functions for a function that takes in a decimal as an accepted type and outputs string.

That way we can cut down on the number of results we return. Thoughts?

libs/data-mapper/src/lib/utils/Connection.Utils.ts
```diff
@@ -250,49 +250,15 @@ export const isValidConnectionByType = (srcDataType: NormalizedDataType, tgtData
     return true;
   }
 
-  switch (tgtDataType) {
```
refortie, line 253: Don't you want to keep this? We only need to convert when the types are incompatible. Like a decimal can be used in a number with no conversion necessary

refortie (approved):

libs/data-mapper/src/lib/components/inputTypes/InputDropdown.tsx
DanielleCogs: So currently none of the conversion functions from the backend actually accept a specific type as input, only 'any', so we would have to get the backend on board with this.

libs/data-mapper/src/lib/utils/Connection.Utils.ts
DanielleCogs: The backend team changed allowed types to explicitly mention "decimal" and "number" so we don't have to infer anything on the frontend. So for your example we would still get the same result.
Ex: 
"allowedTypes": [
                        "Number",
                        "Integer",
                        "Decimal"
                    ],

libs/data-mapper/src/lib/components/inputTypes/InputDropdown.tsx
DanielleCogs: So we are waiting to discuss further with the dev and PM teams on this, will merge until we discuss it

Issue #3046, opened 2023-07-26
TestMap - Test button is disabled until you close and come back to test. 


kewear:
### Describe the Bug with repro steps

Invoke test, the test UX comes up. The test button is disabled since there is no instance. Paste an instance message in the provided space. Test is still disabled. Cloes the test UX and return to it. Test button is not enabled.

### Version

V 2.53.0

AB#24657235

Elaina-Lee:
Fixed with https://github.com/Azure/LogicAppsUX/pull/3337

PR #3337, opened 2023-09-18
fix(Data Mapper): test button is disabled until you close and come back to test

Elaina-Lee:

Bug fix 
User has to type something in the test panel, close it, then reopen it to be able to click on the test button
https://github.com/Azure/LogicAppsUX/issues/3046
As soon as user types something in the test panel, the test button gets enabled
no
NA

### DIFF
--- a/libs/data-mapper/src/lib/components/testMapPanel/TestMapPanel.tsx
+++ b/libs/data-mapper/src/lib/components/testMapPanel/TestMapPanel.tsx
@@ -134,7 +134,26 @@ export const TestMapPanel = ({ mapDefinition, isOpen, onClose }: TestMapPanelPro
 
   const inputDataOptions = useMemo(() => [{ key: 'pasteSample', text: pasteFromSampleLoc }], [pasteFromSampleLoc]);
 
-  const testMap = async () => {
+  useEffect(() => {
+    const generateXsltAsync = async () => {
+      let generatedXslt = '';
+      if (isOpen && !isNullOrEmpty(mapDefinition)) {
+        try {
+          generatedXslt = await generateDataMapXslt(mapDefinition);
+        } catch (error) {
+          console.log(error);
+        }
+      }
+
+      setCurrentXslt(generatedXslt);
+    };
+
+    generateXsltAsync();
+  }, [isOpen, mapDefinition]);
+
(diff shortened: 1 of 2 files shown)


libs/data-mapper/src/lib/components/testMapPanel/TestMapPanel.tsx
```diff
@@ -233,7 +233,7 @@ export const TestMapPanel = ({ mapDefinition, isOpen, onClose }: TestMapPanelPro
       </Stack>
     );
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fileXslt, isMismatchedXslt]);
+  }, [fileXslt, testMapInput, isMismatchedXslt]);
```
DanielleCogs, line 236: oh, the reason we did not see this earlier is because of the above line that suppresses the error. Can you remove it?

Elaina-Lee: removed it

libs/data-mapper/src/lib/components/testMapPanel/TestMapPanel.tsx
ccastrotrejo, line 1: There are some Github actions warnings regarding the `testMap` function

libs/data-mapper/src/lib/components/testMapPanel/TestMapPanel.tsx
Elaina-Lee: image

Disabling the suppress error requires that I include everything in the dependency list including testMap. It seems like including testMap in the dependency list gives this error, which is why the suppress error script was there. I brought back the suppress error dialog and made sure to add all the other dependencies suggested. Would this work?

Elaina-Lee: @refortie would you have any insight?

refortie: Might need to make testMap into a useCallback, which should prevent it from re-rendering all the time

Elaina-Lee: @refortie Got it, changed it accordingly. I tested that things work as expected.

DanielleCogs (approved):

Issue #3361, opened 2023-09-21
[@formatjs/intl Error MISSING_DATA] Missing locale data for locale: "gl-ES" in Intl.NumberFormat


Ram-G:
### Describe the Bug with repro steps 

1. Open v3 designer in Power Automate. 
2. Change language to "galego" (gl-ES). 
3. Observe error banner at the top of the screen. 

Expected:

 Designer loads successfully.

Actual:

 Designer fails to load and error banner is displayed with the following message:

Produciuse un erro ao renderizar [@formatjs/intl Error MISSING\_DATA] Missing locale data for locale: "gl-ES" in Intl.NumberFormat. Using default locale: "gl-ES" as fallback. See [https://formatjs.io/docs/react-intl#runtime-requirements](<https://formatjs.io/docs/react-intl#runtime-requirements>) for more details

### What type of Logic App Is this happening in? 

Power Automate

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

![image](<https://github.com/Azure/LogicAppsUX/assets/10837129/c040497b-3be5-48f4-ab2d-da22cd61e8f3>)

### Browser 

Edge

### Additional context 

In parallel to filing this issue, I am creating a pull request to put in a stopgap fix to not throw error for this missing locale data. This means the designer loads, just with some english (or other fallback) text strings instead of in the desired language.

Stack trace I got for the error from code -- libs\designer\src\lib\core\DesignerProvider.tsx line 36:

{

 "code": "MISSING\_DATA"

 }

"[@formatjs/intl Error MISSING\_DATA] Missing locale data for locale: \"gl-ES\" in Intl.NumberFormat. Using default locale: \"gl-ES\" as fallback. See [https://formatjs.io/docs/react-intl#runtime-requirements](<https://formatjs.io/docs/react-intl#runtime-requirements>) for more details\n"

[...]

PR #3362, opened 2023-09-21
fix(designer): Don't throw for MISSING_DATA error in IntlProvider

Ram-G:

Prevents throwing  for MISSING_DATA error in IntlProvider. This makes it so if language/locale data is missing, we log the error and allow fallback to default language/locale, instead of failing to load the designer.

We throw the error which causes designer to not load. Screenshot below.

We log MISSING_DATA error instead of throwing, as this does not look to be a fatal error.

No.

Current bugged scenario looks like this:
image

After this fix, the designer gets loaded instead.

To fix the underlying issue of missing data, I've filed the following issue: https://github.com/Azure/LogicAppsUX/issues/3361

### DIFF
--- a/libs/designer/src/lib/core/DesignerProvider.tsx
+++ b/libs/designer/src/lib/core/DesignerProvider.tsx
@@ -8,6 +8,7 @@ import { AzureThemeLight } from '@fluentui/azure-themes/lib/azure/AzureThemeLigh
 import { ThemeProvider } from '@fluentui/react';
 import { FluentProvider, webDarkTheme, webLightTheme } from '@fluentui/react-components';
 import type { OnErrorFn as OnIntlErrorFn } from '@formatjs/intl';
+import { LogEntryLevel, LoggerService } from '@microsoft/designer-client-services-logic-apps';
 import { IntlProvider } from '@microsoft/intl-logic-apps';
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/DesignerProvider.tsx
```diff
@@ -35,6 +35,9 @@ export const DesignerProvider = ({ locale = 'en', options, children }: DesignerP
   const onError = useCallback<OnIntlErrorFn>((err) => {
     if (err.code === 'MISSING_TRANSLATION') {
```
Eric-B-Wu, line 36: would we be able to replace this with our browser logger instead of a console log?
```
    if (err.code === 'MISSING_TRANSLATION' || err.code === 'MISSING_DATA') {
      LoggerService().log({
        level: LogEntryLevel.Warning,
        area: 'Designer Provider',
        message: err.message,
      });
      return;
    }
```

ek68794998: Should it also include the code? ``message: `${err.code} - ${err.message}`,``

Ram-G: Addressed both comments.

Eric-B-Wu (approved):

PR #3365, opened 2023-09-21
fix(designer): Fix exception due to usage of LoggingService in DesignerProvider

Ram-G:

Bugfix.

LoggingService throws because it has not been initialized.

Removed usage of LoggingService so code doesn't throw.

No.

Sample of logging in console after this change:

image

AB#25238017


Wu-Bowen (approved):

Eric-B-Wu (approved):

ek68794998:
@Eric-B-Wu I don't want to hold up this PR as it's a blocker for us, but I'm curious if there's a way to ensure that the logger service is initialized beforehand? It would be far more preferable to log this error in a way we can track.

PR #3378, opened 2023-09-25
feat(designer): Added Node Copy/Paste Functionality

Eric-B-Wu:
Feature Request - https://github.com/Azure/LogicAppsUX/discussions/3310

There is a todo of preventing the paste of a copied node where the node dependencies are invalid 

AB#25260721


rllyy97:
How do scope nodes deal with copy / paste?  Do we include all contained nodes when pasting or does it paste empty?

Eric-B-Wu:
> How do scope nodes deal with copy / paste? Do we include all contained nodes when pasting or does it paste empty?

currently, we don't handle any scope nodes, I only added the copy functionality on operation card nodes, as I felt like doing the subgraph stuff might need to be much more extensive

rllyy97 (approved):

PR #3380, opened 2023-09-25
fix(designer): Fix combobox filtering to use display name instead of value

almaasrawi:
Fix combobox filtering to use display name instead of value to have parity with v1 designer

AB#25265513


bnarayanared (approved):

bnarayanared:
This is the only place the change is required right?

almaasrawi:
> This is the only place the change is required right?

Correct, I tested it and it's working.

hartra344 (approved):

PR #3398, opened 2023-09-29
fix(designer): Fix designer load for existing hybrid triggers

almaasrawi:
Fix designer load for existing hybrid triggers. The issue is that items in hybrid triggers would be serialized in the location: `inputs.schema.rows.items`, while for a flow that's saved from the v1 designer the items woudl be in `inputs.schema.items` directly.

To repro the issue: save a flow with a hybrid trigger on the v1 designer, open the same flow on the v3 designer, and click on the trigger, where you'll see an exception and everything disappearing.

AB#25323883

### DIFF
--- a/libs/designer-ui/src/lib/floatingactionmenu/floatingactionmenuinputs/helper.ts
+++ b/libs/designer-ui/src/lib/floatingactionmenu/floatingactionmenuinputs/helper.ts
@@ -65,7 +65,7 @@ export function deserialize(value: ValueSegment[], isRequestApiConnectionTrigger
 
   itemsProperties = rootObject;
   if (isRequestApiConnectionTrigger) {
-    itemsProperties = rootObject.properties ? rootObject.properties.rows.items : rootObject.rows.items;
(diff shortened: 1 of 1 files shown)


wifun:
Can you have a description on how to hit this issue previously?

libs/designer-ui/src/lib/floatingactionmenu/floatingactionmenuinputs/helper.ts
```diff
@@ -65,7 +65,7 @@ export function deserialize(value: ValueSegment[], isRequestApiConnectionTrigger
 
   itemsProperties = rootObject;
   if (isRequestApiConnectionTrigger) {
-    itemsProperties = rootObject.properties ? rootObject.properties.rows.items : rootObject.rows.items;
+    itemsProperties = rootObject.properties?.rows?.items || rootObject.rows?.items || itemsProperties;
```
wifun, line 68: Can you have a description on how to hit this issue previously?

almaasrawi: Done.

Eric-B-Wu (approved):

PR #3404, opened 2023-10-03
fix(designer): Add gateway service config to allow hiding subscription dropdown for gateway connection creation

pranaydubeymicrosoft:
This change adds the option to send a custom config as part of gateway connection creation. As such we can specify with the config whether we want the subscription dropdown to be enabled (default) or not. When it is disabled through the config, the gateway dropdown should remain enabled and visible. The new methods and params are optional and will not affect current behavior. 

AB#25351748


libs/services/designer-client-services/src/lib/base/gateway.ts
```diff
@@ -29,7 +29,7 @@ export class BaseGatewayService implements IGatewayService {
   }
 
   public getGateways(subscriptionId: string, connectorName: string): Promise<Gateway[]> {
-    if (!subscriptionId || !connectorName) return Promise.resolve([]);
+    if (!connectorName) return Promise.resolve([]);
```
ek68794998, line 32: This might need a config check here.

Something like:

```ts
const isSubscriptionRequired = !this.getConfig?.().disableSubscriptionLookup;
if ((isSubscriptionRequired && !subscriptionId) || !connectorName) {
  return Promise.resolve([]);
}
```

libs/services/designer-client-services/src/lib/base/gateway.ts
```diff
@@ -29,7 +29,7 @@ export class BaseGatewayService implements IGatewayService {
   }
 
   public getGateways(subscriptionId: string, connectorName: string): Promise<Gateway[]> {
```
ek68794998, line 31: `subscriptionId: string | undefined`

libs/services/designer-client-services/src/lib/gateway.ts
```diff
@@ -1,17 +1,22 @@
...
   /**
    * Gets subscriptions.
    */
-  getSubscriptions(): Promise<Subscription[]>;
+  getSubscriptions(): Promise<Subscription[] | undefined>;
+
+  /**
+   * Gets configuration values for GatewayService.
+   */
+  getConfig?(): Promise<Partial<GatewayServiceConfig>>;
```
ek68794998, line 19: Does not need to be `Partial<>`.

libs/services/designer-client-services/src/lib/base/gateway.ts
pranaydubeymicrosoft: This is the base implementation of getGateways. Since our service will have it's own implementation we should be able to leave this, right? The default method of grabbing gateways here relies on this parameter.

ek68794998: The implementation should align with the interface it inherits from, IMO, because you could call the interface with `undefined` while the implementation is expecting a string. I'm not sure why TypeScript doesn't enforce this, but they've chosen not to.

Eric-B-Wu (approved):

libs/utils/src/lib/models/gatewayServiceConfig.ts
```diff
@@ -0,0 +1,3 @@
+export interface GatewayServiceConfig {
```
rllyy97, line 1: Let's leave this in the services package instead of the utils package, just have it below the `IGatewayService` interface in `libs/services/designer-client-services/src/lib/gateway.ts`

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/createConnectionTab.tsx
```diff
@@ -52,6 +57,8 @@ const CreateConnectionTab = () => {
   const [selectedSubscriptionId, setSelectedSubscriptionId] = useState('');
   const gatewaysQuery = useGateways(selectedSubscriptionId, connector?.id ?? '');
   const availableGateways = useMemo(() => gatewaysQuery.data, [gatewaysQuery]);
+  const gatewayServiceConfigQuery = useGatewayServiceConfig();
+  const gatewayServiceConfig = useMemo(() => gatewayServiceConfigQuery.data, [gatewayServiceConfigQuery]);
```
rllyy97, line 61: Adjustment from the other hook suggestion

```suggestion
  const gatewayServiceConfig = useGatewayServiceConfig();
```

libs/designer/src/lib/core/state/connection/connectionSelector.ts
```diff
@@ -46,6 +46,8 @@ export const useGateways = (subscriptionId: string, connectorName: string) => {
 
 export const useSubscriptions = () => useQuery('subscriptions', async () => GatewayService().getSubscriptions());
 
+export const useGatewayServiceConfig = () => useQuery('gatewayServiceConfig', async () => GatewayService().getConfig?.() ?? {});
```
rllyy97, line 49: This doesn't need to be a `useQuery`, we can just make it a `useMemo` since we're not actually querying data, it's just a read of the config from the service

```suggestion
export const useGatewayServiceConfig = () => useMemo(() => GatewayService().getConfig?.() ?? {}, []);
```

rllyy97 (commented):
Once you get these fixed I'll approve, looks good otherwise 👍

rllyy97 (approved):

pranaydubeymicrosoft:

pranaydubeymicrosoft:
> @pranaydubeymicrosoft please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

PR #3436, opened 2023-10-11
fix(Data Mapper): User able to paste path directly into "Select existing"

Elaina-Lee:

Bug fix
Users have inconsistent behavior of at times being able to paste local path to add schema panel. If this works, reloading map will crash since the schema cannot be found in Artifacts/Schema folder.
Users will not be able to paste anything to the ComboBox to prevent local path pasting.
no
n/a

AB#25434352


Elaina-Lee:
Through discussions, we had two choices of:
1) Allowing local path pasting then copying the file into Artifacts/Schema (with a prompt of notifying users).
2) Not allowing paste, only letting users type into the panel ComboBox.

For the time being, we decided to proceed with this option (2).

PR #3473, opened 2023-10-18
feat(designer): Add option to hide expression in token picker

ccastrotrejo:

### Main Code Changes
- Add `allowExpression` property to `TokenTypeAheadPlugin` and `TokenPickerButtonProps` components
- Set `allowExpression` property to as optional and default to true

**Please Include Screenshots or Videos of the intended change**:

#### Example of only allowing dynamic content token
Screenshot 2023-10-18 at 11 43 05 AM

#### Single token option for different input height
Screenshot 2023-10-18 at 11 43 23 AM

#### Single token hover example
Screenshot 2023-10-18 at 11 43 30 AM

#### Multiple token options, different input heights and hover example
Screenshot 2023-10-18 at 11 56 25 AM
Screenshot 2023-10-18 at 11 55 08 AM

AB#25526332


Eric-B-Wu:
can you also just send a screenshot of what it looks like with expression enabled as well?

ccastrotrejo:
> can you also just send a screenshot of what it looks like with expression enabled as well, just want to make sure there isn't a UI regression?

Sure, thanks for double checking

andimarc:
Does this also block interpolated strings, or will you make that change in a separate PR?

Eric-B-Wu:
also this is a nit, but can we actually change this prop to hideExpression instead of allowExpression, because by default we should not be hiding the expression, and only when this prop is set should we be hiding the expressions (as opposed to having the default value set as true)

ccastrotrejo:
> Does this also block interpolated strings, or will you make that change in a separate PR?

Just to get it on record. We are going to handle that in the other repo. Thanks for pointing this out.

Eric-B-Wu (approved):

PR #3478, opened 2023-10-18
fix(Designer): Do not update swagger parameter format when in is unde…

jaimereyesmx:
…fined

Fix issue where dynamic inputs is not updated properly store in the Redux state due to due to inputParameter.in being undefined.

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
+++ b/libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
@@ -287,7 +287,7 @@ export async function getDynamicInputsFromSchema(
   // We need to reformat to the below string:
   //     body.$.content.appId
   for (const inputParameter of dynamicInputs) {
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -287,7 +287,7 @@ export async function getDynamicInputsFromSchema(
   // We need to reformat to the below string:
   //     body.$.content.appId
   for (const inputParameter of dynamicInputs) {
-    if (isOpenApiParameter(inputParameter)) {
+    if (isOpenApiParameter(inputParameter) && inputParameter.in) {
```
rllyy97, line 290:
```suggestion
    if (isOpenApiParameter(inputParameter) && inputParameter?.in) {
```

rllyy97 (changes requested):

rllyy97 (approved):

PR #3505, opened 2023-10-23
fix(Designer): Fixes for single click installation bugs

YogiJogiFresh:
- Added new timeout setting (azureLogicAppsStandard.dependencyTimeout) for dependency validate and install
  - This will prevent the "hanging" loading notification for the validate and install command
image

- Added  "omnisharp.dotNetCliPaths" setting at workspace scope for csharp extension
  - fixes debugging (F5) - coreclr attaching to function process by specifying path to dotnet binary path.
  - **https://github.com/dotnet/vscode-csharp/pull/6559**
 
- Added "terminal.integrated.env" setting for linux and osx
  - Allows for the use of dotnet in vscode terminal

- Changed **isFunctionProject** to use 'host.json' and 'local.settings.json' as an indicator that this is a functions project
  - avoid activateAzurite in LogicAppsUX project

- Fix async/await with validate and install functions, set command functions, and start design time functions
  - Fixes the issues with empty funcCoreToolsBinary path


apps/vs-code-designer/src/app/utils/delay.ts
```diff
@@ -4,5 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 export async function delay(ms: number): Promise<void> {
-  await new Promise<void>((resolve: () => void): NodeJS.Timer => setTimeout(resolve, ms));
+  return await new Promise<void>((_, reject) => {
+    setTimeout(() => {
+      reject(new Error());
```
ccastrotrejo, line 9: Is there any particular reason on why we are rejecting instead of resolving after the timeout?

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/utils/delay.ts
YogiJogiFresh: yep, the parent function calling this timeout will catch the error.

YogiJogiFresh: thinking on it...the name "delay" would be confusing.

ccastrotrejo: Yep, thats confusing because its throwing an error everytime after the time. Also, there is an already other delay function.
Also I just noticed that you are updating the function, however I couldn't notice where are you using this function in these changes

ccastrotrejo (approved):

PR #3525, opened 2023-10-25
feat(designer): Add support for recommendation view to work without depending on prefetched data

almaasrawi:
Introduce new methods in the search service to help the recommendation view work without relying on prefetched data. this includes:

* `getOperationById` to help retrieve the operation by id upon click.
* `getOperationsByConnector` to help retrieve the list of operations for a specific connector, when the connector header is clicked on from the browse or search views.

The change has been smoke tested for Power Automate integration without returning any operation in the prefetch call, and completely relying on the new method callbacks.

AB#25597550

### DIFF
--- a/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
+++ b/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
@@ -51,7 +51,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
 
   const { data: allOperations, isLoading: isLoadingOperations } = useAllOperations();
   const [selectedOperation, setSelectedOperation] = useState<DiscoveryOperation<DiscoveryResultTypes> | undefined>(undefined);
-  const [isWaitingOperationsGroup, setIsWaitingOperationGroup] = useState<boolean>(false);
+  const [isLoadingOperationsGroup, setIsLoadingOperationGroup] = useState<boolean>(false);
 
   const selectedOperationGroupId = useSelectedSearchOperationGroupId();
   const { data: allConnectors } = useAllConnectors();
@@ -72,12 +72,15 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
           })
         );
 
-    searchResultPromise.then((filteredOps) => {
-      setAllOperationsForGroup(filteredOps);
-      setIsWaitingOperationGroup(false);
-    });
+    setIsLoadingOperationGroup(true);
+    searchResultPromise
+      .then((filteredOps) => {
+        setAllOperationsForGroup(filteredOps);
+      })
+      .finally(() => {
+        setIsLoadingOperationGroup(false);
+      });
     setSelectionState(SELECTION_STATES.DETAILS);
-    setIsWaitingOperationGroup(true);
   }, [selectedOperationGroupId, allOperations, filters]);
 
   const navigateBack = useCallback(() => {
@@ -176,7 +179,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
               groupOperations={allOperationsForGroup}
               filters={filters}
               onOperationClick={onOperationClick}
-              isLoading={isLoadingOperations || isWaitingOperationsGroup}
+              isLoading={isLoadingOperations || isLoadingOperationsGroup}
               displayRuntimeInfo={displayRuntimeInfo}
             />
           ) : null,
--- a/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
+++ b/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
@@ -51,7 +51,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
 
   const { data: allOperations, isLoading: isLoadingOperations } = useAllOperations();
   const [selectedOperation, setSelectedOperation] = useState<DiscoveryOperation<DiscoveryResultTypes> | undefined>(undefined);
-  const [isLoadingOperationsGroup, setIsLoadingOperationGroup] = useState<boolean>(false);
+  const [isLoadingOperationGroup, setIsLoadingOperationGroup] = useState<boolean>(false);
 
   const selectedOperationGroupId = useSelectedSearchOperationGroupId();
   const { data: allConnectors } = useAllConnectors();
@@ -179,7 +179,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
               groupOperations={allOperationsForGroup}
               filters={filters}
               onOperationClick={onOperationClick}
-              isLoading={isLoadingOperations || isLoadingOperationsGroup}
+              isLoading={isLoadingOperations || isLoadingOperationGroup}
               displayRuntimeInfo={displayRuntimeInfo}
             />
           ) : null,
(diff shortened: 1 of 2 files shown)


ek68794998:
Discussion for reference: #3524

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -49,22 +49,32 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
...
+          })
+        );
+
+    searchResultPromise.then((filteredOps) => {
+      setAllOperationsForGroup(filteredOps);
+      setSelectionState(SELECTION_STATES.DETAILS);
     });
-    setAllOperationsForGroup(filteredOps);
-    setSelectionState(SELECTION_STATES.DETAILS);
   }, [selectedOperationGroupId, allOperations]);
```
wifun, line 78: Would be good to address the warning `React Hook useEffect has a missing dependency: 'filters'. Either include it or remove the dependency array`

It might be worth splitting this up into separate useEffects for if `searchOperation` is defined vs. not defined? My concern is that if anything in the dep array changes, for us this would cause a call to the host (and possibly interfere with the saved results list in PPUX)

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -94,21 +104,27 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
...
+      const searchOperation = SearchService().getOperationById?.bind(SearchService());
+
+      const searchResultPromise = searchOperation
+        ? searchOperation(id)
+        : Promise.resolve((allOperations ?? []).find((o) => (apiId ? o.id === id && o.properties?.api?.id === apiId : o.id === id)));
+
+      searchResultPromise.then((operation) => {
+        if (!operation) return;
+        dispatch(selectOperationId(operation.id));
+        setSelectedOperation(operation);
```
wifun, line 116: what is the need for this new line `setSelectedOperation(operation);`

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -49,22 +49,32 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
...
+      ? searchOperation(selectedOperationGroupId, filters['actionType']?.toLowerCase())
+      : Promise.resolve(
+          (allOperations ?? []).filter((operation) => {
+            const apiId = operation.properties.api.id;
+            return areApiIdsEqual(apiId, selectedOperationGroupId);
+          })
+        );
+
+    searchResultPromise.then((filteredOps) => {
+      setAllOperationsForGroup(filteredOps);
```
wifun, line 75: should we consider some kind of loading state now that this is async, maybe in long term fix

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: I don't believe the split is necessary, the filters dropdown lists don't show up on the connector view.

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -94,21 +104,27 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
 
   const onOperationClick = useCallback(
     (id: string, apiId?: string) => {
-      const operation = (allOperations ?? []).find((o: DiscoveryOperation<DiscoveryResultTypes>) => {
-        return apiId ? o.id === id && o.properties?.api?.id === apiId : o.id === id;
+      const searchOperation = SearchService().getOperationById?.bind(SearchService());
+
+      const searchResultPromise = searchOperation
+        ? searchOperation(id)
+        : Promise.resolve((allOperations ?? []).find((o) => (apiId ? o.id === id && o.properties?.api?.id === apiId : o.id === id)));
```
rllyy97, line 111: I think doing this as a `.find` over the `allOperations` array first, then if the result is null doing the `searchOperation` call (if available) would make more sense if we're going to support a more hybrid approach to prefetch/realtime searching, but until we want to get into all that I think this works better

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: the selected operation used to be set synchronously (deleted line 52), now that it became async I needed to change that into a set method within an effect.

rllyy97 (approved):

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: I don't really believe we should depend on the data for the codepath here. What if the data really is empty (no match)? or what if the operations were still loading? 

 I believe the better choice here is the explicit code choice: did the host implement that method or not.

almaasrawi: Talked offline, the possible future hybrid approach would need to be settled a bit first before we can choose what to do here. For now given the fork, this is the best approach.

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: Thanks for the suggestion, done.

wifun (approved):

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -49,23 +49,36 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
...
+        );
+
+    searchResultPromise.then((filteredOps) => {
+      setAllOperationsForGroup(filteredOps);
+      setIsWaitingOperationGroup(false);
     });
-    setAllOperationsForGroup(filteredOps);
     setSelectionState(SELECTION_STATES.DETAILS);
-  }, [selectedOperationGroupId, allOperations]);
+    setIsWaitingOperationGroup(true);
```
rllyy97, line 80: This might cause a race condition for `isWaitingOperationsGroup` to remain set to true if the search results promise finishes too quick, can you move the `setIsWaitingOperationGroup(true)` up above the `searchResultPromise` just to be safe, and also add a catch in case the promise fails that also resets the `isWaitingOperationsGroup`

Also just a nit but can you rename the state variable to `isLoadingOperationGroup` instead

rllyy97 (changes requested):

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: Thanks for the comment Riley. Like we agreed offline, the race condition isn't something to worry about given the single threaded nature of JS. I moved it up for readability, did the rename and the `finally`

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -49,23 +49,39 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
 
   const [selectionState, setSelectionState] = useState<SelectionState>(SELECTION_STATES.SEARCH);
 
-  const selectedOperationId = useSelectedSearchOperationId();
   const { data: allOperations, isLoading: isLoadingOperations } = useAllOperations();
-  const selectedOperation = allOperations.find((o) => o.id === selectedOperationId);
+  const [selectedOperation, setSelectedOperation] = useState<DiscoveryOperation<DiscoveryResultTypes> | undefined>(undefined);
+  const [isLoadingOperationsGroup, setIsLoadingOperationGroup] = useState<boolean>(false);
```
rllyy97, line 54: Sorry but one last nit, the set is fine but the reference still has an S
```suggestion
  const [isLoadingOperationGroup, setIsLoadingOperationGroup] = useState<boolean>(false);
```

almaasrawi: Thanks for the catch. Done.

rllyy97 (approved):

PR #3542, opened 2023-10-27
fix(designer): Remove action type filtering for where the prefetched operations is empty

almaasrawi:
Remove action type filtering for where the prefetched operations is empty, given how browse would be empty otherwise. A subsequent change needed here is to have the connector details view not respect the 'actions' filter type in such case, given how there'd be no other way for users in the browse view to list triggers when trying to add actions.

### DIFF
--- a/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
+++ b/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
@@ -84,7 +84,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
         setIsLoadingOperationGroup(false);
       });
     setSelectionState(SELECTION_STATES.DETAILS);
-  }, [selectedOperationGroupId, allOperations, filters]);
+  }, [selectedOperationGroupId, allOperations, filters, hideActionTypeFilter]);
 
   const navigateBack = useCallback(() => {
     dispatch(selectOperationGroupId(''));
--- a/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
+++ b/libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
@@ -185,7 +185,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
               onOperationClick={onOperationClick}
               isLoading={isLoadingOperations || isLoadingOperationGroup}
               displayRuntimeInfo={displayRuntimeInfo}
-              respectActionsFilter={!hideActionTypeFilter}
+              ignoreActionsFilter={hideActionTypeFilter}
             />
           ) : null,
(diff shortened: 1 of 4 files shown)


libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -57,14 +57,17 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
   const { data: allConnectors } = useAllConnectors();
   const selectedConnector = allConnectors?.find((c) => c.id === selectedOperationGroupId);
 
+  // hide actions type filter if we don't have any operations for the browse view
+  const hideActionTypeFilter = (!allOperations || allOperations.length === 0) && !searchTerm;
```
rllyy97, line 61: With the current setup it looks like you're going to be seeing the action type filter pop in and out or rendering when you have things in the search bar and not, rather than just disabling the field.
I'm fine with either, logic apps will not be using this scenario, just wanted to bring it up.

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -181,6 +185,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
               onOperationClick={onOperationClick}
               isLoading={isLoadingOperations || isLoadingOperationGroup}
               displayRuntimeInfo={displayRuntimeInfo}
+              respectActionsFilter={!hideActionTypeFilter}
```
rllyy97, line 188: Nit but I'd prefer the prop here to be inverted, ie:
`ignoreActionsFilter={hideActionTypeFilter}`

rllyy97 (approved):

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: Yeah I believe we're okay with that given the change in the list nature (one list shows connectors, the other shows operations)

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
almaasrawi: Makes sense, done.

libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -86,6 +89,7 @@ export const RecommendationPanelContext = (props: RecommendationPanelContextProp
   const navigateBack = useCallback(() => {
```
wifun, line 89: warning for missing dependency (i don't know why github doesn't let me put the comment on line 87)

PR #3564, opened 2023-10-30
fix(vscode): Set default path when opt-out

ccastrotrejo:
### Main Code Changes
- Set the system wide default binary dependencies path when user opts out

AB#25634898

### DIFF
--- a/apps/vs-code-designer/src/app/utils/binaries.ts
+++ b/apps/vs-code-designer/src/app/utils/binaries.ts
@@ -3,14 +3,18 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 import {
+  DependencyDefaultPath,
   DependencyVersion,
   Platform,
   autoBinariesInstallationSetting,
   defaultDependencyPathValue,
   dependenciesPathSettingKey,
   dependencyTimeoutSettingKey,
+  dotNetBinaryPathSettingKey,
   dotnetDependencyName,
+  funcCoreToolsBinaryPathSettingKey,
   funcPackageName,
+  nodeJsBinaryPathSettingKey,
 } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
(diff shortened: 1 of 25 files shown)


apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -409,7 +413,7 @@ export async function promptInstallBinariesOption(context: IActionContext) {
   if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
     const workspace = await getWorkspaceFolder(context);
     const projectPath = await tryGetFunctionProjectRoot(context, workspace);
-    const message = localize('useBinaries', 'Always validate and install the latest dependency binaries at launch');
+    const message = localize('useBinaries', 'Always validate and install the latest dependency binaries at launch TEST');
```
Eric-B-Wu, line 416: the TEST part should be removed, right?

Eric-B-Wu (approved):

apps/vs-code-designer/src/package.json
```diff
@@ -847,7 +847,7 @@
           },
           "azureLogicAppsStandard.autoBinariesInstallation": {
             "type": "boolean",
-            "description": "Start using binaries installation",
+            "description": "Start using binaries installation Test",
```
Eric-B-Wu, line 850: should be removed?

ccastrotrejo, line 850: @ecfan Could you please review this string?
This is the text for the configuration setting that will tell the user if they select the auto validation/installation of the binary dependencies

apps/vs-code-designer/src/app/utils/binaries.ts
ccastrotrejo: Yep, just made the change so Esther can review the strings 😅

apps/vs-code-designer/src/package.json
ccastrotrejo: Yep, just made the change so Esther can review the strings 😅

apps/vs-code-designer/src/app/utils/binaries.ts
ccastrotrejo, line 416: @ecfan Could you please review this string?
This is the text for the warning message that will tell the user if they select the auto validation/installation of the binary dependencies
Screenshot 2023-10-30 at 9 41 12 AM

ecfan:
#assign:ecfan

apps/vs-code-designer/src/package.json
ecfan, line 850: We don't need "Test", correct?

```suggestion
            "description": "Start using binaries installation",
```

apps/vs-code-designer/src/app/utils/binaries.ts
ecfan, line 416:
```suggestion
    const message = localize('useBinaries', 'Always validate and install the latest dependency binaries at launch');
```

apps/vs-code-designer/src/package.json
ccastrotrejo: Yes, no need for the TEST. Sorry I just added for the PR to be able to mark where I wanted you to review the string. 
Here it is how its going to look like
Screenshot 2023-10-30 at 11 01 40 AM

ecfan:
[like]  Esther Fan reacted to your message:
________________________________

ecfan (approved):

PR #3565, opened 2023-10-30
feat(copilot): support for dark mode in copilot

Eric-B-Wu:
dark

AB#25634770


hartra344 (approved):

hartra344:
We may need to adjust the AI disclaimer at hte bottom for a better a11y contrast ratio

PR #3573, opened 2023-10-30
feat(Data Mapper): Add Data Mapper Vs Command

Elaina-Lee:

adding Data Mapper extension command & removing unused vs-code-data-mapper

n/a


.vscode/launch.json
```diff
@@ -20,21 +20,11 @@
       "type": "extensionHost",
```
ccastrotrejo, line 20: I don't know if its worth keeping this second command. As far as I remember this one was created to actually debug it , whereas the last one was only for running it. And now as there is only one single vscode extension, the upper one will do the work of debugging it. Unless there are specific scenarios to run and keep this one.

extension-copy-svgs.js
```diff
@@ -17,11 +17,7 @@ const copyDoc = async (projectPath) => {
  */
 const copyFiles = async (argumentsArray) => {
   let projectPath = '';
```
ccastrotrejo, line 19: NIT: I think you can declare the projectPath variable as `const projectPath = 'dist/apps/vs-code-designer';` since the beginning as it will not change anymore. And remove the argumentsArray

package.json
```diff
@@ -23,19 +23,15 @@
...
     "build:vscode-react": "nx run vs-code-react:build",
     "build:vscode-designer": "nx run vs-code-designer:build && nx run vs-code-data-mapper-react:build && nx run vs-code-designer-react:build && nx run vs-code-react:build",
-    "build:vscode-datamapper": "nx run vs-code-data-mapper:build && nx run vs-code-data-mapper-react:build",
+    "build:vscode-datamapper": "nx run vs-code-designer:build && nx run vs-code-data-mapper-react:build",
     "build:utils": "nx run utils:build",
     "build:docs": "nx build logic-apps-ux-docs",
     "start:docs": "nx serve logic-apps-ux-docs",
     "bump": "standard-version --no-verify",
     "vscode:designer:pack:step1": "cd ./dist/apps/vs-code-designer && npm install",
     "vscode:designer:pack:step2": "node extension-copy-svgs.js designer",
```
ccastrotrejo, line 32: NIT: We can remove the `designer` word since we are not using it any more in the `extension-copy-svgs.js` (this is passed as the `argunmentsArray` )

ccastrotrejo (approved):

.vscode/launch.json
Elaina-Lee: @DanielleCogs Any thoughts on this - was there a reason you wanted to keep this separate data mapper command?

package.json
Elaina-Lee: Got it, so just run the `nx run vs-code-data-mapper-react:build` for the vscode-datamapper. Checked it is working as expected and pushed.

extension-copy-svgs.js
Elaina-Lee: Done!

package.json
ccastrotrejo: ohhh ok if that works fine its great. But I meant in the `vscode:designer:pack:step2` script (line 32) sorry if I caused misunderstanding.

.vscode/launch.json
DanielleCogs: I was thinking that this could save us some time by not having to build the other projects at every update! Carlos lmk if that makes sense here

DanielleCogs (approved):

ccastrotrejo (approved):

package.json
Elaina-Lee: Modified to remove "designer" argument from step 2 as per your explanation!

PR #3499, opened 2023-10-20
fix(Designer): Update Combo box to sort alphabetically.

Ryanb58:
- **Please check if the PR fulfills these requirements**

* [X] The commit message follows our guidelines
* [x] Tests for the changes have been added (for bug fixes/features)
* [ ] Docs have been added / updated (for bug fixes / features)

- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

- **What is the current behavior?** (You can also link to an open issue here)

- **What is the new behavior (if this is a feature change)?**

- **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

- **Please Include Screenshots or Videos of the intended change**:

AB#25556424

PR #3582, opened 2023-10-31
fix(Designer): Sort parameters alphabetically in `designer` instead of inside `designer-ui`. 

Ryanb58:

Adding back the default sort alphabetically(ASC) to dropdowns. I previously added this functionality, but according to [this comment](https://github.com/Azure/LogicAppsUX/pull/3499#discussion_r1367742809) learned that I implemented the logic in the wrong place. This PR is to improve on that previous change and fix my previous mistake.

https://github.com/Azure/LogicAppsUX/pull/3499

Sort dropdown items alphabetically(ASC) by default.

No

image

AB#25650822

### DIFF
--- a/libs/designer-ui/src/lib/combobox/__test__/combobox.spec.tsx
+++ b/libs/designer-ui/src/lib/combobox/__test__/combobox.spec.tsx
@@ -65,7 +65,7 @@ describe('lib/combobox', () => {
     });
   });
 
-  it('ensures options are sorted alphabetically and special option is at the end', () => {
+  it('ensures options conform to the sort order of items passed in', () => {
     const { getByRole, getAllByRole } = render(<Combobox {...defaultProps} />);
     const combobox = getByRole('combobox');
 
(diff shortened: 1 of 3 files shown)


Ryanb58:
> Can we move this logic into libs\designer\src\lib\core\utils\parameters\helper.ts because, on the designer-ui side, we want to render the UI based on whatever props are passed in, and avoid manipulating the parameters as much as possible. This would require the sort to be done in 2 places:
> 
> 1. getParameterEditorProps (for initializing combobox) - when we set the options in editorOptions (line 398)
> 2. loadDynamicValuesForParameter (for initializing dynamic data in combobox) - before passing the editorOptions in updateNodeParaemters (2054)

[Comment Source](https://github.com/Azure/LogicAppsUX/pull/3499#discussion_r1367742809)

@Eric-B-Wu the change for `loadDynamicValuesForParameter` works (manually tested).

I was unable to get my mocks/spys working appropriately, so further investigation into adding unit tests around this is needed.

I also looked at the `getParameterEditorProps` line you mentioned but couldn't make sense of the adding the change there. It seems that line only gets hit when `editor` is `undefined`. I am having a challenging time wrapping my head around this change. Possibly stemming from a lack of domain knowledge on this method. Will circle back around to this when my brain is less foggy. Any direction or clarity around this would be appreciated.

Eric-B-Wu:
something we can do instead is add a prop on the designer-ui level that determines whether or not to sort, this way whoever uses the combobox ui component can choose not to sort if they wish (by default we can have it set it to be sorted)

PR #3589, opened 2023-11-02
feat(copilot): Adding support for redirect to Azure Copilot

Eric-B-Wu:
azurecopilot


libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/scratchTab.tsx
```diff
@@ -106,9 +109,30 @@ export const ScratchTab = () => {
     setNaturalLanguageQueryInput(example);
   }, []);
 
+  const azureCallback = (prompt?: string) => {
+    alert(prompt);
```
ccastrotrejo, line 113: is the alert expected?

ccastrotrejo (approved):

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/scratchTab.tsx
Eric-B-Wu: This is just the scratch tab, the callback is normally passed back by the azure portal, but because we don't have the function in standalone, I just used an alert temporarily

PR #3590, opened 2023-11-02
fix(Copilot): Changing from Copilot to Workflow Assistant

valentina-vallalta:
Changing the text within the chatbox to say Workflow Assistant instead of Copilot.

Before:

 After:


libs/chatbot/src/lib/ui/styles.less
```diff
@@ -34,14 +34,13 @@
...
   .msla-chatbot-header-title {
     font-weight: 600;
-    font-size: 24px;
+    font-size: 16px;
     line-height: 28px;
     margin: 0px 12px;
   }
 
   .msla-chatbot-header-subtitle {
     margin-left: 15px;
```
Eric-B-Wu, line 43: very nit, but could you decrease the the margin-left of the subtitle so that it aligns  a bit better with the title

Eric-B-Wu (approved):

libs/chatbot/src/lib/ui/styles.less
valentina-vallalta: done!

PR #3628, opened 2023-11-07
feat(Copilot): Adding copy button

valentina-vallalta:
Adding a new button next to the thumbs up/down that allows the user to copy the text in the chat bubble. There is currently no visual cue to let the user know that the text has been successfully copied so that might be something we want to add.

Before:
image

After:
copy-button

On Hover:
copy-button-hover

Dark Mode:
image


hartra344 (approved):

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
```diff
@@ -71,6 +79,15 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
             <div className={'msla-bubble-footer-disclaimer'}>{intlText.aIGeneratedDisclaimer}</div>
             {onThumbsReactionClicked && (
               <div className={'msla-bubble-reactions'}>
+                <IconButton
+                  className={'msla-copy-button'}
+                  title={'Copy'}
+                  iconProps={{ iconName: 'Copy' }}
+                  onClick={() => {
+                    const text = reactElementToText(children);
```
Eric-B-Wu, line 87: instead of extracting the string from the Markdown children props, let's just pass down the text down to the chatBubble. This way we can also conditionally render this copy button whether or not there is a text there - currently this would happen in all cases of the chat bubble

valentina-vallalta: I made those changes, when you get a chance please review the latest code!

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
```diff
@@ -71,6 +73,16 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
             <div className={'msla-bubble-footer-disclaimer'}>{intlText.aIGeneratedDisclaimer}</div>
             {onThumbsReactionClicked && (
               <div className={'msla-bubble-reactions'}>
+                {text && (
+                  <IconButton
+                    className={'msla-copy-button'}
+                    title={'Copy'}
```
Eric-B-Wu, line 79: let's localize this text

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
```diff
@@ -71,6 +73,16 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
             <div className={'msla-bubble-footer-disclaimer'}>{intlText.aIGeneratedDisclaimer}</div>
             {onThumbsReactionClicked && (
               <div className={'msla-bubble-reactions'}>
+                {text && (
```
Eric-B-Wu, line 76: we should also move this above the onThumbsReactionClicked component on line 74 that way it doesn't get hidden if we don't pass in the callback to this component

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
valentina-vallalta: done!

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
valentina-vallalta: done!

libs/designer-ui/src/lib/chatbot/components/chatBubble.tsx
valentina-vallalta: Made the changes! please take a look at the latest when you get a chance and let me know if there is anything else!

Eric-B-Wu (approved):

PR #3648, opened 2023-11-08
fix(copilot): Adding ellipses to placeholder text

valentina-vallalta:
Adding ellipses at the end of the placeholder text in the chatbot to match Azure Copilot

Before:
image

After:
image


Eric-B-Wu (approved):
nit: should we include a space before ellipses? Might need to confirm with Esther

PR #3701, opened 2023-11-15
feat(Data Mapper): Organize Reducer to Keep Project Name

Elaina-Lee:

  - removed WorkflowState's Uninitialized State since we are moving initialized variable out of this Slice
  - set WorkflowState's initialState referencing DesignerSlice initialState
  - moved initialized / project to a projectState (since we are sharing this between all webviews)
  - checking projectName for event listener actions
  - added message types for designer and data mapper using extensionCommand
  - passing project: 'designer' and 'dataMapper' to initialize_frame calls from vs-code projects


apps/vs-code-react/src/webviewCommunication.tsx
```diff
@@ -1,46 +1,153 @@
...
-      case ExtensionCommand.update_export_path:
-        dispatch(updateTargetDirectory(message.data));
-        break;
-      case ExtensionCommand.add_status:
-        dispatch(addStatus(message.data));
+    if (message.command === ExtensionCommand.initialize_frame) {
+      dispatch(initialize(message.data.project));
+    }
+
+    switch (projectState?.project) {
```
Elaina-Lee, line 75: Something to consider when testing: will dispatch(initialize(message.data.project)) call update this value right away for this switch case to work?

apps/vs-code-react/src/state/store.ts
```diff
@@ -2,10 +2,12 @@
 import { dataMapSlice } from './DataMapSlice';
 import { designerSlice } from './DesignerSlice';
 import { workflowSlice } from './WorkflowSlice';
+import { projectSlice } from './projectSlice';
 import { configureStore } from '@reduxjs/toolkit';
 
 export const store = configureStore({
   reducer: {
+    project: projectSlice.reducer,
     workflow: workflowSlice.reducer,
```
Elaina-Lee, line 11: @ccastrotrejo Would this 'workflow' make sense for the reducer name for containing overview, review, export?

libs/data-mapper/src/lib/models/WebviewMessages.ts
```diff
@@ -32,7 +34,7 @@ export type MessageToVsix =
       data: string;
     }
   | {
-      command: 'webviewLoaded' | 'readLocalSchemaFileOptions' | 'readLocalCustomXsltFileOptions';
+      command: 'initialize' | 'webviewLoaded' | 'readLocalSchemaFileOptions' | 'readLocalCustomXsltFileOptions';
     }
```
Elaina-Lee, line 38: @ccastrotrejo Should we use ExtensionCommands constants here when this is data-mapper not related to vs-code?

ccastrotrejo: Yeah, I would say yes. The purpose its the same the command between the vscode and react

apps/vs-code-react/src/webviewCommunication.tsx
ccastrotrejo: Yes, I assume it will do. Lets test it to make sure it does

apps/vs-code-react/src/webviewCommunication.tsx
```diff
@@ -1,46 +1,153 @@
...
+            break;
+          case ExtensionCommand.setXsltData:
+            dispatch(changeXsltFilename(message.data.filename));
+            dispatch(changeXsltContent(message.data.fileContents));
+            break;
+          case ExtensionCommand.getConfigurationSetting:
+            dispatch(changeUseExpandedFunctionCards(message.data));
+            break;
+          default:
+            console.warn(`Unexpected message received:`);
```
ccastrotrejo, line 127: Is there a reason on why it is only console logging the error and not throwing an error?

apps/vs-code-react/src/state/store.ts
ccastrotrejo: Yeah, I really don't have anything in mind to change that name to.

apps/vs-code-react/src/webviewCommunication.tsx
```diff
@@ -1,46 +1,153 @@
...
+  changeSourceSchemaFilename,
+  changeTargetSchemaFilename,
+  changeUseExpandedFunctionCards,
+  changeXsltContent,
+  changeXsltFilename,
+} from './state/DataMapSlice';
+import { initializeDesigner, updateCallbackUrl, updateFileSystemConnection, updatePanelMetadata } from './state/DesignerSlice';
+import type { InitializePayload } from './state/WorkflowSlice';
+import {
+  initialize as initializeWorkflow,
```
ccastrotrejo, line 34: Lets change it to `initializeWorkflow` and keep the initialize to the project, therefore whenever we imported it we need to import it with an alias

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
```diff
@@ -166,6 +166,7 @@ export default class OpenDesignerForLocalProject extends OpenDesignerBase {
         this.sendMsgToWebview({
           command: ExtensionCommand.initialize_frame,
           data: {
+            project: 'designer',
```
ccastrotrejo, line 169: There are const for the project name in here apps/vs-code-react/src/run-service/types.ts

apps/vs-code-react/src/webviewCommunication.tsx
Elaina-Lee: I was unsure on the reason on why the original code was console logging instead of throwing error. I'll change that to reflect throwing error for other cases

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
Elaina-Lee: I am aware of those project names existing in vs-code-react but was worried for importing that into vs-code-designer. Maybe I am missing something, but I couldn't find const export/imports from vs-code-react folder so it seemed that we cannot import that with @microsoft/vs-code-react. 

project: 'export', 'overview', 'review' also are not using those const for the project name. Would you suggest moving ProjectName const to vscode-extension folder so we can share the const in vs-code-react and vs-code-designer, or would I be missing something to just import the project name from apps/vs-code-react/src/run-service/types.ts into vs-code-designer?

ccastrotrejo: yeah if that works we can use the `vscode-extension` lib for that

ccastrotrejo (approved):

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
Elaina-Lee: Got it, I'll do that in the next PR!

PR #3702, opened 2023-11-15
fix(designer-ui): Prevent text formatting in non-HTML inputs

### LINKED ISSUE #3700: Text in non-HTML editors can be formatted (e.g., bolded)

1. Open any workflow in the designer.
2. Ensure the open workflow has some action or trigger with a text-based input that is **NOT** an HTML editor.
3. Enter something in the above text-based input.
4. Highlight the entered text.
5. Hit Ctrl+B (or Cmd+B on Mac).

Expected: Nothing should happen.
Actual: Text becomes bold.

(Also works with italics using Ctrl+I, etc.)

Standard (Portal)

New Designer

Yes

### Workflow JSON

_No response_

### Screenshots or Videos

format-non-html

### Browser

Windows 11 / Chrome 119.0.6045.159.

### Additional context

_No response_

AB#25854136

ek68794998:

Bug fix

Text can be bolded, italicized, etc. in non-HTML editors using keyboard shortcuts. (See #3700.)

Plaintext editor is used when toolbar is not shown, preventing use of rich text. Fixes #3700.

No

It's very difficult to show, but I tried my best using keystrokes in the GIF below.

format-non-html-fix

### DIFF
--- a/libs/designer-ui/src/lib/editor/base/index.tsx
+++ b/libs/designer-ui/src/lib/editor/base/index.tsx
@@ -25,6 +25,7 @@ import { LexicalComposer } from '@lexical/react/LexicalComposer';
 import { ContentEditable } from '@lexical/react/LexicalContentEditable';
 import LexicalErrorBoundary from '@lexical/react/LexicalErrorBoundary';
 import { HistoryPlugin as History } from '@lexical/react/LexicalHistoryPlugin';
+import { PlainTextPlugin } from '@lexical/react/LexicalPlainTextPlugin';
 import { RichTextPlugin } from '@lexical/react/LexicalRichTextPlugin';
 import { useEffect, useRef, useState } from 'react';
 import { createPortal } from 'react-dom';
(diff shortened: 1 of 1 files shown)


libs/designer-ui/src/lib/editor/base/index.tsx
```diff
@@ -163,6 +164,8 @@ export const BaseEditor = ({
   };
 
   const id = useId('msla-described-by-message');
+  const TextPlugin = toolbar ? RichTextPlugin : PlainTextPlugin;
```
ek68794998, line 167: Open question: Is `toolbar` here the best way to detect if we're in rich text mode?

ek68794998: Looks like this is what's used on line 135, so should be safe.

Eric-B-Wu: yeah this is fine, though I'm gonna create an item to move the toolbar prop into something a bit more inline with the HTML editor

Eric-B-Wu (approved):

PR #3748, opened 2023-11-24
feat(designer): Add credential mapping support to connector

cdgolmar-msft:
# feat(designer): Add credential mapping support to connector #3749

See discussion #3749.

This PR is adding an (optional) capability to support Power Automate credentials in the Create Connection experience.
- reading "credential mapping" annotations from connector definition (data-driven)
- allowing host service to leverage these annotations and provide a dedicated custom editor (host service-driven)

These changes are **additive** (no breaking change), **optional** (no-op if service is not implementing new method). 
They will be leveraged by Power Automate - for example to support Windows credentials for Power Automate Desktop operation - since the Credentials feature is based on Dataverse record and APIs.

Learn more with the documentation: [Use credentials in desktop flow connections (preview)](https://learn.microsoft.com/en-us/power-automate/desktop-flows/manage-credentials).

## Requirements

* [N/A] Docs have been added / updated (for bug fixes / features)
* Does this PR introduce a breaking change? No.

## Description and changes and behavior

**Credential connector metadata parsing.**
From the connector definition, the connection setup panel (`\designer-ui\src\lib\createConnection\`) - which renders each connection parameter - identifies if parameters belong to a Credential mapping.

**Custom Credential component rendering.** 
A new method in `IConnectionParameterEditorService` is called to determine if the host supports Credentials - hence should provide the react component to use

Note: This is optional, not implementing this method or not returning a component keeps the current behavior

If a component is returned, the corresponding parameters are not rendered in favor of rendering tghe provided Credential picker component.

## Screenshots

| Screenshots | |
|--|--|
| Credential picker in "credentials" mode | Screenshot 2023-11-16 165035 |
| Credential picker in "parameters" mode | Screenshot 2023-11-16 164646 |

See discussion #3749.

AB#26007671


libs/utils/src/lib/models/connector.ts
```diff
@@ -82,6 +82,31 @@ export interface ConnectionParameterUIDefinitionBase {
...
+   */
+  credentialMapping?: ParameterCredentialMappingUIDefinition;
+}
+
+export interface ParameterCredentialMappingUIDefinition {
+  /**
+   * Name (identifier) of the mapping, parameters with the same mappingName belong to the same credentials item.
+   */
+  mappingName: string;
+  displayName?: string | null | undefined;
```
Eric-B-Wu, line 96: nit: we don't need | undefined if we have ?

Eric-B-Wu (approved):

libs/designer-ui/src/lib/createConnection/index.tsx
```diff
@@ -583,29 +615,47 @@ export const CreateConnection = (props: CreateConnectionProps): JSX.Element => {
...
+                    };
+                    return <CredentialsMappingEditorComponent key={mappingName} {...props} />;
+                  }
+                }
+
+                // If we encounter an already active mapping,
+                // we skip the parameter rendering since an Editor was already found and rendered.
+                if (activeParameterMappings.has(mappingName)) {
+                  return null;
+                }
```
rllyy97, line 655: Can we switch the logic for the default parameters vs mapping parameters here around?  It looks like you moved our default logic apps code out to a different spot, I'd rather you do that for the mapping code.
We can leave our existing parameter code in place and just have it run `if (mappingName) return renderMappingParameter(etc)` before and return the special case.  No need to move what we have out to a new spot if it's still the "default" behavior.

rllyy97 (approved):

PR #3765, opened 2023-11-28
fix(Designer): Run after settings dots are not ordered properly

Elaina-Lee:

bug fix
The runAfter dots do not align with the order of the nodes
The runAfter dots align with the order of the nodes

Before:
278495837-df04965e-d9a4-47f1-a987-31179905ade4
After:
image

AB#26001478

### DIFF
--- a/libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
+++ b/libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
@@ -123,11 +123,9 @@ export const getWorkflowNodeIndexFromGraphLevel = (
   runAfters: string[],
   actionId: string
 ) => {
-  const sortedRunAfters = runAfters.slice(0).sort(function compareFn(id1, id2) {
-    const workflowNode1PositionX = workflowGraphNodes?.[id1]?.position?.x ?? 0;
-    const workflowNode2PositionX = workflowGraphNodes?.[id2]?.position?.x ?? 0;
-    return workflowNode2PositionX - workflowNode1PositionX;
-  });
+  const sortedRunAfters = runAfters
+    .slice(0)
+    .sort((id1, id2) => (workflowGraphNodes?.[id2]?.position?.x ?? 0) - (workflowGraphNodes?.[id1]?.position?.x ?? 0));
 
   return sortedRunAfters?.findIndex((key) => key === actionId);
 };
@@ -149,7 +147,7 @@ export const useWorkflowNode = (actionId?: string) => {
   });
 };
 
-export const useWorkflowNodeIndex = (actionId: string, runAfters: string[]): number => {
+export const useRunAfterIndexBySource = (actionId: string, runAfters: string[]): number => {
   return useSelector((state: RootState) => {
     if (!actionId || !state?.workflow?.graph?.children) {
       return 0;
--- a/libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
+++ b/libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
@@ -118,6 +118,22 @@ export const getWorkflowNodeFromGraphState = (state: WorkflowState, actionId: st
   return traverseGraph(graph);
 };
 
+export const getWorkflowNodeIndexFromGraphLevel = (
+  workflowGraphNodes: {
+    [hi: string]: WorkflowNode;
+  },
+  runAfters: string[],
+  actionId: string
+) => {
+  const sortedRunAfters = runAfters.slice(0).sort(function compareFn(id1, id2) {
+    const workflowNode1 = workflowGraphNodes?.[id1]?.position?.x ?? 0;
+    const workflowNode2 = workflowGraphNodes?.[id2]?.position?.x ?? 0;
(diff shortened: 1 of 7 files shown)


ccastrotrejo (approved):

libs/designer/src/lib/core/state/workflow/workflowSlice.ts
```diff
@@ -248,6 +249,27 @@ export const workflowSlice = createSlice({
...
+
+      while (stack.length) {
+        const node = stack.shift();
+        const change = positionChangesById[node?.id ?? ''];
+        const nodePosition = change?.position;
+        if (node && nodePosition) {
+          node.position = nodePosition;
+        }
+        !!node?.children?.length && stack.push(...node.children);
+      }
```
rllyy97, line 272: With your current setup you're not just passing position changes, you're passing every node in the workflow.  This might become pretty impactful on performance on higher node count workflows if we update _every_ node's data on each change

libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
```diff
@@ -135,6 +149,21 @@ export const useWorkflowNode = (actionId?: string) => {
   });
 };
 
+export const useWorkflowNodeIndex = (actionId: string, runAfters: string[]): number => {
```
rllyy97, line 152: This should be more descriptive, explain what you mean by index, index in relation to what?

libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
```diff
@@ -118,6 +118,20 @@ export const getWorkflowNodeFromGraphState = (state: WorkflowState, actionId: st
   return traverseGraph(graph);
 };
 
+export const getWorkflowNodeIndexFromGraphLevel = (
+  workflowGraphNodes: Record<string, WorkflowNode>,
+  runAfters: string[],
+  actionId: string
+) => {
+  const sortedRunAfters = runAfters.slice(0).sort(function compareFn(id1, id2) {
```
rllyy97, line 126: Nit: You can make this an arrow function
```suggestion
  const sortedRunAfters = runAfters.slice(0).sort((id1, id2) => {
```

libs/designer/src/lib/core/state/workflow/workflowSelectors.ts
rllyy97: `useRunAfterIndexBySource` or something along those lines

libs/designer/src/lib/ui/connections/edge.tsx
```diff
@@ -82,7 +87,7 @@ export const ButtonEdge: React.FC<EdgeProps<LogicAppsEdgeProps>> = ({
     [operationData?.runAfter]
   );
   const numRunAfters = Object.keys(filteredRunAfters).length;
-  const raIndex = useMemo(() => Object.entries(filteredRunAfters).findIndex(([key]) => key === source), [filteredRunAfters, source]);
+  const raIndex: number = useRunAfterIndexBySource(source, Object.keys(filteredRunAfters));
```
Elaina-Lee, line 90: should this be a useMemo too?

rllyy97: No this should be fine

Elaina-Lee:
Discussed with Riley: this is having an infinite loop. Going to create a separate data structure within workflowState

libs/designer/src/lib/core/state/workflow/workflowSlice.ts
```diff
@@ -249,6 +252,24 @@ export const workflowSlice = createSlice({
         !!node?.children?.length && stack.push(...node.children);
       }
     },
+    updateNodeInfo: (state: WorkflowState, action: PayloadAction<{ nodes: Node[]; elkNodes?: ElkNode[] }>) => {
```
Elaina-Lee, line 255: @rllyy97 Passing both nodes and elkNodes since nodes here do not contain dimensions data. Since we are not using height/width, I wasn't sure how necessary this is for now. Alternatively, we can update this in the updateNodeSizes call, but that would be another huge loop so wasn't sure if we want that.

Elaina-Lee: Seems like height/width of the WorkflowNode is being used in `convertWorkflowGraphToElkGraph` in `elkLayout.ts`, if we would change the height/width being used here as ReactFlowNodeInfo, we can even get rid of the `updateNodeSizes ` - not sure if this is the direction we want to go to.

rllyy97: I'm a little confused on what you're asking / proposing.  From our conversation yesterday let's leave the size logic the same as it was, and put the position logic into a new place in state (which it looks like you've done in your last commit).  Is what you have working at the moment?

Elaina-Lee: I am passing in an additional parameter (elkNodes) just to save the height/width into the new structure that is not currently being used anywhere. I was wondering if this still aligns with what we discussed

Issue #2681, opened 2023-06-06
Create a custom clipbard plugin for Lexical so that we can do proper interpolation patterns with tokens


hartra344:
AB#24191777

PR #3793, opened 2023-12-04
feat(designer): Supporting Pasting Tokens in Editors

Eric-B-Wu:
Adding Support for Pasting Tokens in all editors
pasteTokens

Fixes https://github.com/Azure/LogicAppsUX/issues/2681

Most of this PR is moving the new props down the editor ancestry chain
1. tokenMapping?: Record<string, ValueSegment>;
2. loadParameterValueFromString?: (value: string) => ValueSegment[];
Hopefully when we move designer-ui into designer project, we won't need to pass it down and instead have it all in a store.

Some caveats:
Currently if we a copy a token to a place where it would be invalid, it will serialize as a faulty token and show up as our default invalid token. 
HTML editor can copy tokens from html editor to html editor, but unable to copy from stringified tokens into html editor

Other Changes
[Potentially Breaking for PA]: Moving our Base Editor inside a lexical wrapper:
-     Previously we couldn't reference the the lexical editor/editorstate  directly within BaseEditor.tsx, so adding an Editor Wrapper around that to initialize lexical

Removed Delete button from tokens when in readonly mode

BaseEditorPlugin toolbar prop renamed to isHtmlEditor

Fixed a bug where selecting a set of tokens (tokens become highlighted), and then blurring the editor that the selected tokens remain selected

Fixed an issue when hovering over an expression token in an editor (non-string editor) would not show tooltip of expression value


libs/designer-ui/src/lib/querybuilder/Group.tsx
```diff
@@ -204,19 +207,19 @@ export const Group = ({
...
+  //   } else if (childIndex >= groupProps.items.length - 1 && moveOption === MoveOption.DOWN) {
+  //     // come back
+  //   } else {
+  //     const newItems = { ...groupProps };
+  //     const child = newItems.items[childIndex];
+  //     newItems.items[childIndex] = newItems.items[childIndex + (moveOption === MoveOption.UP ? -1 : 1)];
+  //     newItems.items[childIndex + (moveOption === MoveOption.UP ? -1 : 1)] = child;
+  //     handleUpdateParent(newItems, index);
+  //   }
+  // };
```
rllyy97, line 222: Are these and the other commented code in this file not needed anymore?

rllyy97 (approved):

libs/designer-ui/src/lib/querybuilder/Group.tsx
Eric-B-Wu: This logic was in part trying to emulate the original Query builder from an external library, I forgot exactly what issue I ran into, but because at that point we were rushing to public preview I was planning on revisiting it in the future - commented out because this code is still unused, but I do hope to come back to it (I'll create an issue to track it)

Eric-B-Wu: btw this was a purely cosmetic miss, which was why it was not prioritized https://github.com/Azure/LogicAppsUX/issues/3796 - here is issue for tracking

PR #3802, opened 2023-12-05
fix(Designer): Add Bounding Box to Folder Level Button in Panel

### LINKED ISSUE #3797: Luminosity ratio of selected item with its background is less than 3:1
### Describe the issue

1. Open Designer standalone (nx serve).
2. Add an action that uses folder picker editor.
3. Tab navigate to folders in the folder picker pop-out.
4. Observe that there is only a faint highlight around currently selected item.

**Expected:**
There should be a high contrast highlight or there should be a bounding box around the currently selected item.

**Actual:**
Currently selected item is only faintly highlighted.

New Designer (With action panels)

### Screenshots or Videos

Not sufficient highlight around currently selected item:
image

Bounding box around button to navigate to next folder level:
image

### Browser

Chrome

### Additional context

Related ADO bug: https://msazure.visualstudio.com/One/_workitems/edit/14566641

Elaina-Lee:

This fixes #3797 

image
- doesn't have the bounding box (border) on the button on focus (on tab)
- uses border when focus-visible (outline would not show so used border here)
- Not using 3:1 ratio as suggested since that'd be too dark for text

no
image


rllyy97 (approved):

Eric-B-Wu:
the offcenter folder is only appearing in PA, right?

Elaina-Lee:
> the offcenter folder is only appearing in PA, right?

Yeah I wasn't able to reproduce the offcenter folder, and the screenshot with the offcenter folder is from the ticket (PA).

PR #3815, opened 2023-12-08
fix(Designer): Connections panel refactor

rllyy97:
> I will be doing a self-review of this PR before merging

## Main Changes

- Moved connections code out of the previous location in the NodeDetails Panel
  - No more managing active panel tabs when performing any sort of panel navigation
  - Connection flow now is completely separate from the previous panel (code-wise)
    - Still accessible in the same ways as before
- Changes to the existing connections code 
  - Switched the `select connections` component out for a similar fluent/react-components component, fixed a lot of unnecessary weirdness we had to deal with previously
  - 
- New Connections panel (Hidden behind designer option)
  - Accessed from command bar
  - Shows all current connections
    - Shows state of connection (valid / invalid) 
    - Option to navigate directly to connection resource (top right of connection card)
    - Shows linked actions, clicking on the node buttons takes you straight to editing the connection of the node
    - Also shows a `Reassign all` button at the end of the list, this will take you to the select/create connection flow, and will apply your changes to all nodes that currently have the same connection
      - This is for if you have a lot of nodes all hooked up to a connection and need to switch them all for whatever reason, or if you have a lot of disconnected nodes you can connect them all at once
- Connection flow and components have fully been moved out of `Designer-UI` into `Designer`
- Multiple node selection is now possible from a Redux standpoint
  - I put this in for the multiple-node connection editing, but utilizing this to allow for bulk-deletion or any other bulk actions should now be possible
- Various redux changes/cleanup related to panel state / connection state

## Breaking changes

All of our `switchTo{panel_name}()` redux actions were combined into a new action called `openPanel()` which takes a few parameters depending on what you're doing.  It will always take a `panelMode` to just determine which panel needs to be opened, but also optional `nodeId`, `nodeIds`, and `referencePanelMode` (referencePanelMode we are using currently just as a sort of `this is where we were before the new panel mode was opened`, this is only used in the connections flow right now to determine if we should go to the parameters or all connections view after editing a connection).  
Current panel modes are `'Operation' | 'Discovery' | 'WorkflowParameters' | 'NodeSearch' | 'Error' | 'Connection'`

Our connection components were moved out of `designer-ui`, I don't think anyone was pulling out our connections components but if you were, it is the same name, just from the `designer` package instead

## Notes

There is a jitter when going from to any other panel to the default NodeDetailsPanel.  This appears to be because we render the NodeDetailsPanel as an actual separate panel component instead of modifying a singular panel's content like we do for every other panel.  I'm working on fixing this and moving the NodeDetailsPanel over to use the global panel, but there is a lot of logic specific to the NodeDetailsPanel we need to account for.

Also the package lock was updated thats part of why the PR is so big

---

image

image


hartra344:
Just make sure there's lots of pictures lol

.vscode/settings.json
```diff
@@ -1,5 +1,9 @@
 {
   "typescript.tsdk": "node_modules\\typescript\\lib",
   "cSpell.words": ["fluentui", "reactflow"],
-  "eslint.validate": ["json"]
+  "eslint.validate": ["json"],
+  "terminal.integrated.env.windows": {
```
Eric-B-Wu, line 5: nit: remove, we should probably add this file to gitignore, I noticed this was being added to my PRs as well

libs/designer/src/lib/ui/panel/connectionsPanel/allConnections/connectionEntry.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+  const connectedActionsText = intl.formatMessage({
+    defaultMessage: 'Connected actions',
+    description: 'Header for the connected actions section',
+  });
+  const reassignConnectionTooltipText = intl.formatMessage({
+    defaultMessage: 'Reassign all connected actions to a new connection',
+    description: 'Tooltip for the button to reassign actions',
+  });
+  // const connectionValidStatusText = intl.formatMessage({
```
Eric-B-Wu, line 44: nit: do we still need this text?

Eric-B-Wu (approved):

.vscode/settings.json
rllyy97: Yeah I agree, I'm just going to add to the .gitignore

libs/designer/src/lib/ui/panel/connectionsPanel/allConnections/connectionEntry.tsx
rllyy97: I was going to add a tooltip for the connection status with these, it was causing an error for some reason but I'll get it sorted

PR #3824, opened 2023-12-11
Fix(Data Mapper): fixing edifact nested loop mapping bug 

DanielleCogs:

bug fix
- see attached bug
allow backout of current loop using '../'
no


libs/data-mapper/src/lib/utils/DataMap.Utils.ts
```diff
@@ -406,6 +426,7 @@ export const Separators = {
   CloseParenthesis: ')',
   Comma: ',',
   Dollar: '$',
+  //ForwardSlash: '/'
```
Elaina-Lee, line 429: Would this be for future reference?

libs/data-mapper/src/lib/mapDefinitions/MapDefinitionSerializer.ts
```diff
@@ -109,103 +113,151 @@ export const generateMapDefinitionBody = (mapDefinition: MapDefinitionEntry, con
...
+            if (
+              formattedLmlSnippetForSource === inputNode.node.key &&
+              inputNode.node.nodeProperties.includes(SchemaNodeProperty.Repeating)
+            ) {
+              formattedLmlSnippetForSource = '.';
             } else if (valueToTrim) {
-              value = value.replace(`${valueToTrim}/`, '');
+              // account for source elements at different level of loop
+              let backoutValue = '';
+              if (valueToTrim !== lastLoop.loop && !valueToTrim.includes('/*')) {
```
Elaina-Lee, line 234: In which case would a part of data map value include '/*'?

Elaina-Lee (approved):

libs/data-mapper/src/lib/utils/DataMap.Utils.ts
DanielleCogs: yeah I think we will need it soon!

libs/data-mapper/src/lib/mapDefinitions/MapDefinitionSerializer.ts
DanielleCogs: this is for json arrays! And there is a bug in the JSON array serialization, so this feature won't work for json for now

PR #3837, opened 2023-12-13
fix(Data Mapper): Disabled allowing undo add schema

### LINKED ISSUE #3507: Nit: Able to "Undo" add schema

1. open new instance and add a new schema
2. Select "undo" and the schema disappears
seems like something that we should not allow

### Screenshots or Videos

_No response_

### Version

2.75.0

### Additional context

_No response_

AB#25572560

Elaina-Lee:

fix #3507 
- can undo adding initial schemas
- undo is disabled until adding both of the schemas and making an action

### DIFF
--- a/libs/data-mapper/src/lib/core/state/Store.ts
+++ b/libs/data-mapper/src/lib/core/state/Store.ts
@@ -8,7 +8,6 @@ import { configureStore } from '@reduxjs/toolkit';
 import undoable, { includeAction } from 'redux-undo';
 
 const includedActions = [
-  'dataMap/setInitialSchema',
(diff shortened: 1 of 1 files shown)


Elaina-Lee:
This disables "undo" until both initial schemas are selected & an action is done.

When a new schema is selected in the middle of actions, the user can still "undo" the change as it is not considered as a schema change but rather a state change in multiple parts (nodes, connections, etc). If we want to disable "undo" after a schema is changed in the middle, we may want to consider emptying the entire undo-redo stack on those.

ccastrotrejo (approved):

PR #3849, opened 2023-12-14
feat(designer-ui): Support raw code view in HTML editor

ek68794998:

Feature: Adds a button to the HTML editor that allows switching to view and edit the raw HTML format. Resolves #3456.

Only rich editor is supported.

Both rich editor and raw editor are supported.

No, not in terms of consumption interfaces, etc. This does introduce a number of changes to the HTML editor, though.

html-editor


libs/designer-ui/src/lib/html/plugins/toolbar/helper/HTMLChangePlugin.tsx
```diff
@@ -1,35 +1,77 @@
...
 }
 
-export const HTMLChangePlugin = ({ setValue }: HTMLChangePluginProps) => {
+export const HTMLChangePlugin = ({
+  isValuePlaintext,
+  loadParameterValueFromString,
+  setIsValuePlaintext,
+  setValue,
+}: HTMLChangePluginProps) => {
   const onChange = (editorState: EditorState, editor: LexicalEditor) => {
```
ek68794998, line 35: I've found one bug where if you add a token outside of an element in the raw editor, then switch, it does something janky:

html-editor-bug-1

libs/designer-ui/src/lib/html/plugins/toolbar/Toolbar.tsx
```diff
@@ -199,13 +225,62 @@ export const Toolbar = ({ readonly = false }: toolbarProps): JSX.Element => {
...
       <Divider />
-      <Format activeEditor={activeEditor} readonly={readonly} />
+      <Format activeEditor={activeEditor} readonly={formattingButtonsDisabled} />
       <ListPlugin />
       <LinkPlugin />
+      {setIsRawText ? (
+        <button
+          aria-label="Raw code toggle"
+          className={css('toolbar-item', isRawText && 'active')}
+          disabled={readonly}
```
ek68794998, line 244: There is not yet support for disabling this button if Lexical would fail to parse the raw HTML provided by the user.

libs/designer-ui/src/lib/html/plugins/toolbar/helper/HTMLChangePlugin.tsx
Eric-B-Wu: this might be something we want to look at in a future PR, but if im gonna guess its when converting it from lexical editor to dom, its inferring the html elements to get that

Eric-B-Wu (approved):

libs/designer-ui/src/lib/html/plugins/toolbar/helper/HTMLChangePlugin.tsx
ek68794998: Yeah, I think everything at the top level needs to be a known tag type, but will leave this as an open bug to address.

PR #3872, opened 2023-12-18
fix(Designer): Converted many fluent8 components to fluent9

rllyy97:
## Main Changes

Updated a lot of old fluent components to use v9 over v8 (using `@fluentui/react-components`
- Updated buttons, tooltips, menus, icons across project (icon changes were discussed with design team)
  - Left input fields as-is, we have a lot of custom logic with those
- Moved menu items into their own components so they can be reused as needed
- Changed a few things for better visibility of surfaces, especially with overlapping components
- File size diff is mostly from snapshot changes, v9 is a little slimmer

This is not all-encompasing, I just tried to hit most of the commonly accessed areas without getting into _each_ of the form inputs.

---
image
image
image
image
image


rllyy97 (commented):
Self-reviewed

hartra344 (approved):

ccastrotrejo (approved):

PR #3873, opened 2023-12-18
feat(designer): add support prevent casting in complex array editors

Eric-B-Wu:
Added a prop in the array editor to prevent casting in the complex array editor

### DIFF
--- a/apps/designer-standalone/src/components/LocalizationSettings.tsx
+++ b/apps/designer-standalone/src/components/LocalizationSettings.tsx
@@ -38,6 +38,10 @@ export const LocalizationSettings = () => {
           text: 'French',
         },
         {
+          key: 'id',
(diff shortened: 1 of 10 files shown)


hartra344 (changes requested):
Let’s discuss this one offline. This is a slippery slope that worries me.

ek68794998:
My understanding of the issue here is that LA wants to serialize w/ casting even if `suppressCasting` is true for that operation, while PA wants to suppress the casting when serializing.

Maybe the more flexible solution here is a flag like `suppressCastingForSerialize`? In this case, LA would be able to leave a default `false` and PA could set this to `true`, which would mean that the `suppressCasting` value for the operation would apply when serialization occurs.

hartra344 (approved):

PR #3881, opened 2023-12-19
fix(designer): Fix regressions moving token parsing from  $[ ]$ to @{}

Eric-B-Wu:
Instead of generating a uniqueID, we've moved to use the token.value which is unique, which was done in this PR:
https://github.com/Azure/LogicAppsUX/pull/3849

This should fix some of the regressions we've had

Also, instead of using loadParameterValueFromString from designer and supplying the tokenMapping in html editor, I've gone ahead and used the desinger-ui's stringToSegments function given that we have the nodeMap when transitioning between raw and rich html


rllyy97 (approved):

Eric-B-Wu:
before: 
before


after: 
after

Issue #3890, opened 2023-12-20
Prevent MSI Authentication error message users enter a token


Eric-B-Wu:
### Describe the Bug with repro steps

1. Create an HTTP action
2. Add authentication parameter
3. Switch to Mananged Identity Auth Type
4. Use a token, ex. @appSettings(...)
5. See error message even though it is valid

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

edge

PR #3957, opened 2024-01-09
fix(designer): Some cleanup of auth serialization 

Eric-B-Wu:
Fixes: https://github.com/Azure/LogicAppsUX/issues/3890

Most of this was refactoring how the collapsed auth editor serialization. 
updated changes

There are a couple of smaller bugfixes in this
ex. 
1. It would show the wrong error message when there were missing parameters in the collapsed editor
2. tokens were being serialized in the collapsed view despite being an invalid authentication


libs/designer-ui/src/lib/authentication/plugins/CollapsedAuthenticationSerialization.tsx
```diff
@@ -0,0 +1,86 @@
...
+  setToggleEnabled,
+  setCurrentProps,
+  setOption,
+  serializeValue,
+}: CollapsedAuthenticationValidationProps): JSX.Element => {
+  const intl = useIntl();
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const showError = (errorMessage: string) => {
+    setToggleEnabled(false);
```
hartra344, line 35: instead of juggling multiple stages for this, couldn't you just infer toggleEnabled from the errorMessage being set?

rllyy97 (approved):

Issue #3965, opened 2024-01-10
Entering custom values is not allowed for multi select enums


rarayudu:
### Describe the Bug with repro steps

When we have a multi select enum, designer is not allowing to enter a custom value – This is for the cases where customer wants to take this as an output of previous action.
 
You can repro this with Xml Transform action, “Transform Options” parameter.  
 
Manifest of the operation for this parameter.
 
{
    "transformOptions", new SwaggerSchema
    {
        Title = DisplayMessages.TransformXML_TransformOptions_Title,
        Description = DisplayMessages.TransformXML_TransformOptions_Description,
        Type = SwaggerSchemaType.String,
        VisibilityExtension = SwaggerCustomVisibility.Advanced,
        Extensions = new OrdinalDictionary<JToken>
        {
            { "x-ms-editor", "dropdown" },
            {
                "x-ms-editor-options", new JObject
                {
                    { "multiSelect", true },
                    { "titleSeparator", "," },
                    {
                        "options", new JArray
                        {
                            new JObject
                            {
                                { "value", "DisableByteOrderMark" },
                                { "displayName", "Disable byte order mark" },
                            },
                            new JObject
                            {
                                { "value", "GenerateTextOutput" },
                                { "displayName", "Generate text output" },
                            },
                            new JObject
                            {
                                { "value", "ApplyXsltOutputAttributes" },
                                { "displayName", "Apply XSLT output attributes" },
                            },
                        }
                    },
                }
            },
        },
    }
},

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Chrome

PR #4017, opened 2024-01-19
feat(Consumption): Changing XML Transform Options to have Custom Option

valentina-vallalta:
Changing the XML transform options parameter to have the option of a custom option. This is done by changing it to a combo box from a drop down and adding multi select functionality to the implementation of combo box. The implementation of multi select was largely taken from how it is implemented in the dropdpwn currently. A new option of multiSelect, which is defaulted to false, must be passed down from the manifest and when not set to true, the old behavior of Combobox remains with only one item selected at a time. 

If the custom option is selected, the rest of the selected options are erased and user must use only the custom option. 

Addresses #3965

### DIFF
--- a/libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
+++ b/libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
@@ -154,6 +154,7 @@ export const TokenField = ({
           getTokenPicker={getTokenPicker}
           onChange={onValueChange}
           onMenuOpen={onComboboxMenuOpen}
+          multiSelect={editorOptions?.multiSelect}
           dataAutomationId={`msla-setting-token-editor-combobox-${labelForAutomationId}`}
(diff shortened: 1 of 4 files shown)


hartra344 (approved):

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -154,6 +154,7 @@ export const TokenField = ({
           getTokenPicker={getTokenPicker}
           onChange={onValueChange}
           onMenuOpen={onComboboxMenuOpen}
+          multiSelect={editorOptions?.multiSelect}
```
rllyy97, line 157: You also probably need to add `serialization={editorOptions?.serialization}` here, the dropdown component has it and it looks like you added it inside the combobox but you aren't passing it in

rllyy97 (approved):

PR #4028, opened 2024-01-22
fix(Designer): Multi select drop down is not showing options in new designer

### LINKED ISSUE #3931: Multi select drop down is not showing options in new designer

we noticed multiselect dropdown in action's parameter don't show values in new designer. I can repro the issue in Azure portal. 

we have this "Severity" parameter in our custom connector action, which allows multi select. This is working fine in the old designer.
image

connector json is defined as below given for that "Severity" parameter. 
image

but when the same logic app is opened in the new designer, drop down options are not shown in Azure portal. 
image

I see this error for opening the same logic app with new designer in my local box,
image

could you please check this issue with priority. This is a P0 scenario for usage of our custom connector. 

Consumption (Portal)

New Designer

Yes

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

Edge

### Additional context

_No response_

Elaina-Lee:

fixes #3931
dropdown options select disappears after coming back to the screen in some scenarios
dropdown options select remains

https://github.com/Azure/LogicAppsUX/assets/144840522/25184672-ad73-4d02-8867-a22e41569608


libs/designer-ui/src/lib/dropdown/index.tsx
```diff
@@ -130,7 +131,7 @@ const getSelectedKey = (options: DropdownItem[], initialValue?: ValueSegment[]):
   if (initialValue?.length === 1 && initialValue[0].type === ValueSegmentType.LITERAL) {
     return (
       options.find((option) => {
-        return option.value === initialValue[0].value;
+        return option.value.toString() === initialValue[0].value.toString();
```
takyyon, line 134: case insensitive check?

Elaina-Lee: Shouldn't the case be the same if we are comparing the same value that was serialized then deserialized?

Eric-B-Wu: initialValue[0] should always be a string so we shouldn't need to stringify it; however I think what we can do is instead make option.key a stringified value when initializing the editorViewModel that way we ever only compare a string to a string.

Eric-B-Wu (approved):

Xiaoyu-Huang:
@Elaina-Lee I have already pulled the version 2.106 which has this fix in. However, the issue still persists.
image

PR #4029, opened 2024-01-23
feat(copilot): Add Support for Additional Params URL

Elaina-Lee:

displays URL passed through additional parameters in the same manner azure copilot displays it
 
RP has not been changed yet, so the link is with mock data
image


Elaina-Lee:
Discussion:

1. URL description: azure copilot has description for the URL. Should we pass this from backend and use it instead of displaying URL itself?
2. Number of URLs: is there a reason of restricting to one URL?
3. Null/undefined issue: Currently local testing suggests null values for additional parameters. Would it always be null and not undefined?

hartra344 (approved):

Eric-B-Wu:
> Discussion:
> 
> 1. URL description: azure copilot has description for the URL. Should we pass this from backend and use it instead of displaying URL itself?
> 2. Number of URLs: is there a reason of restricting to one URL?
> 3. Null/undefined issue: Currently local testing suggests null values for additional parameters. Would it always be null and not undefined?

1. How our vectors are stored, we don't have a description linked with each datapoint, so this would be a challenging task to accomplish
2. The reason why we restrict to 1 url is because of our 3 sources of data only 2 of those 3 have urls linked to them. We then only use 1 of those data sources as our knowledge base in our response. We can increase the urls to be 2 depending on the relevance we receive, but beacuse we only use 1 in our response I had limited it to the single one.
3. There isn't an explicitly "undefined" value in C#, uninitialized variables are assigned default values based on their types, and reference types default to null. So in this case, because its an array if it's unset it would default to null

libs/designer-ui/src/lib/chatbot/feedbackHelper.tsx
```diff
@@ -29,6 +29,29 @@ export function useReportBugButton(disabled: boolean): IButtonProps {
...
+      defaultMessage: 'For more detailed information, you can refer to the following resources',
+      description: 'Message instructing to follow below links for more detailed information',
+    }),
+  };
+  return (
+    <div>
+      {`${intlText.guideMoreInfoText}:`}
+      <ul>
+        <li>
+          {' '}
```
Eric-B-Wu, line 45: do we need this empty string?

libs/chatbot/src/lib/ui/chatbot.tsx
```diff
@@ -236,6 +236,8 @@ export const Chatbot = ({
             __rawRequest: requestPayload,
             __rawResponse: response,
             reaction: undefined,
+            additionalDocDes: additionalParameters?.urlDes, //TODO: discuss adding this to backend
```
Eric-B-Wu, line 239: we don't have a great way of adding this information in

Eric-B-Wu (approved):

Issue #3131, opened 2023-08-14
Screen readers are not announcing the name for the icons or type of connecters are used in the flow


Ram-G:
### Describe the Bug with repro steps

1. Open designer standalone (`nx serve`)
2. Load any saved workflow
3. Click on one of the cards on the designer canvas
4. Open accessibility tool; NVDA screen reader or Microsoft Narrator should work
5. Keyboard tab between the cards on the canvas.
6. Observe what gets read out by the screen reader -- something like "clickable manual button". Notably, there is no mention of the icon of the card -- this information is lost to a blind user / keyboard user.

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/10837129/0dbd52c9-a3cf-485f-a27f-6d11c85b95f5)

### Browser

Edge

### Additional context

This bug is an Accessibility "sev2" -- which means it's necessary to fix to maintain Accessibility Grade C.

A potential fix could be to add `alt` text to the icon component: https://github.com/Azure/LogicAppsUX/blob/96f9e3aa64208e6fc9063bd3bf4c96d948d74cd2/libs/designer-ui/src/lib/card/index.tsx#L103

Link to internal ADO work item with further investigation: https://msazure.visualstudio.com/One/_workitems/edit/24398720

AB#24836815

PR #4030, opened 2024-01-23
fix(designer-ui): Action icons with no connector display name have bad ARIA text

ek68794998:

Bug fix

Connector icon ARIA text works fine:

image

But if no connector is defined in the manifest, the fallback is not good:

image

Both work correctly, using the operation name (i.e., operation display name, **not** the user-inputted name) if there is no connector name:

image

image

No

See above


libs/designer-ui/src/lib/card/index.tsx
```diff
@@ -105,30 +107,40 @@ export const Card: React.FC<CardProps> = ({
...
+          defaultMessage: '{connectorName} connector icon',
+          description: 'Alt text icon on action/trigger card when there is a connector name',
+        },
+        {
+          connectorName,
+        }
+      )
+    : intl.formatMessage(
+        {
+          defaultMessage: '{operationName} icon',
```
ek68794998, line 122: Wasn't sure if this should be `___ icon` or `___ operation icon`?

hartra344:
I actually wonder if an argument could be made that these images are decorative? And maybe shouldn't have alt text at all? Is this the only way we indicate the connector type? Maybe a better solution would be to add screen reader hints to tell the connector type and the icon doesn't matter.

W3C guidance on this: https://www.w3.org/WAI/tutorials/images/decorative/

ek68794998:
> I actually wonder if an argument could be made that these images are decorative? And maybe shouldn't have alt text at all? Is this the only way we indicate the connector type? Maybe a better solution would be to add screen reader hints to tell the connector type and the icon doesn't matter.
> 
> W3C guidance on this: https://www.w3.org/WAI/tutorials/images/decorative/

We used to do so, but bug #3131 (resolved in PR #3145) was an accessibility sev-2 on us that indicated they should be announced rather than seen as decorative. Thoughts?

hartra344:
I think that's a valid bug if it's the only way we indicate connector type. However I think a better fix is to make connector type better accessable(even for non screen reader tech hoestly, who memorizes connector by icon) than to make decorative icons screen redundantly read out in a way that doesn't matter.

If I'm seeing impaired, or fully blind, what does "Outlook connector icon" mean to me? The only valuable information I'm getting out of that is "Outlook Connector". I likely don't know what the icon looks like and why would I care that it's there? It's extra information that does nothing.

Eric-B-Wu (approved):

ek68794998:
> I think that's a valid bug if it's the only way we indicate connector type. However I think a better fix is to make connector type better accessable(even for non screen reader tech hoestly, who memorizes connector by icon) than to make decorative icons screen redundantly read out in a way that doesn't matter.
> 
> If I'm seeing impaired, or fully blind, what does "Outlook connector icon" mean to me? The only valuable information I'm getting out of that is "Outlook Connector". I likely don't know what the icon looks like and why would I care that it's there? It's extra information that does nothing.

This is a valid point. Is it perhaps better to read the card out as "SharePoint, Create item card, button" or similar, rather than trying to read the icon? I'm not sure if the accessibility team's concern here is that the icon needs to be readable, or whether the card button doesn't have enough ARIA content for screen readers.

hartra344:
> Member

I like this idea, the only change would be that I would go "Action Name, Connector" since action name is the primary information (button announcement should come from role).

ek68794998:
Sounds good, I will look at that when I have cycles. Dropping this back to draft for now.

hartra344:
And if they a11y team wants to push adding a description to the icon, I'm more than happy to go to bat on that. It goes directly against W3C guidance, likely coming from automated tooling without any thought behind if it's right or not.

libs/designer/src/lib/core/state/selectors/actionMetadataSelector.ts
```diff
@@ -107,40 +109,53 @@ const useNodeAttribute = (operationInfo: NodeOperation, propertyInManifest: stri
...
   const useManifest = operationManifestService.isSupported(operationInfo?.type ?? '', operationInfo?.kind ?? '');
 
-  const { data: connectorData } = useConnectorAndSwagger(operationInfo?.connectorId, !useManifest);
-  const { result, isLoading } = useNodeAttribute(operationInfo, ['connector', 'properties', 'externalDocs'], ['externalDocs']);
+  const result = useNodeAttributeOrSwagger(operationInfo, ['summary'], ['summary'], 'summary', { useManifest });
+  if (result.result === undefined && operationInfo?.operationId) {
+    result.result = titleCase(operationInfo.operationId);
+  }
+
+  console.log(operationInfo, result);
```
ek68794998, line 141: Need to remove this.

Eric-B-Wu (approved):

hartra344:
> @hartra344 Found some time between meetings to update the PR :) Let me know if this meets your expectations.
> 
> Edit: My only confusion was which of these `<div>`s should have `role=button`:
> 
> image

Probably just the one that has focus. I think only the card itself gets focus, inside the card does not but would be good to check.

ek68794998:
> > @hartra344 Found some time between meetings to update the PR :) Let me know if this meets your expectations.
> > Edit: My only confusion was which of these `<div>`s should have `role=button`:
> > image
> 
> Probably just the one that has focus. I think only the card itself gets focus, inside the card does not but would be good to check.

Sounds good. I just validated that Narrator properly reads out when tabbing between cards using the latest build.

image

PR #4053, opened 2024-01-24
feat(designer): Adding support for warnings and custom messages to ErrorsPanel

wifun:
    * Adds capability for host to supply custom errors/warnings to be displayed in ErrorsPanel
    * Adds new `setHostErrorMessages` for hosts to use to supply custom errors/messages

    * no breaking changes

image


wifun:
Still doing validation on my end, leaving this comment up to block merge

wifun (commented):
test

libs/utils/src/lib/helpers/run.ts
```diff
@@ -1,5 +1,5 @@
-import { isObject } from '../helpers';
 import type { CallbackInfo, CallbackInfoWithRelativePath } from '../models';
+import { isObject } from './functions';
```
wifun, line 2: Will revert

libs/designer/src/lib/core/state/panel/panelInterfaces.ts
```diff
@@ -1,6 +1,6 @@
 import type { PanelLocation } from '@microsoft/designer-ui';
 
-export type PanelMode = 'Operation' | 'Discovery' | 'WorkflowParameters' | 'NodeSearch' | 'Error' | 'Connection';
+export type PanelMode = 'Operation' | 'Discovery' | 'WorkflowParameters' | 'NodeSearch' | 'FlowChecker' | 'Connection';
```
wifun, line 3: Still doing validation, adding comment to block merge

hartra344:
why 2 PRs for the same thing? Is the other one valid?

libs/designer/src/lib/core/state/panel/panelInterfaces.ts
hartra344: Will this cause this to be a breaking change to the Hosts?

wifun: yes should be minor, I've called it out in the PR description. For example for PAuto to open the panel we will have to specify the new name.

wifun:
> why 2 PRs for the same thing? Is the other one valid?

Closed it, the other was targeting the version PAuto uses so I could verify it worked for our needs.

libs/designer/src/lib/core/state/panel/panelInterfaces.ts
hartra344: Why are we changing the underlying name then? We should be able to change the text without changing the names under,Error still makes plenty of sense.

wifun: There are a couple components and classnames that use 'Error' which are now shared (ex. nodeErrorCard, nodeErrors, node-error-category-title, etc.).
I thought it be clearer to rename all references of 'ErrorsPanel' to FlowCheckerPanel and remaining references of 'error' to 'message'.

Not a strong preference, so if you would like to keep a mix such as keeping 'errorsPanel.tsx' as is and doing nodeErrors -> nodeMessages.tsx, node-error-category -> node-message-category I can do that

rllyy97 (approved):

rllyy97:
Accidentally linked this PR on an unrelated pr

PR #4064, opened 2024-01-25
fix(designer-ui): Ensure `isAdvanced: false` tokens are shown if far down in a list

ek68794998:

Bug fix

If a set of token outputs from an action has the first six as `isAdvanced: true` and then some output after that has `isAdvanced: false`, that output **should** be shown without clicking "See more"; however, it is currently not.

The first six `isAdvanced: false` tokens will be shown. (I explicitly split out the logic into a helper to make it easier to test.)

No

N/A


libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickerhelpers.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  let filteredTokens = tokens.filter((token) => !token.isAdvanced || showAllOptions || hasSearchQuery);
+  if (!showAllOptions) {
+    filteredTokens = filteredTokens.slice(0, maxRowsShown);
+  }
+
+  return filteredTokens;
+};
+
+export const hasAdvanced = (tokens: OutputToken[]): boolean => {
```
ek68794998, line 17: This is a lift-and-shift from `tokenpickeroption.tsx`.

libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickerhelpers.ts
```diff
@@ -0,0 +1,19 @@
+import type { OutputToken } from '..';
+
+export const getReducedTokenList = (
```
ek68794998, line 3: This is an extraction of logic from `tokenpickeroption.tsx` to filter the tokens list, rather than doing a `return null;` versus `return {component};`.

libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickerhelpers.ts
```diff
@@ -0,0 +1,19 @@
...
+  tokens: OutputToken[],
+  options: { hasSearchQuery: boolean; maxRowsShown: number; showAllOptions: boolean }
+): OutputToken[] => {
+  const { hasSearchQuery, maxRowsShown, showAllOptions } = options;
+
+  let filteredTokens = tokens.filter((token) => !token.isAdvanced || showAllOptions || hasSearchQuery);
```
ek68794998, line 9: The logic change is on lines 9-12. Previously, we checked the index in the filter logic, rather than applying the filter logic _first_ and then taking the first N results.

libs/designer-ui/src/lib/tokenpicker/tokenpickersection/tokenpickerhelpers.ts
ek68794998: The logic here is tested in the first test of the new spec file. The failing logic returns only `Value 1` option, instead of both `Value 1` and `Value 3`.

ek68794998: Change can be seen in: 6da54828aa503248acd56dd9ca3f89cf6a2a6f85

hartra344:
why can't "Visibility: important" be used for this? Why are we adding another concept for this?

hartra344 (approved):

PR #4066, opened 2024-01-25
fix(comsumption): Setting Default Recurrence Interval based on Sku

### LINKED ISSUE #3976: Default trigger polling period has changed from 1 minute in v1 to 3 minute in v3

in v1: 
image

In v3:
image

My ask is to make it 1 minute so that it aligns with what we had in v1. Also, customers might feel like designer is slow because of this setting.

Consumption (Portal)

New Designer

Yes

### Workflow JSON

_No response_

### Screenshots or Videos

_No response_

### Browser

Chrome

### Additional context

_No response_

valentina-vallalta:
Changes default interval values for the recurrence trigger to be based on the sku. Now, there are different default values for standard, consumption, premium, and free. 

To breakdown some of the main changes:

- This PR adds sku as a designer options parameter

- It also adds a custom hook to access the sku so that we can add it to the query. The reason we need to add it to the query is so that we are able to access outside of functional components, namely in the recurrence file.

- Finally, it access the sku in the recurrence file in order to set the correct interval value.

Examples:
-Standard default:
image
-Consumption default:
image

Fixes #3976

### DIFF
--- a/apps/designer-standalone/src/app/AzureLogicAppsDesigner/laDesigner.tsx
+++ b/apps/designer-standalone/src/app/AzureLogicAppsDesigner/laDesigner.tsx
@@ -257,6 +257,7 @@ const DesignerEditor = () => {
           isMonitoringView,
           hostOptions,
           showConnectionsPanel,
+          sku: 'standard',
         }}
       >
         {workflow?.definition ? (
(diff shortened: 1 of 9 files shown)


hartra344:
Can we abstract this away from Designer knowing about SKU? Power automate is technically consumption but it's not the SKU they have. It also locks them out of having other options. Can this be something that is just sent as an option?

hartra344 (approved):

PR #4072, opened 2024-01-26
feat(Designer): Added Consumption run service

rllyy97:
## Main Changes

Added Consumption run service to be used over in Portal.

Mostly the same as the Standard run service, mainly changes with the endpoint formats.


Eric-B-Wu (approved):

libs/services/designer-client-services/src/lib/consumption/run.ts
```diff
@@ -0,0 +1,289 @@
...
+    try {
+      const response = await httpClient.get<ArmResources<Run>>({
+        uri,
+        headers: headers as Record<string, any>,
+      });
+
+      const { nextLink, value: runs }: ArmResources<Run> = response;
+      return { nextLink, runs };
+    } catch (e: any) {
+      throw new Error(e.message);
```
Elaina-Lee, line 132: Is there a reason why we try catch if we just throw another Error with the same message?

PR #4073, opened 2024-01-26
fix(Designer): String Expression Uncasting to Literal On Initialization

Elaina-Lee:

leaves string expression token as an expression instead of uncasting to string literal
on initialization, automatically converts string expression into literal

https://github.com/Azure/LogicAppsUX/assets/144840522/80524221-bc4d-416b-9bd7-d7047a214153


rllyy97 (approved):

Elaina-Lee:
After syncing with Tony, I learned that wrapped/nested string expressions - theoretically - may have problem by not being uncasted. As suggested, I was going to make modification as to remove from uncast logic only if the string expression is "string('')". After checking the uncasting functions, specifically uncastOnce I modified, I realized that it checks for the most outer layer of expression type only (expression.name), which means for nested cases, anything including string expression will still be casted. Therefore, I will leave this change and not make further modifications.

Issue #3781, opened 2023-12-01
Cannot create a new SendGrid Connector in a standard logic app


murdockcrc:
### Describe the Bug with repro steps 

1. Go to the logic app workflow designer 
2. Try to create a new sendgrid connector (send email V4) 
3. The GUI will ask for a name and an API key. 
4. Create the connector. The portal fails with this error:

![image](<https://github.com/Azure/LogicAppsUX/assets/4062119/afdde439-8170-414b-be6f-b9826a30523f>) 

### What type of Logic App Is this happening in? 

Standard (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

![image](<https://github.com/Azure/LogicAppsUX/assets/4062119/f30bcc75-0d51-4054-998a-a610ff2a0c9f>)

### Browser 

MacOS Edge Version 119.0.2151.97 (Official build) (arm64)

### Additional context 

*No response*

AB#26025027

rllyy97:
After talking to some folks that worked on the `testLink` logic this appears to be controlled by the connector owner, and would need to be fixed by them.
However I just tried to repro and it looks like this has been fixed!
Let us know if you have any other difficulties.

Mark-Pearson-APD:
I appear to be having the same issue as above, all teh same details aprt from i'm using a firefox browser - 122.0 (64-bit

rllyy97:
I'm also seeing this issue again, reopening for investigation.

rllyy97:
Looks like the test request is failing specifically when run through our batch api, replicating the request without it returns data as expected and avoids the error

Mark-Pearson-APD:
Appologies @rllyy97, How do i go about making this connection work now within the LogicApp itself? I;m not sure what you mean by removing the batch API, as i don't see that option within the Workflow option?

rllyy97:
@Mark-Pearson-APD, it was an issue on our side!  You shouldn't have to change anything to have it work again, the change is a part of our next release and should be live in all regions by EoD Feb 16th.  Until then I'd recommend using [our fallback url](https://aka.ms/lafallback) to access legacy designer to create your connection.  Once the connection is created it should work fine over in new designer.

Mark-Pearson-APD:
[...]

PR #4103, opened 2024-02-02
fix(Designer): Changed test connection request to not run through batch api

rllyy97:
## Main Changes

Adjusted our connection test request function to not run the request through batch.
During my investigation, running some connector test requests through batch would fail _specifically_ when going through the batch api, with no other changes in the request.
We have a catch over in the portal httpClient where if a request has any explicit headers set, it will not go through batch, so I've just added a dummy header for now.

Fixes https://github.com/Azure/LogicAppsUX/issues/3781


hartra344:
Can we just go ahead and remove batch API all together for now? It’s not batching by default anyway.

rllyy97:
> Can we just go ahead and remove batch API all together for now? It’s not batching by default anyway.

I'm fine with that as well, I can go ahead and make that change over in Portal

PR #4106, opened 2024-02-02
fix(Designer): Added catch for designer options race condition

rllyy97:
## Main Changes

Added a catch to not load graph state until after the host options are available in redux.
Was just a race condition that wasn't accounted for.

Added explicit hooks for `useAreDesignerOptionsInitialized` and `useAreServicesInitialized`

### DIFF
--- a/libs/designer/src/lib/core/state/designerOptions/designerOptionsSlice.ts
+++ b/libs/designer/src/lib/core/state/designerOptions/designerOptionsSlice.ts
@@ -27,6 +27,7 @@ const initialState: DesignerOptionsState = {
   isMonitoringView: false,
   isDarkMode: false,
   servicesInitialized: false,
+  designerOptionsInitialized: false,
   useLegacyWorkflowParameters: false,
   isXrmConnectionReferenceMode: false,
   showConnectionsPanel: false,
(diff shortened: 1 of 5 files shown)


Elaina-Lee (approved):

libs/designer/src/lib/core/state/designerOptions/designerOptionsSlice.ts
```diff
@@ -111,6 +111,7 @@ export const designerOptionsSlice = createSlice({
         ...state.hostOptions,
         ...action.payload.hostOptions,
       };
+      state.designerOptionsInitialized = true;
```
takyyon, line 114: just checking, should this be "state.designerOptions.designerOptionsInitialized" ?

takyyon: Also, I am guessing this is the default value? It should be "false" right?

rllyy97: Since we're in the designerOptions slice, the `state` variable here is already in the `designerOptions` section of the state, so just `state.designerOptionsInitialized` is correct
And the default value for this is just undefined but I can make it explicit as false as the default state.  This call should be true since after this is run it shows that we initialized the designer options state.

takyyon (approved):

Issue #4087, opened 2024-01-31
Designer crashes when naming a Compose action "toString"


jaimereyesmx:
### Describe the Bug with repro steps 

1. Create a flow in Azure logic apps with any trigger and a Compose action 
2. Save the flow to validate this works properly. 

![image](<https://github.com/Azure/LogicAppsUX/assets/12244548/01f081d2-f92e-45dd-8551-25681f2bb585>)

3. Rename the Compose action to "toString" 

![image](<https://github.com/Azure/LogicAppsUX/assets/12244548/d59ae341-361c-4cd1-859b-7d1b5419c2d7>)

4. Click Save 
5. Navigate away and then try to open the flow to see the crash 

![image](<https://github.com/Azure/LogicAppsUX/assets/12244548/7c5845e6-a381-4b5d-a29e-479ed0e33174>)

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Edge

### Additional context 

*No response*

hartra344:
@rllyy97 can you take a look at this tomorrow? Designer crashes are hard GA stoppers

rllyy97:
The issue appears to be that naming a node like `toString` is causing a lot of property overlap all over the project.  Since prototype functions can still be referenced by bracket notation (which we use all over the app to get our node record data), asking for any `toString` node data is returning the `toString` function instead.

The proper way around this is to check records with `Object.hasOwn` before using bracket notation to grab the data, but that would be a _huge_ amount of reference changes on _every_ record.  For now the easiest option would be to check the name during renaming and just have prototype functions declared as invalid values, or append a whitespace character so it appears the same but references correctly.

PR #4109, opened 2024-02-02
fix(Designer): Prototype function names are now invalid node names

rllyy97:
## Main Changes

Nodes cannot be named the same as prototype functions now.
Ex: `toString`, `toLocaleString`

Naming a node a prototype function broke lots of the app.

Fixes:  https://github.com/Azure/LogicAppsUX/issues/4087


hartra344 (changes requested):
Actually how does this work if they already have action with the name? Would it be possible to allow these but not crash?

rllyy97:
> Actually how does this work if they already have action with the name? Would it be possible to allow these but not crash?

I'm working on that now yeah, that's why I left it as a draft, it's looking like we'll need to change _every_ instance of fetching data from a record to catch this situation (which is a lot)

rllyy97:
Closing in favor of this PR: https://github.com/Azure/LogicAppsUX/pull/4140

PR #4112, opened 2024-02-03
fix(Designer): Skip non-dependent parameters in updateParameterAndDependencies

pranaydubeymicrosoft:
Add a nullcheck to prevent parameters that are not dependent from being added to the payload for update.

This fixes a bug whereby the Trello create a card action in Power Automate would fail. Currently, .[*] wildcard parameters missing the id field will cause updateParameterAndDependencies to fail. The result is that the value is never set in the designer. 

With fix:
image

image

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/helper.ts
+++ b/libs/designer/src/lib/core/utils/parameters/helper.ts
@@ -1705,7 +1705,7 @@ export async function updateParameterAndDependencies(
           LoggerService().log({
             level: LogEntryLevel.Verbose,
             area: 'UpdateParameterAndDependencies',
-            message: `Connection name: ${connectionReference.connectionName} - Parameter key: ${key}`,
+            message: `Dependent parameter was not set. Connection name: ${connectionReference.connectionName} - Parameter key: ${key}`,
           });
           continue;
         }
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1700,9 +1700,15 @@ export async function updateParameterAndDependencies(
...
       if (inputDependencies[key].dependencyType === 'ListValues' && inputDependencies[key].dependentParameters[parameterId]) {
-        const dependentParameter = nodeInputs.parameterGroups[groupId].parameters.find(
-          (param) => param.parameterKey === key
-        ) as ParameterInfo;
+        const dependentParameter = nodeInputs.parameterGroups[groupId].parameters.find((param) => param.parameterKey === key);
+        if (!dependentParameter) {
+          LoggerService().log({
+            level: LogEntryLevel.Verbose,
+            area: 'UpdateParameterAndDependencies',
+            message: `Connection name: ${connectionReference.connectionName} - Parameter key: ${key}`,
```
ek68794998, line 1708: Can we add a message as to what went wrong here?

    message: `Dependent parameter was not set. Connection name: ${connectionReference.connectionName} - Parameter key: ${key}`,

ek68794998: This just covers us in case more logging is added later and makes telemetry easier to follow.

hartra344 (approved):

PR #4134, opened 2024-02-06
feat(designer): Resizable side panel

ccastrotrejo:
The changes include the introduction of a resizable panel feature and the addition of new styles and dependencies.

* Introduced state variables `isResizing` and `width` to manage the resizing process. Also, added `startResizing` and `stopResizing` functions to control the resizing state.
* Added a `resize` function to handle the resizing process and attached it to the `mousemove` event. The `mouseup` event is attached to the `stopResizing` function. Both event listeners are cleaned up in the `useEffect` hook.
* Modified the `Panel` component to support the new resizable feature and added a new `div` element which triggers the resizing process on `mousedown` event. 
* Minimum size is 300px

https://github.com/Azure/LogicAppsUX/assets/102700317/f5baef75-dfed-438b-a55e-3532e9c13298

### DIFF
--- a/libs/designer/src/lib/ui/panel/panelRoot.tsx
+++ b/libs/designer/src/lib/ui/panel/panelRoot.tsx
@@ -59,15 +59,13 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
 
   const collapsed = useIsPanelCollapsed();
   const currentPanelMode = useCurrentPanelMode();
+
   const [isResizing, setIsResizing] = useState(false);
-  const [width, setWidth] = useState<any>(PanelSize.Auto);
-  const [mouseX, setMouseX] = useState(0);
-  const startResizing = useCallback((test: any) => {
-    setMouseX(test.clientX);
-    return setIsResizing(true);
-  }, []);
+  const [width, setWidth] = useState<PanelSize | string>(PanelSize.Auto);
+  const startResizing = useCallback(() => setIsResizing(true), []);
   const stopResizing = useCallback(() => setIsResizing(false), []);
   const animationFrame = useRef<number>(0);
+  const isResizable = currentPanelMode === 'Assertions';
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/ui/panel/panelRoot.tsx
```diff
@@ -70,14 +99,39 @@ export const PanelRoot = (props: PanelRootProps): JSX.Element => {
     </div>
   );
 
+  const resize = useCallback(
+    ({ clientX }: MouseEvent) => {
+      animationFrame.current = requestAnimationFrame(() => {
+        if (isResizing) {
+          const newWidth = window.innerWidth - clientX < 300 ? 300 : window.innerWidth - clientX;
```
rllyy97, line 106: Nit: Can use the `Math.max()` function here instead of the  conditional
ie. `Math.max(window.innerWidth - clientX, 300)`

rllyy97 (approved):
Very cool

libs/designer/src/lib/ui/panel/panelRoot.tsx
ccastrotrejo: Sure thing, thanks!

libs/designer/src/lib/ui/panel/panelRoot.tsx
```diff
@@ -28,15 +28,44 @@ const layerProps = {
   eventBubblingEnabled: true,
 };
 
+const useStyles = makeStyles({
```
ek68794998, line 31: Has this been tested with panel on the left also?

ccastrotrejo: Hi @ek68794998, yep. As for now this feature will only support right side panels.

ek68794998: Cool, thanks for keeping that in mind!

PR #4142, opened 2024-02-07
fix(vscode): Remove update local.settings.json and api calls in deployment scripts

ccastrotrejo:
The changes aim to streamline the codebase, improve the user experience, and enhance the maintainability of the code. 

* Removed unnecessary imports and the `SaveAzureContext` class. The `createAzureWizard` function no longer requires the `projectPath` parameter, and the `SourceControlPathListStep` class has been changed from a prompt step to an execute step. The `ConfigureInitialLogicAppStep` class now includes a `showLoadingPrompt` property. 

* The `callStandardApi` function no longer requires the `scriptContext` parameter. The `callConsumptionApi` function and the `callManagedConnectionsApi` function have been updated to use `localSubscriptionId` and `localResourceGroup` instead of `subscriptionId` and `resourceGroup`. The `gatherAndValidateInputs` function now always launches the Azure Wizard and includes `localSubscriptionId` and `localResourceGroup` in the returned object.


ccastrotrejo (commented):
Self reviewed

PR #4174, opened 2024-02-12
fix(Designer): Fixed issue causing schema parameters to not appear as required

rllyy97:
## Main Changes

Fixed issue causing schema parameters to not appear as required.
Parameters were ignoring the defined `required` key array prop if their parent in their path (like `body`) wasn't also `required: true`.
With APIM, the body param isn't able to be set as required, which was causing this issue.


ccastrotrejo (approved):

rllyy97:
Closing, will revisit

PR #4175, opened 2024-02-12
feat(vscode): Binaries dependencies opt-in  by default

ccastrotrejo:
This pull request primarily focuses on setting the opt-in as default for binaries installation.

- Reset binaries -> set the auto installation as true and binaries settings path to undefined
- Disabled binaries -> set the auto installation as false and binaries settings path to undefined
- Binaries settings paths are set later in logic.

* A new function `disableValidateAndInstallBinaries` has been added to disable the auto validation and installation of binary dependencies.
* The `promptInstallBinariesOption` function has been renamed to `installBinaries` and its install always installs the binaries unless the setting is set to true.
* The default value for the `autoRuntimeDependenciesValidationAndInstallation` setting has been changed from `null` to `true` in `package.json`.


ccastrotrejo (commented):
Self reviewed

hartra344 (approved):

PR #4191, opened 2024-02-14
fix(designer): Fix two places where `undefined` switch `cases` caused fatal errors

ek68794998:

Bug fix

If an empty switch block is loaded into designer, it will break the panel, and cannot have any cases added to it.

Above are fixed and behave as expected.

Context: `subOperation` could be `undefined`, and so `Object.keys()` call would fail. Also, use of `any` prevented the `cases == undefined` case from being type-checked.

No

**Before**

switch-broken

**After**

switch-working

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
@@ -292,7 +292,7 @@ const processChildGraphAndItsInputs = (
   if (subGraphDetails) {
     for (const subGraphKey of Object.keys(subGraphDetails)) {
       const { inputs, inputsLocation, isAdditive } = subGraphDetails[subGraphKey];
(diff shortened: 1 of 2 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
```diff
@@ -292,7 +292,7 @@ const processChildGraphAndItsInputs = (
   if (subGraphDetails) {
     for (const subGraphKey of Object.keys(subGraphDetails)) {
       const { inputs, inputsLocation, isAdditive } = subGraphDetails[subGraphKey];
-      const subOperation = getPropertyValue(operation, subGraphKey);
+      const subOperation = getPropertyValue(operation, subGraphKey) || {};
```
hartra344, line 295:
```suggestion
      const subOperation = getPropertyValue(operation, subGraphKey) ?? {};
```

hartra344 (approved):

PR #4204, opened 2024-02-15
fix(designer): Parameter default value from manifest should be honored at load time

### LINKED ISSUE #3699: Default values not populated when opening a template

During our tests, we noticed an issue when creating a flow from a template with designer v3.

 Here is an [example](<https://make.test.powerautomate.com/galleries/public/templates/bc6f31a29dbc47fdb05f37829e7f5db9?environment=a631a81e-a556-4b40-9a40-09c9f4c11bce%c2%b6meters.aibuilder.modelId=189c0d28-c2ee-498a-b637-105b2892184c%c2%b6meters.aibuilder.data1=Total:%20@%7boutputs%28%27Extract_information_from_forms%27%29?%5b%27body%2fresponsev2%2fpredictionOutput%2flabels%2fTotal%2fvalue%27%5d%7d%20%28Confidence%20score:%20@%7boutputs%28%27Extract_information_from_forms%27>)%3F%5B%27body%2fresponsev2%2fpredictionOutput%2flabels%2fTotal%2fconfidence%27%5D%7D)&parameters.aibuilder.label1Value=@%7Boutputs(%27Extract\_information\_from\_forms%27)%3F%5B%27body%2fresponsev2%2fpredictionOutput%2flabels%2fTotal%2fvalue%27%5D%7D&powerautomate-core.forceTip2InTest=false).

After digging into LAUX codebase, I think the issue is there. It doesn't use the default value from the manifest when it should. This is most likely affecting any operation in similar situations and not only AIB.

Here is what we have in our manifest:

image

and we end up with an empty value which can't even be fixed by the user, since the parameter is internal.

image (1)

I think the issue comes from [getInputParametersFromManifest](<https://github.com/Azure/LogicAppsUX/blob/main/libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts>) which calls [updateParameterWithValues](<https://github.com/Azure/LogicAppsUX/blob/main/libs/designer/src/lib/core/utils/parameters/helper.ts>) which doesn't take the default value from the manifest and leaves "value" to undefined. Unless this is on purpose and it was meant to be assigned later on in a different part of the codebase, but it didn't feel like it.

PS: This was initially posted in a teams channel a few weeks ago, the link to the source in GH may not be accurate anymore.

Power Automate

New Designer

No

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Any browser

### Additional context 

*No response*

AB#25849468

ek68794998:

Bug fix

Default values from manifests are being ignored when loading parameters.

image

image

Fixed, see screenshots. Closes #3699 (cc @speray).

No

image


ek68794998:
Checking with @speray to make sure this fix works as expected in his test case.

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -855,7 +855,9 @@ export function shouldIncludeSelfForRepetitionReference(manifest: OperationManif
 }
 
 export function loadParameterValue(parameter: InputParameter): ValueSegment[] {
-  const valueObject = parameter.isNotificationUrl ? `@${constants.HTTP_WEBHOOK_LIST_CALLBACK_URL_NAME}` : parameter.value;
+  const valueObject = parameter.isNotificationUrl
```
hartra344, line 858: At this point can we break this down to an actual if/else, mixing ternaries with pretty much anything makes readability difficult. (I say this even as a ternary expression apologist)

ek68794998: Yup, will do. I tried to break it down with `()`s but Prettier was having none of that.

ek68794998:
Confirmed a repro case with @speray and was able to test that it fails before this fix, and works after.

PR #4229, opened 2024-02-20
Feat(Designer): Moving first lib into shared folder

DanielleCogs:
Starting the process of combining some of our libs into one. When this is done, designer-client-services, intl, parsers, utils and designer-ui will be included in the same library. This will be fairly disruptive, so I am trying to break up the process into multiple PRs.

moving parsers lib into a new library that will hold multiple in the future
NA
NA
Yes


hartra344 (approved):

libs/designer/src/lib/core/utils/parameters/__test__/helper.spec.ts
```diff
@@ -510,7 +510,7 @@ describe('core/utils/parameters/helper', () => {
           {
             id: '6',
             type: ValueSegmentType.TOKEN,
-            value: 'item()?[\'someItem\']',
```
ccastrotrejo, line 513: Does these changes mean that something happened/changed with the original function that the test ended up updating it?

DanielleCogs: I tried changing this back to what it was here, and running prettier on it. The result is the same (back to double quote). I changed it back again and ran 'lint' but it did not change the file (kept single quote). I think the previous commit must have been made using some difference in prettier, either config or version. Worth looking into but I think this can still be merged. Came across this when researching... https://github.com/prettier/prettier/issues/973

ek68794998:
@DanielleCogs What changes do we need to make to consume the package following this update?

Issue #4196, opened 2024-02-14
Unsupported functions shown in expression editor


kisubedi:
### Describe the Bug with repro steps 

![image](<https://github.com/Azure/LogicAppsUX/assets/91279975/b9510ee0-0362-4fa2-865d-b3505252e8f9>)

We should hide the functions until we have support for it. Can't make it work, also documentation is missing.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Chrome

### Additional context 

*No response*

kisubedi:
![image](https://github.com/Azure/LogicAppsUX/assets/91279975/b10f3467-1802-4088-8a7b-f93758b37237)
simple invokation of function fails- utf8length('test')

PR #4231, opened 2024-02-21
fix(desinger): made utfLength Expressions optional based on host options

Eric-B-Wu:
fixes https://github.com/Azure/LogicAppsUX/issues/4196

to remove utf expressions, please add `hideUTFExpressions: true` to host options

### DIFF
--- a/libs/designer-ui/src/lib/editor/monaco/index.tsx
+++ b/libs/designer-ui/src/lib/editor/monaco/index.tsx
@@ -60,6 +60,7 @@ export interface MonacoOptions {
   lineHeight?: number;
   minimapEnabled?: boolean;
   scrollBeyondLastLine?: boolean;
+  hideUTFExpressions?: boolean;
   wordWrap?: 'off' | 'on' | 'wordWrapColumn' | 'bounded';
   wordWrapColumn?: number;
   contextMenu?: boolean;
@@ -82,6 +83,7 @@ export const MonacoEditor = forwardRef<editor.IStandaloneCodeEditor, MonacoProps
       minimapEnabled = false,
       value,
       scrollBeyondLastLine = false,
+      hideUTFExpressions,
       height,
       width,
       lineNumbersMinChars,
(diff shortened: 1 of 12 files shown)


libs/designer-ui/src/lib/editor/monaco/index.tsx
```diff
@@ -115,21 +117,21 @@ export const MonacoEditor = forwardRef<editor.IStandaloneCodeEditor, MonacoProps
...
-    const initTemplateLanguage = async () => {
-      const { languages, editor } = await loader.init();
-      if (!languages.getLanguages().some((lang: any) => lang.id === Constants.LANGUAGE_NAMES.WORKFLOW)) {
-        registerWorkflowLanguageProviders(languages, editor);
-      }
-      setCanRender(true);
-    };
-
     useEffect(() => {
+      const initTemplateLanguage = async () => {
```
takyyon, line 121: Why not just add the implementation directly on Line 130 ?

libs/designer-ui/src/lib/workflow/languageservice/workflowlanguageservice.ts
```diff
@@ -446,3 +450,7 @@ export function getTemplateFunctions(): FunctionDefinition[] {
 function signatureHasVariableParameters(signature: SignatureInfo): boolean {
   return signature.parameters.some((parameter) => !!parameter.isVariable);
 }
+
+export function removeUTFExpressions(functionGroupFunctions: FunctionDefinition[]): FunctionDefinition[] {
+  return functionGroupFunctions.filter((func) => func.name !== 'utf8Length' && func.name !== 'utf16Length');
```
takyyon, line 455: case insensitive check?

libs/designer-ui/src/lib/editor/monaco/index.tsx
Eric-B-Wu: It is a dependency of the useEffect and registerWorkflowLanguageProviders is what initializes the template functions, I can however, move initTemplateLanguage into usecallback

takyyon (approved):

PR #4233, opened 2024-02-21
feat(Designer): Hybrid preload / active search

rllyy97:
## Main Changes

Implemented hybrid search to do both preloading and active searching depending on the state of the search data.
**_Search in general will feel much faster than it did before, both for browse and search_**

One thing to note is that the active search is not fully fuzzy.  It still fuzzy searches over preloaded data combined with the active searched data, but the data returned from the active search is the same request we make in legacy designer, just a raw string search.

Once the preload finishes, it stops active searching and only does full fuzzy searches on the local data.

Fixed bug where preloading was enabled on monitoring view and disabled otherwise
Fixed memoization bug where we would sometimes make multiple search page requests

### Old (~19 seconds to search for "Microsoft Forms" from designer load, not impacted by search time)
OLD_SEARCH

### New (~7 seconds from designer load, ~3 seconds from executing search)
NEW_SEARCH


libs/designer/src/lib/ui/panel/recommendation/recommendationPanelContext.tsx
```diff
@@ -45,16 +46,48 @@ export const RecommendationPanelContext = (props: CommonPanelProps) => {
...
+    [preloadedOperations, activeSearchOperations]
+  );
+
+  // Active search
+  useDebouncedEffect(
+    () => {
+      // if preload is complete, no need to actively search
+      if (!isLoadingOperations) return;
+      if (searchTerm === '') return;
+      if (searchedTerms.includes(searchTerm)) return;
```
takyyon, line 76: Can we bundle all of this in one "if - OR" ?

PR #4235, opened 2024-02-21
fix(consumption): Changing until timeout default based on Stateful vs Stateless

### LINKED ISSUE #2166: Stateless workflow until loop timeout defaults to 1 hour instead of 5 minutes
### Describe the Bug 

When creating a stateless workflow in the VS Code extension and then adding an Until loop it sets the Timeout value to PT1H as default. This is violating the maximum allowed time that a stateless workflow can run (5 minutes).

This results in some strange error messages and after some digging in the error logs I found the reference to that setting.

### Plan Type 

Standard

1. Create a new stateless workflow in any project. 
2. Add an until loop. 
3. Expand the "Change Limits" and see that the Timout setting is set to PT1H (1 hour) 

### Workflow JSON 

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "Until": {
                "type": "Until",
                "expression": "@equals('', '')",
                "limit": {
                    "count": 60,
                    "timeout": "PT1H"
                },
                "actions": {},
                "runAfter": {}
            }
        },
        "triggers": {
            "manual": {
                "type": "Request",
                "kind": "Http",
                "inputs": {}
            }
        },
        "contentVersion": "1.0.0.0",
        "outputs": {}
    },
    "kind": "Stateless"
}
```

### Screenshots or Videos

image

### Additional context

*No response*

AB#16636564

valentina-vallalta:
This PR fixes the "Until" action timeout default to be different if the logic app is stateless or stateful. 

It does so by adding the workflowKind (stateful or stateless) to the query in the Designer.tsx file and accessing it when we retrieve and set the default value in the helper.ts file. The initial value of workflowKind is undefined until the value is set and so we needed to write the query in a way that will store the right value once it is defined.
We also needed to update the manifest to include a stateless default option that we access when we set the default value.

Stateless default is PT5M, or 5 minutes:

Stateful default is PT1H, or 1 hour. This is also the default for consumption logic apps:

fixes #2166


takyyon:
Do you need package-lock changes here?

takyyon (approved):

PR #4249, opened 2024-02-23
fix(vscode): Fix useEffect on workflows success data - export tool

ccastrotrejo:
This pull request introduces changes to the `WorkflowsSelection` component in the `workflowsSelection.tsx` file. The changes primarily focus on simplifying the component's logic by removing the `useCallback` hook and integrating its functionality directly into the `useEffect` hook. 

Here are the main changes:

* Removed the `useCallback` hook named `onWorkflowsSuccess`. This hook was previously used to process the workflows data and update several state variables. The logic of this hook has been moved directly into a `useEffect` hook that runs when `workflowsData` changes. 
* Removed the import of `useCallback` from the `react` package, as it is no longer used in this file.

These changes simplify the component's logic and may improve performance by reducing the number of callbacks created.

#### Issue

https://github.com/Azure/LogicAppsUX/assets/102700317/829a932d-c62b-46eb-a0b9-4694b93e96e3

image

#### Fix

https://github.com/Azure/LogicAppsUX/assets/102700317/327b6db6-bfe0-4006-9c3e-67158f0ba0ec


ccastrotrejo (commented):
Self reviewed

ccastrotrejo (commented):
Self-reviewed

apps/vs-code-react/src/app/export/workflowsSelection/helper.ts
```diff
@@ -113,3 +113,31 @@ export const getAdvanceOptionsSelection = (
...
+ * @returns An array of workflows with updated selection and rendering status.
+ */
+export const parseSelectedWorkflows = (workflowsData: WorkflowsList[], allItemsSelected: SelectedWorkflowsList[]) => {
+  const currentSelection = [...allItemsSelected];
+  return workflowsData.map((workflow: WorkflowsList) => {
+    const selectedIndex = currentSelection.findIndex((selectedItem: SelectedWorkflowsList) => selectedItem.key === workflow.key);
+    if (selectedIndex !== -1) {
+      return { ...workflow, selected: true, rendered: true };
+    }
+    return { ...workflow, selected: false, rendered: true };
```
takyyon, line 141:
```suggestion
    return { ...workflow, selected: selectedIndex !== -1, rendered: true };
```

takyyon (approved):

apps/vs-code-react/src/app/export/workflowsSelection/helper.ts
ccastrotrejo: Ohh yes, thats really smart. Thanks for pointing this out. I will make the change!

Issue #4246, opened 2024-02-22
Pasted HTML code in email editor is garbled.


kisubedi:
### Describe the Bug with repro steps 

In send email action, toggle to HTML view (<>), and paste the following code: Save the flow and notice that the pretty formatting has changed so it's impossible to make sense of the HTML, and also resulting formatting in email is different.

![image](<https://github.com/Azure/LogicAppsUX/assets/91279975/275277fb-f04b-4732-baf7-56123f074e02>)

<title>Simple Transactional Email</title>

[...]

PR #4284, opened 2024-02-29
fix(designer): Pasted HTML code in HtmlEditor shows raw text

pranaydubeymicrosoft:
Fixes https://github.com/Azure/LogicAppsUX/issues/4246

Email now displays text and not underlying html:

image

Previously the email would send like so:

image

### DIFF
--- a/libs/designer-ui/src/lib/html/plugins/toolbar/helper/__test__/util.spec.ts
+++ b/libs/designer-ui/src/lib/html/plugins/toolbar/helper/__test__/util.spec.ts
@@ -13,20 +13,63 @@ import type { ValueSegment } from '@microsoft/designer-client-services-logic-app
 
 describe('lib/html/plugins/toolbar/helper/util', () => {
   describe('cleanHtmlString', () => {
-    it.each([
-      ['<p>text1<span>\n</span>text2</p>', '<p>text1<br>text2</p>'],
-      ['<p>text</p>', '<p>text</p>'],
-      ['<p>text1</p><p><br></p><p>text2</p>', '<p>text1</p><br><p>text2</p>'],
-      ['<p>text1<br></p><p><br></p><p>text2</p>', '<p>text1</p><br><br><p>text2</p>'],
-      ['<span>text</span>', 'text'],
-      ['<span id="$[abc,variables(\'abc\'),#770bd6]$">text</span>', '<span id="$[abc,variables(\'abc\'),#770bd6]$">text</span>'],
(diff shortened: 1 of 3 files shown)


libs/designer-ui/src/lib/html/plugins/toolbar/helper/__test__/util.spec.ts
```diff
@@ -13,20 +13,63 @@ import type { ValueSegment } from '@microsoft/designer-client-services-logic-app
...
+        ['<span id="$[concat(),concat(\'&lt;\'),#ad008c]$">text</span>', '<span id="$[concat(),concat(\'&lt;\'),#ad008c]$">text</span>'],
+        [
+          '<span id="$[concat(),concat(\'%26lt;\'),#ad008c]$">text</span>',
+          '<span id="$[concat(),concat(\'%26lt;\'),#ad008c]$">text</span>',
+        ],
+        ['<span id="@{variables(\'abc\')}">text</span>', '<span id="@{variables(\'abc\')}">text</span>'],
+        ['<span id="@{concat(\'&lt;\')}">text</span>', '<span id="@{concat(\'&lt;\')}">text</span>'],
+        ['<span id="@{concat(\'%26lt;\')}">text</span>', '<span id="@{concat(\'%26lt;\')}">text</span>'],
+      ])('should properly convert HTML: %p', (input, expected) => {
+        expect(cleanHtmlString(input, /* isValuePlaintext */ true)).toBe(expected);
```
ek68794998, line 43: One way you can avoid having to specify `/* isValuePlaintext */` is by doing the "options object" pattern, if you like.

image

ek68794998: `libs\designer-ui\src\lib\editor\base\utils\parsesegments.ts` also has examples of this.

pranaydubeymicrosoft: Will leave as is, should be clear what is being tested.

hartra344: I don't think the point here is that it's unclear what's being tested. It's unclear what true/false here means when it's being used in live code. Using booleans as a parameter is generally bad form and leads to bad readability/maintainability. That's not to say we don't have it in other parts of code(generally code that was carried over from old designer. But it would be better if it was clearer and an options object is quite good for that.

hartra344 (changes requested):
Blocking on the suggestion above

hartra344:
I just tested the fix and I also don't think this is actually working in all cases

hartra344:
Closing due to inactivity. Feel free to reopen when issues are addressed and ready to review

PR #4299, opened 2024-03-02
feat(designer): Allow host to define conditions for built-in and custom connectors

### LINKED ISSUE #1811: Logic for built-in and custom connectors should be customizable
### Describe the Bug

This is more of a feature request than a bug, but should be kept here to be tracked as work for host service consumers.

Currently, logic for what decides which actions go under which dropdown is done by helper methods in `libs\utils\src\lib\helpers\connections.ts`. However, this logic is specific to Logic Apps' implementation, and does not apply to others consuming the designer.

In our host service's case, none of our connectors (including built-ins and custom) have an ID starting with `/subscriptions/`, so they all meet the following check:

```ts
export function isArmResourceId(resourceId: string): boolean {
  return resourceId ? resourceId.startsWith('/subscriptions/') : false;
}

export const isBuiltInConnector = (connectorId: string) => {
  return !isArmResourceId(connectorId);
};
```

This results in _all_ actions showing as built-in for us. (See images below.)

N/A. Simply need to have built-in actions that are not based on ARM resource IDs.

### Workflow JSON

```json
N/A, related only to search service & recommendation panel.
```

### Expected behavior

There should be a way to customize the logic of what is considered a built-in or custom connector, and it should not rely only on the connector ID.

### Screenshots or Videos

image

### Platform And Browser

N/A

### Additional context

_No response_

ek68794998:

Mix of bug fix and feature

All built-ins are presumed to have connector IDs beginning with `/subscriptions`, but Power Automate does not follow this format. There are also rules for custom connectors defined which do not align with Power Automate's.

New host service functions for built-in and custom connector matching can be used by hosts to define their own logic for this. Closes #1811.

Yes:

1. Function `getBrandColorFromConnector` was moved from `@microsoft/logic-apps-designer` to `@microsoft/designer-ui`.
1. Function `getIconUriFromConnector` was moved from `@microsoft/logic-apps-designer` to `@microsoft/designer-ui`.

**Before:**

image

image

image

image

**After:**

image

image

image

image


hartra344 (commented):
Don't move functions away from the primary designer library. Were working on consolidating libraries and this makes it more difficult

hartra344:
logic-apps-shared would probably be safest if it would work there since that's the congregation point at the moment

ek68794998:
@hartra344 I remember now why I didn't move them there: having them in `logic-apps-shared` is problematic because of the use of `HostService`, which comes from the client services library, causing a circular dependency:

image

My understanding is that only the following libraries can reference the client services library:

- `chatbot`
- `designer`
- `designer-ui`
- `vscode-extension`

So if I need a function using `HostService` to be accessible in both `designer` and `designer-ui`, it seems like `designer-ui` is the only place where it can safely go. Thoughts?

hartra344:
They're fairly small so I guess this should be fine for now....

hartra344:
@rllyy97 before I sign off can you just verify this doesn't cause any issues for us? It doesn't look like it should but I'd rather way for you to sign off to be sure

libs/designer-ui/src/lib/utils/utils.ts
```diff
@@ -348,7 +349,7 @@ export const filterRecord = <T>(data: Record<string, T>, filter: (_key: string,
     .reduce((res: any, [key, value]: any) => ({ ...res, [key]: value }), {});
 };
 
-export const getConnectorCategoryString = (connectorId: string): string => {
+export const getConnectorCategoryString = (connector: Connector | OperationApi | string): string => {
```
rllyy97, line 352: Is there a reason we're keeping this parameter typed as `| string`?  It looks like it should only work on Connector / Api objects now.

And I'm guessing so, but did you adjust all references of this method?

ek68794998: There are a handful of places where we still pass a string.

- `libs\designer\src\lib\ui\panel\connectionsPanel\allConnections\connectorConnectionsCard.tsx` &mdash; This can have `connector === undefined` so it falls back on `connectorId`.
- `libs\designer\src\lib\ui\panel\nodeDetailsPanel\tabs\aboutTab.tsx` &mdash; This does not have access to the `Connector` object because it comes from `operationInfo`, which only has the ID.

All other calls have access to the `Connector` or `OperationApi` object.

PR #4305, opened 2024-03-04
feat(vscode):  Improve status step indicator for export experience

ccastrotrejo:
This pull request primarily focuses on improving the user feedback during the workflow export process in the VS Code Designer and React applications. The changes include adding success messages in the `ExportEngine` class, updating the status rendering in the React application, and adding new styles for the status items.

* Added success messages in the `ExportEngine` class to inform the user when the workflows are exported successfully. These messages are added in two places, once when the resource group name is not provided or the template does not exist, and once after the parameters and settings are updated.

* Updated the `renderStatus` function to include an icon and a spinner for the status items. The spinner is shown for the last item if the final status is not 'Succeeded'. Also, the `FinalStatusGadget` component now takes `finalStatus` as a prop and displays a check icon when the final status is 'Succeeded'. 

* Added new styles for the status items. Each status item is displayed as an inline-flex item with centered alignment.

#### Before
Screenshot 2024-03-02 at 2 01 44 PM

#### After
Screenshot 2024-03-02 at 2 06 02 PM

https://github.com/Azure/LogicAppsUX/assets/102700317/d541768b-9bf4-4ebd-8b01-87a7a6cc2de8

### DIFF
--- a/apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
+++ b/apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
@@ -50,6 +50,22 @@ interface Deployment {
 }
 
 class ExportEngine {
+  private intlText = {
+    SUCESSFULL_EXPORTED_MESSAGE: localize('workflowsExportedSuccessfully', 'The selected workflows exported successfully.'),
+    DONE: localize('done', 'Done.'),
+    DEPLOYING_CONNECTIONS: localize('deployConnections', 'Deploying connections ...'),
+    DOWNLOADING_PACKAGE: localize('downloadingPackage', 'Downloading package ...'),
+    UNZIP_PACKAGE: localize('unzipPackage', 'Unzipping package ...'),
+    FETCH_CONNECTION: localize('fetchConnectionKeys', 'Retrieving connection keys ...'),
+    UPDATE_FILES: localize('updateFiles', 'Updating parameters and settings ...'),
+  };
+
+  private finalStatus = {
+    InProgress: 'InProgress',
+    Succeeded: 'Succeeded',
+    Failed: 'Failed',
+  };
+
   public constructor(
     private getAccessToken: () => string,
     private packageUrl: string,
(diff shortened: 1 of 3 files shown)


apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
```diff
@@ -83,6 +83,7 @@ class ExportEngine {
       const templateExists = await fse.pathExists(templatePath);
       if (!this.resourceGroupName || !templateExists) {
         this.setFinalStatus('Succeeded');
+        this.addStatus(localize('workflowsExportedSuccessfully', 'The selected workflows exported successfully.'));
```
takyyon, line 86: Can we add this - `localize('workflowsExportedSuccessfully', 'The selected workflows exported successfully.')` to a constant and use it all the places instead?

apps/vs-code-react/src/app/export/status/status.tsx
```diff
@@ -26,28 +39,34 @@ export const Status: React.FC = () => {
...
+  switch (finalStatus) {
     case FinalStatus.Succeeded:
       return (
-        <Text block>
-          {message} {targetDirectory.path}/.logs/export/README.md
-        </Text>
+        <div className="msla-export-status--item">
+          <Icon iconName={'SkypeCheck'} />
+          <Text style={{ marginLeft: '5px' }}>
+            {message} {targetDirectory.path}/.logs/export/README.md
```
takyyon, line 67: We should have this path "/.logs/export/README.md" as a constant so others touching this code have an idea of what it really means/does

takyyon (approved):
Just a couple of small ones (Approving with suggestions)

apps/vs-code-react/src/app/export/status/status.tsx
ccastrotrejo: Sure! thanks

apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
ccastrotrejo: Sure! thanks

PR #4314, opened 2024-03-05
fix(consumption): Adding 'Invalid Parameter' message to Subgraph nodes

### LINKED ISSUE #4278: Switch cases don't show errors on node

1. Have a switch case in a workflow 
2. Add a new case to the switch 
3. Add a value to the case's "Equals" parameter 
4. Remove the parameter to update the error checking 
5. The error should appear in the parameters and in the errors command bar panel, but not on the action node in the canvas 

image

Consumption (Portal)

New Designer

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Chrome

### Additional context 

*No response*

valentina-vallalta:
Adding parameter error messaging to the subgraph scope nodes. This is the only type of error that subgraph nodes can have since all other types of errors are handled at the scope level. Before, there would only be error messaging in the node panel itself and in the errors command bar panel but now it will display a message in the subgraph node itself, how it does for other nodes. 

Before:
image

After:
imagefixes #4278

### DIFF
--- a/libs/designer-ui/src/lib/card/subgraphCard/subgraphCard.less
+++ b/libs/designer-ui/src/lib/card/subgraphCard/subgraphCard.less
@@ -32,11 +32,38 @@
     color: white;
     flex-grow: 1;
     align-self: center;
-    display: inline-block;
-    font: 14px/20px normal @semibold-font-family;
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
     text-align: left;
     word-break: break-word;
     padding: 4px 8px;
     background-color: var(--brand-color, black);
+  }
+
+  .msla-subgraph-title-text {
+    font: 14px/20px normal @semibold-font-family;
+  }
+  .msla-panel-card-error-wrapper {
+    width: 100%;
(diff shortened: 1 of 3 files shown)


libs/designer-ui/src/lib/card/subgraphCard/subgraphCard.less
```diff
@@ -32,11 +32,38 @@
     color: white;
     flex-grow: 1;
     align-self: center;
-    display: inline-block;
-    font: 14px/20px normal @semibold-font-family;
```
rllyy97, line 36: It looks like you removed the `font` prop, was it causing issues?
From your screenshot the large sized node looks fine without it but the small sized node is incorrect

valentina-vallalta: oh, good catch! I removed it there and added it to a new class because it was making the error message appear bold and I didnt notice it affected the small node, let me fix that

hartra344 (approved):

PR #4321, opened 2024-03-07
feat(vscode): Introducing iac to vscode extension

ccastrotrejo:
### Main Code Changes
- Introduce `Generate deployment scripts` command.
- Set `Create new code project` command to MacOs, this will also allow to create workspaces 
- A new class `NewCodeProjectTypeStep` is added, which is responsible for setting up the project based on the user's choices. It creates the necessary directories, sets up the sub-wizard based on the context, and prompts the user for project information.
- The `OpenFolderStepCodeProject` class is updated to check if a `.code-workspace` file exists in the project path and opens it if it does. 
- A new class `SetLogicAppName` is added, which prompts the user to enter a name for the Logic App project.
-  The `setWorkspaceName` class is updated to save the workspace path and ensure the directory is created. It also sets the `workspaceFolder` based on the `workspacePath`. 
- A new file was added that defines the promptParameterizeConnections and parameterizeConnections functions, which handle the logic for parameterizing connections in the project.

### Screenshots

generate-deployment-script

parameterize-connection-deployment
parameterize-connection
parameterize-startup


ccastrotrejo (commented):
Self reviewed

Issue #4322, opened 2024-03-07
The secure string is missing on new designer parameter panel


Xiaoyu-Huang:
### Describe the Bug with repro steps 

In new logic app designer parameter panel, the secure string is missing.

![image](<https://github.com/Azure/LogicAppsUX/assets/22843713/c9bccced-db45-4bc8-8bf6-9397d002155d>)

Old designer parameter panel view is like below:

![image](<https://github.com/Azure/LogicAppsUX/assets/22843713/9a628917-9ad3-44b3-905e-11a493062a58>)

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Edge

### Additional context 

*No response*

PR #4325, opened 2024-03-07
fix(Designer): Showing secure workflow parameters in consumption

rllyy97:
## Main Changes

Fixed an issue that was preventing secure workflow parameter types from appearing as options in designer.
Can now select and properly render `SecureString` and `SecureObject`
I also removed the secure string warning on the value fields since it should only appear on the default value.

(Also fixed a small unrelated issue where we didn't run a string through INTL in the delete modal)

Fixes https://github.com/Azure/LogicAppsUX/issues/4322

image
image


hartra344:
Can you make sure this yellow warning meets a11y contrast standards?

takyyon (approved):

PR #4354, opened 2024-03-12
fix(Designer): Only resolve app settings if they are used in resource ids

rllyy97:
## Main Changes

Only resolve app settings if they are used in resource ids.

This allows users to have key vault references in app settings / connections data without causing the initialization issues we were seeing when subscriptions had app settings values.
The only bad behavior with this change is when a user has a key vault reference as an app setting that is used in place of a subscription ID, I don't see a way around that one.

I've also updated a couple situations where we weren't parsing errors properly and returning generic errors on accident.


hartra344 (commented):
Instead of just a subscription ID, maybe a better way to do this would be to resolve it if it’s any part of the resource ID. Realistically no part of the resource ID should be coming from key vault. It’s not a secret what’s happening here it’s like the primary cause of the issues.

valentina-vallalta (approved):

PR #4355, opened 2024-03-12
fix(vscode): Implement function to determine if connections have been already parameterized

ccastrotrejo:
This pull request primarily focuses on enhancing the parameterization of connections in the VS Code Designer application. The main changes include the addition of a function to check if connections are already parameterized, and the subsequent use of this function in the `parameterizeConnections` command. 

* A new function `isConnectionsParameterized` is added. This function checks if the connections data is already parameterized by examining various connection types and their properties. If any connection is found to be parameterized, the function returns `true`.

* Before parameterizing connections, the `parameterizeConnections` command now checks if the connections are already parameterized using the `isConnectionsParameterized` function. If they are, it shows an information message and returns without making any changes.

### Screenshots
Screenshot 2024-03-12 at 11 37 28 AM


ccastrotrejo (commented):
Self-reviewed

PR #4374, opened 2024-03-15
fix(Designer): "Internal" swagger operation parameters are now initialized properly

rllyy97:
## Main Changes

We had a bug where swagger parameters labeled as "internal" were not getting initialized.
In our schema / parameter processors we have some flags to exclude "internal" properties, and in one place we were not respecting this flag.
This has now been fixed and I'm seeing "internal" swagger parameters get serialized properly.

The below "body" parameters were marked as internal in this operation.

image

### DIFF
--- a/libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
+++ b/libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
@@ -258,7 +258,9 @@ export class SchemaProcessor {
     titlePrefix = this._concatenateString(titlePrefix, schema.title);
     summaryPrefix = this._concatenateString(summaryPrefix, summary);
 
-    if (keys.length && !this._isInternalParameter(schema) && !isReadOnlyInputParameter) {
+    const skipInternal = this.options.excludeInternal && this._isInternalParameter(schema);
+
+    if (keys.length && !skipInternal && !isReadOnlyInputParameter) {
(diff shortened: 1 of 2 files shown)


ccastrotrejo (approved):

libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
```diff
@@ -258,7 +258,9 @@ export class SchemaProcessor {
     titlePrefix = this._concatenateString(titlePrefix, schema.title);
     summaryPrefix = this._concatenateString(summaryPrefix, summary);
 
-    if (keys.length && !this._isInternalParameter(schema) && !isReadOnlyInputParameter) {
+    const skipInternal = this.options.excludeInternal && this._isInternalParameter(schema);
+
+    if (keys.length && !skipInternal && !isReadOnlyInputParameter) {
```
preetriti1, line 263: This isnt right. Now for the internal parameters (legacy) defined in swaggers, any parameter marked as internal should always have a default value.
This function is expanding object properties, if the object has properties it cannot be marked as internal because we should not be parsing it.

While serializing we would be just using the default value to put in the definition for any property marked as internal regardless of its swagger type.

preetriti1 (changes requested):

rllyy97:
These changes are not being merged in.

Issue #4350, opened 2024-03-12
Paste text on logic app standard on Firefox will have duplicate text but no issue on Edge or Chrome


jiecao-574:
### Describe the Bug with repro steps 

**Issue:**

 On Firefox, copy a text and paste on logic app standard ‘Run query and list results’ action Query body, it will paste twice - duplicate texts.

We can see it works well on Edge and Chrome.

Is it a logic app code bug?

![image](<https://github.com/Azure/LogicAppsUX/assets/62946066/bb6b1cbf-f5b7-4ee5-9d71-247111573144>)

### What type of Logic App Is this happening in? 

Standard (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

![image](<https://github.com/Azure/LogicAppsUX/assets/62946066/2d89f0bb-bb50-49c4-98b6-bd6d42b0a159>)

### Browser 

Firefox

### Additional context 

*No response*

PR #4378, opened 2024-03-16
fix(Designer): Override default Paste Behavior on FireFox

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/4350


hartra344 (approved):
Why is Firefox starting to feel like IE? >.<

PR #4408, opened 2024-03-21
fix(Designer): Secure string workflow parameters now pass validation

rllyy97:
## Main Changes

Secure string workflow parameters now pass validation.
Previously our value / default value checking was verifying the values were objects, they are now evaluated just as-is like strings.

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -220,17 +220,19 @@ const getWorkflowParameters = (
     delete parameterDefinition['name'];
     delete parameterDefinition['isEditable'];
 
-    const valueIsString =
+    const isStringParameter =
       equals(parameterDefinition.type, UIConstants.WORKFLOW_PARAMETER_TYPE.STRING) ||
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -220,15 +220,17 @@ const getWorkflowParameters = (
...
-      : typeof value !== 'string'
-      ? value
-      : JSON.parse(value);
+    const valueIsString =
+      equals(parameterDefinition.type, UIConstants.WORKFLOW_PARAMETER_TYPE.STRING) ||
+      equals(parameterDefinition.type, UIConstants.WORKFLOW_PARAMETER_TYPE.SECURE_STRING);
+
+    parameterDefinition.value = valueIsString ? value : value === '' ? undefined : typeof value !== 'string' ? value : JSON.parse(value);
+
+    const defaultValueIsString =
```
ccastrotrejo, line 229: NIT: Isn't this the same as `valueIsString`?

ccastrotrejo (approved):

Eric-B-Wu (approved):

PR #4419, opened 2024-03-22
fix(Designer): Create connection now uses `testRequests` when available

rllyy97:
## Main Changes

Implemented `testRequest` support when creating connections.
When both `testLink` and `testRequest` options are available, we prioritize `testRequest`.
// This is partially due to this being implemented in order to mitigate the current SendGrid issue


libs/logic-apps-shared/src/utils/src/lib/models/connection.ts
```diff
@@ -57,6 +58,13 @@ export interface ConnectionProperties {
...
 export interface TestLink {
+  body?: string;
+  method: string;
+  requestUri: string;
+}
+
+export interface TestRequest {
+  body?: string;
   method: string;
   requestUri: string;
```
takyyon, line 69: TestLink and TestRequest are exact similar structures right? Do we really need a copy or can we just re-use one?

takyyon (approved):

rllyy97:
I am doing some more testing on this before merging

PR #4423, opened 2024-03-23
chore(all): Move formatter from Prettier to Biome and Speed up eslint

hartra344:
Moving Formatter to Biome which is much faster, also optimizing what we eslint on commit. Overall this leads to a MUCH faster commit:

https://github.com/Azure/LogicAppsUX/assets/13208452/4db11b52-3990-4570-a2fa-88f6ef4fa331

I think we could also adopt Biomes linter and import sorter but will save that for later since it would be big file writes. The formatter is almost 1:1 compatible with prettier output except for better tertiary handling.


hartra344:
I also realized we were compiling our loc files to the wrong location. So that's fixed too.

takyyon (approved):

ematipico:
Hi, Biome maintainer here, thank you choosing Biome!

Out of curiosity, do you know how much time will you save using Biome? Either locally or in CI?

hartra344:
Hey @ematipico 

Won't save any in CI yet since we're only using the formatter, it'll save us a few seconds on each commit though since we run it as a commit hook.

I'm pretty excited though about what I'm seeing from y'all in the linter space though. The only real hold back for me switching us to that is that we rely on a few eslint plugins. Most importantly the one for formatjs. But that would  come with signifigant time savings locally and in CI. I would also imagine other savings such as battery life on our laptops.

Issue #4386, opened 2024-03-18
SFTP-SSH Azure connector "File path" dynamic content field has the wrong code path


kaly-d:
### Describe the Bug with repro steps 

1. In the new designer, open a Logic App Standard workflow with the Azure SFTP-SSH connector as the trigger (ex: "When a file is added or modified") 
2. Add an Azure SFTP-SSH action, like "Delete file", and in the "File" field, select the dynamic content as "File path" 
3. The code path for that dynamic content is "triggerOutputs()['x-ms-file-path-encoded']", but this causes failures as the path is wrong. 
4. The "x-ms-file-path-encoded" is inside the "headers" tag, so we need to update the Code View to look like this: "triggerOutputs()['headers']['x-ms-file-path-encoded']" 
5. The issue does not reproduce in the old designer. 

### What type of Logic App Is this happening in? 

Standard (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "Delete_file": {
                "inputs": {
                    "headers": {
                        "SkipDeleteIfFileNotFoundOnServer": false
                    },
                    "host": {
                        "connection": {
                            "referenceName": "sftpwithssh-3"
                        }
                    },
                    "method": "delete",
                    "path": "/datasets/default/files/@{encodeURIComponent(encodeURIComponent(base64ToString(triggerOutputs()['x-ms-file-path-encoded'])))}"
                },
                "runAfter": {},
                "type": "ApiConnection"
            }
        },
        "contentVersion": "1.0.0.0",
        "outputs": {},
        "triggers": {
            "When_a_file_is_added_or_modified": {
                "inputs": {
                    "host": {
                        "connection": {
                            "referenceName": "sftpwithssh-2"
                        }
                    },
                    "method": "get",
                    "path": "/datasets/default/triggers/onupdatedfile",
                    "queries": {
                        "folderId": "L2hvbWUvc2Z0cHVzZXI=",
                        "includeFileContent": false,
                        "inferContentType": true,
                        "queryParametersSingleEncoded": true
                    }
                },
                "metadata": {
[...]

PR #4454, opened 2024-03-26
fix(designer): Trigger Tokens Having Wrong Titles and Values With Output Source

Eric-B-Wu:
There was an issue with getting the right token title and value with a given output source
Fixes https://github.com/Azure/LogicAppsUX/issues/4386

Before:
sftpsshTokenBefore

After:
sftpsshTokenAfter


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1027,17 +1027,32 @@ export function getExpressionValueForOutputToken(token: OutputToken, nodeType: s
...
     return actionName ? `${OutputSource.Body}(${convertToStringLiteral(actionName)})` : constants.TRIGGER_BODY_OUTPUT;
   } else {
-    return actionName ? `${constants.OUTPUTS}(${convertToStringLiteral(actionName)})` : constants.TRIGGER_OUTPUTS_OUTPUT;
+    if (actionName) {
+      return `${OutputSource.Outputs}(${convertToStringLiteral(actionName)})`;
+    }
+    if (source) {
+      const propertyInQueries = !!source && equals(source, OutputSource.Queries);
+      const propertyInHeaders = !!source && equals(source, OutputSource.Headers);
+      const propertyInStatusCode = !!source && equals(source, OutputSource.StatusCode);
```
rllyy97, line 1046: Is there a reason to have the `!!source` on each line if we're checking it in the if statement to get into this block?

rllyy97 (approved):

ccastrotrejo (approved):

libs/designer/src/lib/core/utils/parameters/helper.ts
Eric-B-Wu: whoops added the if block after, will fix this

PR #4473, opened 2024-03-27
fix(powerautomate): Serialize combobox value as json if value is an object

Ram-G:

Bugfix

In cases where `selectedValue` is an `object`, doing `toString` on it causes it to set `value` as `[object Object]`.

The workaround is to check if the `value` being serialized is an object, and if so, call `JSON.stringify` instead.

No

Before:
image

After:
image


Ram-G:
I think this issue is fixed in the latest release...

hartra344 (approved):

PR #4483, opened 2024-03-28
feat(Designer): Dynamic data performance improvements

rllyy97:
## Main Changes

Overall reduced the number of dispatches run due to dynamic data.
- Combined dynamic data clear actions into one per action, removed multiple unnecessary runs
- Moved `updateExistingInputTokenTitles` to `extraReducers` to avoid extra state update, runs once per action
- Moved dynamicSchema fetch to ReactQuery so we don't duplicate the same fetches for each similar action
- Changed dynamic parameter updates during initialization to run once per action instead of once per parameter (in my test case with 4 parameters per action it ran the flow 4x less) (crazy math I know)

There's still more work to do here but I don't want to cram too much into one PR.
The main next big step is to combine the "once per action" dispatches into one large dispatch.  This is not always an option but it's worth investigating more.  These are the "once per action" dynamic actions left to cut down on:

image

---

Also added little performance tool helper thingy just for development (along with a `dev` redux slice)
The redux dev extension really likes to crash under load I've found out so I'm doing some things in this component instead

image


libs/logic-apps-shared/src/utils/src/lib/helpers/functions.ts
```diff
@@ -1037,6 +1037,25 @@ export const filterRecord = <T>(data: Record<string, T>, filter: (_key: string,
...
 };
 
+/**
+ * Returns a sorted record object.
+ * @param data - The record to sort.
+ * @param sort - The sort function.
+ * @returns {Record<string, T>} - The sorted record.
+ */
+
+export const sortRecord = <T>(
```
ccastrotrejo, line 1047: NIT: should we start adding tests whenever we add reusable functions in this file?

ccastrotrejo (approved):

PR #4493, opened 2024-03-29
fix(Designer): Collapsed view of 'Run After' 

### LINKED ISSUE #4370: Node run after component overflow

When there is an action that has a large number of run afters, we currently are just rendering all of the traffic light components in a row.

image

In legacy designer we just had a small tooltip pop up, we should do something similar (not necessarily the same)

image

Standard (Portal)

New Designer

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Any

### Additional context 

*No response*

valentina-vallalta:
This PR addresses the display issue for the 'Run After' dots when their count is 6 or more. Previously, all dots were displayed in a row, which led to visibility issues when there is a large number of them. The updated design introduces a collapsed view, improving readability and user experience. The tooltip design is similar to old designer, with a small pill displaying the number of Run After components there are, and, on hover, displaying all the information in one scrollable panel. Similar to current behavior, an action that only has 'Is Successful' status selected will not be displayed in the tooltip content or included in the count.

Before:

After:

fixes #4370


hartra344 (approved):

libs/designer/src/lib/ui/connections/runAfterIndicator/styles.less
```diff
@@ -26,6 +26,23 @@
   display: flex;
 }
 
+.msla-run-after-overflow {
+  box-sizing: border-box;
+  font-family: 'Segoe UI Semibold', sans-serif;
+  font-size: 10px;
+  line-height: 1.4;
+  background-color: #d3d3d3;
```
ccastrotrejo, line 34: Nit: are you takin in consideration dark theme/mode?

ccastrotrejo (approved):

libs/designer/src/lib/ui/connections/runAfterIndicator/styles.less
valentina-vallalta: good point! Added code for that

ccastrotrejo (approved):

rllyy97:
Talked with Valentina offline, we found a couple bugs related to run afters only having the success status

ccastrotrejo (approved):

PR #4498, opened 2024-03-29
feat(vscode): Show error message in overview when workflow runtime api isn't running

ccastrotrejo:
The most significant changes include the addition of a new error message in the localization strings, modifications to the `openOverview` function in the VS Code Designer application to check if the workflow runtime is running, and changes in the React application to display the new error message when the workflow runtime is not running.

* Added a new error message string for debugging a Logic App project. 

* Added a new variable `isWorkflowRuntimeRunning` to check if the workflow runtime is running. This variable is set based on whether the `ext.workflowRuntimePort` is null or undefined, and is passed to the web view.

* Added a new variable `intlText` to store the new error message string. This message is displayed when the workflow runtime is not running. Also, the `useInfiniteQuery` is enabled only when the workflow runtime is running.

* Added a new state variable `isWorkflowRuntimeRunning` to the `WorkflowState` and `InitializePayload` interfaces. This variable is set based on the payload of the `initializeWorkflow` action.

### Screenshot
Screenshot 2024-03-29 at 3 53 32 PM


apps/vs-code-react/src/app/overview/app.tsx
```diff
@@ -11,13 +11,25 @@ import { useCallback, useContext, useMemo } from 'react';
...
 }
 export const OverviewApp = () => {
   const workflowState = useSelector((state: RootState) => state.workflow);
   const vscode = useContext(VSCodeContext);
+  const { isWorkflowRuntimeRunning } = workflowState;
+  const intl = useIntl();
+
+  const intlText = {
+    DEBUG_PROJECT_ERROR: intl.formatMessage({
+      defaultMessage: 'Please start the project by pressing F5 or run it through the Run and Debug view',
```
ccastrotrejo, line 28: @ecfan could you please review this error message?

ecfan: @ccastrotrejo: Apologies, am on vacation until April 8.

ccastrotrejo: No problem at all, I can go back later and change the wording. Thanks!

Eric-B-Wu (approved):

Issue #4481, opened 2024-03-28
Problem with the Count parameter of the Until action


xuehongg:
### Describe the Bug with repro steps 

1. Create a standard logic app and create a workflow in it 
2. Add the Until loop to it and you can see its Count parameter has a default value of 60 
3. Save the workflow, check code view and designer view, the Count parameter is good with value 60 
4. Now In designer change the Count parameter to some other value like 300, save workflow, check code view and designer view, the Count parameter is good with value 300 
5. Now in designer, delete the Count parameter, save workflow, the code view and designer view look good with no Counter parameter. 

![image](<https://github.com/Azure/LogicAppsUX/assets/49498706/7479cdaa-f60d-4f4d-a553-f80e1d665d6c>)

6. Now add some action like Compose to the workflow, save it, Check code view and designer view , you will find the Counter parameter is back with its default value of 60. 

The old designer does not have this issue, because in old designer, the Counter parameter cannot be deleted.

### What type of Logic App Is this happening in? 

Standard (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Chrome

### Additional context 

ICM 486728741

AyanaKobashikawa:
@Eric-B-Wu 
I'm sorry to bother you.
Please confirm this problem as soon as possible, as it is keeping our clients waiting.

Eric-B-Wu:
Apologies for the delay, we've identified a code bug and will be hotfixing this issue today, we expect this to complete EOD today to all regions.

AyanaKobashikawa:
@Eric-B-Wu 
Before reporting to the customer, I have some additional questions on this issue.
 
Is the Count parameter an optional parameter?
If the Count parameter is deleted, is my understanding correct that the default value will not be set unless the user subsequently modifies the Count parameter with the value?
Is my understanding correct that the loop count of the Until action will exceed the default limit If the Count parameter is deleted (initially 60 loops)?
If the Count paramete been deleted, is it correct that the number of loops in the Until action will exceed 5000?

Eric-B-Wu:
> Apologies for the delay, we've identified a code bug and will be hotfixing this issue today, we expect this to complete EOD today to all regions.

[...]

PR #4505, opened 2024-04-01
fix(designer): Only use default value when the parameter is required

Eric-B-Wu:
On optional parameters, we should not use the default value, as there are times where the customer wants to keep the value unset.

Fixes https://github.com/Azure/LogicAppsUX/issues/4481

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/helper.ts
+++ b/libs/designer/src/lib/core/utils/parameters/helper.ts
@@ -871,7 +871,10 @@ export function loadParameterValue(parameter: InputParameter): ValueSegment[] {
   if (parameter.isNotificationUrl) {
     valueObject = `@${constants.HTTP_WEBHOOK_LIST_CALLBACK_URL_NAME}`;
   } else {
-    valueObject = parameter?.value ?? parameter?.default;
+    valueObject = parameter?.value;
+    if (parameter?.required && valueObject === undefined) {
+      valueObject = parameter?.default;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


hartra344 (approved):

takyyon (approved):

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -871,7 +871,10 @@ export function loadParameterValue(parameter: InputParameter): ValueSegment[] {
   if (parameter.isNotificationUrl) {
     valueObject = `@${constants.HTTP_WEBHOOK_LIST_CALLBACK_URL_NAME}`;
   } else {
-    valueObject = parameter?.value ?? parameter?.default;
+    valueObject = parameter?.value;
+    if (parameter?.required && valueObject === undefined) {
```
ek68794998, line 875: Curious if you have any examples of this? Just curious since I know we've had feedback from people who expect defaults and I don't recall if the parameters they complained about were optional or not.

Eric-B-Wu: One such example of this is: https://github.com/Azure/LogicAppsUX/issues/4481
Here we have the optional parameter of `Count` which when the user deletes, we end up repopulating it with a default value of 60.

Eric-B-Wu: Discussed offline with Eric, was a miscommunication we're still using defaults on creation/adding action, just if the user explicitly deletes the optional parameter do we not use the default parameter

Issue #4492, opened 2024-03-29
Switch case doesn't accept integer values and when forced to use a string, it restores the integer value after a save


mohamedaymenkarmous:
### Describe the Bug with repro steps 

1. Create a logic app 
2. Add an "Initialize variable". Set its type "String". Call it "test" as an example. 
3. After it, add another step "Set variable". Choose the "test" variable and add a value (as an example of a numerical value): 2 
4. Add a Switch (control) action. Set its input to the "test" variable: `@{variables('test')}`. 
5. Add a "Case" branch and set its equals value to: 2 
6. Add anything inside the first branch and anything in the default branch just to test where the execution will go. I have initialized and set another variable in the beginning called "hello" and I changed its value in the first branch of the switch and in the default branch of the Switch 
7. Save the logic app and run the test 

When you run the test you'll get an error inside the Switch where the "set variable" I have created inside the 2 branches will turn to red with the following error: `The execution of template action 'Set_variable_1' failed: an unexpected exception encountered when evaluating branching condition.`.

When you re-run these steps but you change in the step number 4 the variable to a function to force the input as a string like this `@{string(variables('test'))}` the error will not disappear because the issue is with the case values that considers the input provided to them as an integer in this example.

The error was no longer present when I changed the Case value from `2` to `"2"`. So, I think the logic app cannot compare a string with an integer and I was totally fine with that but once I save the workflow, the case value changes from `"2"` to `2` which is not practical for someone who has multiple switch cases where he wants to edit something in the workflow he have to make sure to add the double quotes that were deleted. In the old UI this was never an issue because the `2` value was never generating an error. But when I changed to the new UI, when I changed something not related with the switch case, I started getting this error and I was confused because I didn't touch the switch case. And after all this investigation I discovered that the issue is with the case conditions that consider any integer value inside them as an integer not as a dynamic value "that is a string if the switch value is a string and that is an integer value if the switch value is an integer".

I'm not asking for too much like to make the case comparison to use a dynamic value (this solution is the best solution). But I'm just asking to at least correct the "Case" condition to avoid deleting the double quotes since I don't want to add them every time I edit the logic app

Thank you for all your hard work.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

[...]

PR #4568, opened 2024-04-06
fix(designer): Stringified Integers are not handled in switch cases and are casted to integers

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/4492
Before:
before

After
after


Eric-B-Wu:
converting to draft as I look to see if there's a better way of handling this, the logic added here is very Switch/Case-specific, which I would like to avoid

Eric-B-Wu:
closing this issue in favor of https://github.com/Azure/LogicAppsUX/pull/4588

PR #4583, opened 2024-04-09
fix(designer): Support extracting dynamic schema when schema path contains empty properties

preetriti1:

Bug fix

If schema path after splitting on '/' contains empty properties, schema extracted would be null, fixing that to ignore the empty segments. This will help in extracting root schema from dynamic schema response by connectors only specifying '/' in value-path

No


libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -120,13 +120,8 @@ export const getLegacyDynamicSchema = async (
...
   }
 
-  const schemaPath = extension['value-path'] ? extension['value-path'].split('/') : undefined;
-  return schemaPath
-    ? getObjectPropertyValue(
-        response,
-        schemaPath.length && equals(schemaPath[schemaPath.length - 1], 'properties') ? schemaPath.splice(-1, 1) : schemaPath
-      ) ?? null
-    : { properties: response, type: Types.Object };
+  const schemaPath = extension['value-path'] ? extension['value-path'].split('/').filter((s) => s) : undefined;
```
ccastrotrejo, line 123: NIT: Should the filter be more specific? e.g. its null or undefined or !isEmptyString

preetriti1: its a split on string, so it will always be empty string not a number or object or array, hence just used simple notation for non-empty

ccastrotrejo (approved):

PR #4592, opened 2024-04-10
feat(Designer): Add Custom Editor for Vs-Code To Display DM Editor

Elaina-Lee:

Introduces bahavior of dealing with displaytext editor for dataMapper connector
doesn't display navigating to data mapper on the panel
will display navigating to data mapper message

Using http (local) connector for showing changes, will show this on data mapper with changed manifest on back end
image

https://github.com/Azure/LogicAppsUX/assets/144840522/f7618e1a-1c8d-46e0-b858-72a7775554d1

Reliant on the manifest change on back-end. This editor (to be confirmed with back-end) is to be added to the data mapper connector afterwards.
`
        text: {
          type: 'string',
          title: 'Info',
          description: 'Data mapper navigate description',
          'x-ms-editor': 'displaytext',
          'x-ms-editor-options': {
            hideLabel: true,
            visibility: 'custom',
            displayText: {
              text: 'To complete this data transformation, add an existing map or create a new map with',
              relativeLinkText: 'Data Mapper in VSCode',
              relativeLink: '/dataMapper',
            },
          },
        },
`


hartra344:
Make sure this only happens in VSCode though since this would be broken in portal

Eric-B-Wu:
Instead of adding a whole new displayText property in the manifest just for this editor, I think it would be better to add it in the x-ms-editor-options, and build a viewmodel for this new editor

Elaina-Lee:
> Make sure this only happens in VSCode though since this would be broken in portal

Will do! As written in the PR description, still working on the flag part. I had to leave it as draft until I'm able to test vscode.

Elaina-Lee:
> Instead of adding a whole new displayText property in the manifest just for this editor, I think it would be better to add it in the x-ms-editor-options, and build a viewmodel for this new editor

Thanks for the suggestion. I will reach out offline - I may need more clarification.

libs/designer-ui/src/lib/displayText/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { Link } from '@fluentui/react';
+
+export interface DisplayTextEditorProps {
+  displayTextText: string;
```
Eric-B-Wu, line 4: nit: can we just change this to text? Given that this is already in the displayText Editor, we don't need to let the user know this.

libs/designer/src/lib/ui/settings/settingsection.tsx
```diff
@@ -301,9 +303,11 @@ const Setting = ({ id, settings, isReadOnly }: { id?: string; settings: Settings
           return (
             <SettingTokenField
               {...{
+                useDisplaytextEditor: useDisplaytextEditor,
```
Eric-B-Wu, line 306: are we able to up this a layer to parametersTab? We don't want to use node-specific logic on this layer

libs/designer-ui/src/lib/displayText/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+}): JSX.Element => {
+  return (
+    <div className="msla-displayText-editor-container">
+      {displayTextText ?? ''}
+      <Link className="msla-workflow-parameters-link-text" onClick={() => openRelativeLink?.(relativeLink ?? '')}>
+        {relativeLinkText}
+      </Link>
+    </div>
+  );
+};
```
Eric-B-Wu, line 24: This is quite specific logic on what this editor does, and I'm wondering how extensible it is. Namely, it will only has the option to show some text and follow it with a link. Although I wouldn't say its blocking for this PR, I think a more relevant editor would to create a markdown editor, where we let this read-only editor be something that the user can customize through the manifest

libs/designer-ui/src/lib/displayText/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const DisplayTextEditor: React.FC<DisplayTextEditorProps> = ({
+  displayTextText,
+  relativeLinkText,
+  relativeLink,
+  openRelativeLink,
+}): JSX.Element => {
+  return (
+    <div className="msla-displayText-editor-container">
+      {displayTextText ?? ''}
```
ccastrotrejo, line 18: NIT: is the `displayTextText` ever going to be null or undefined? What I mean is that the `nullish` operator (??) is never going to fall to the empty string case, since according to props is going to be there always

apps/Standalone/src/desginer/state/workflowLoadingSlice.ts
```diff
@@ -27,6 +27,7 @@ export interface WorkflowLoadingState {
   hostOptions: {
     displayRuntimeInfo: boolean; // show info about where the action is run(i.e. InApp/Shared/Custom)
     forceEnableSplitOn?: boolean; // force enable split on for all actions
+    useDisplaytextEditor?: boolean; // show displaytext editor
```
Eric-B-Wu, line 30: Instead of choosing to use vs not use the editor, can we make this host specific to DM connector? The token editors are supposed to be extendable to the connector owners, and if we're limiting it on the host-level, then we could have just used the custom editor.

libs/designer-ui/src/lib/displayText/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Link className="msla-workflow-parameters-link-text" onClick={() => openRelativeLink?.(relativeLink ?? '')}>
```
ccastrotrejo, line 19: NIT: I think that despite the fact that this `msla-workflow-parameters-link-text` class works for this component, maybe a good idea would be to not use it because the DisplayTextEditor can be use not only in parameters tab

takyyon (approved):

libs/designer-ui/src/lib/displayText/index.tsx
Elaina-Lee: I had to move away from the Markdown Editor because I need to use a specific logic for vscode extension. I found out href does not work in vscode when it's relevant path, so I have to use callback for this link OnClick. I wasn't able to make Markdown to do this, but I'm open to suggestions.

ccastrotrejo: Yes, Elaina and I discussed this.  I think overall the idea of having a Markdown component that can be render based on the manifest is quite good. However, in this specific scenario in which we want to render links is different.
For web browsers and web pages in general it works fine. The browser will identify the markdown link and open the page in a new tab.
For vscode extension it will do the same thing, open a tab in the web browser, however, that's not the desired behavior. What is needed in vscode is to open a new vscode tab and that requires a callback function. Also, despite that we are using webviews that doesn't mean that it's an actual web browser being rendered inside vscode. 
I can explain this further if anyone else is interested

libs/designer-ui/src/lib/displayText/index.tsx
```diff
@@ -14,11 +14,9 @@ export const DisplayTextEditor: React.FC<DisplayTextEditorProps> = ({
   openRelativeLink,
 }): JSX.Element => {
   return (
-    <div className="msla-displayText-editor-container">
+    <div>
```
ccastrotrejo, line 17: nit: Use <></> unless you need the `display: block` from it

libs/designer-ui/src/lib/displayText/index.tsx
Eric-B-Wu: yeah Im thinking for this here we can have some placeholder text here to encorporate the action needed in the vscode command

Eric-B-Wu: that being said, I'm starting to like more and more the idea of using a custom editor for this case and moving it outside of the designer?

apps/vs-code-react/src/app/designer/customEditorService.tsx
```diff
@@ -0,0 +1,52 @@
...
+  areCustomEditorsEnabled?: boolean;
+  openRelativeLink?: (relativeLink: string) => void;
+}
+
+export class CustomEditorService implements IEditorService {
+  _areCustomEditorsEnabled = true;
+
+  constructor(public readonly options: CustomEditorServiceOptions) {
+    const { areCustomEditorsEnabled } = options;
+    this._areCustomEditorsEnabled = areCustomEditorsEnabled || true;
```
ccastrotrejo, line 14: NIT: this is always going to be true, not matter what you send.  I wonder if the CustomEditorsEnabled is needed

apps/vs-code-react/src/app/designer/customEditorService.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public getEditor = (props: IEditorParameterInfo) => {
+    const { operationInfo, parameter } = props;
+    const { connectorId, operationId } = operationInfo ?? {};
+    const { parameterName, editor, editorOptions } = parameter ?? {};
+
+    if (!this._areCustomEditorsEnabled) {
```
ccastrotrejo, line 22: Nit: I guess its better to return it at the top as a first instance

libs/designer-ui/src/lib/settings/settingsection/customTokenField.tsx
```diff
@@ -51,13 +51,15 @@ const customEditorName = 'internal-custom-editor';
...
   const { editor, editorOptions } = props;
+
   return (
-    equals(editor, customEditorName) &&
-    typeof editorOptions === 'object' &&
-    !!editorOptions &&
-    typeof (editorOptions as { EditorComponent: unknown }).EditorComponent === 'function'
+    editorOptions?.visibility === 'custom' ||
+    (equals(editor, customEditorName) &&
+      typeof editorOptions === 'object' &&
```
ccastrotrejo, line 60: Nit: there is an isObject() function

ccastrotrejo:
I wonder, if we are treat it as a service, does it need to be initiliazed in the designer side?

apps/vs-code-react/src/app/designer/customEditorService.tsx
Elaina-Lee: I'm following the format in existing customEditorService in LocalDesigner, would this be unnecessary?

Elaina-Lee:
> I wonder, if we are treat it as a service, does it need to be initiliazed in the designer side?



> I wonder, if we are treat it as a service, does it need to be initiliazed in the designer side?



> I wonder, if we are treat it as a service, does it need to be initiliazed in the designer side?

This shouldn't be needed to be initialized in the designer side. CustomEditorService already exists on local standalone (apps\Standalone\src\desginer\app\LocalDesigner\customEditorService.tsx), and I'm following the same way in the vs-code.

apps/vs-code-react/src/app/designer/customEditorService.tsx
ccastrotrejo: I noticed that you changed it to `??` which makes more sense and it works fine.
What I meant with the original comments was that 
`this._areCustomEditorsEnabled = false || true;` its equals to true.
`this._areCustomEditorsEnabled = false ?? true;` its equals to false.

ccastrotrejo (approved):

PR #4620, opened 2024-04-15
fix(vscode): Add eslint configuration to vs-code-designer and vs-code-react

ccastrotrejo:
This pull request primarily focuses on introducing ESLint configuration files and removing ESLint disable comments in the `vs-code-designer` and `vs-code-react` applications. The addition of the ESLint configurations will help to enforce code style and quality. The removal of the ESLint disable comments indicates that the code now adheres to the rules specified in the ESLint configurations.

* Added ESLint configuration files for both applications, extending from the root configuration and specifying rules for different file types. 

* Added a `lint` script to the `package.json` to facilitate running ESLint on the codebase.
* Removed ESLint disable comments, indicating that the code now adheres to the rules specified in the ESLint configurations.

### DIFF
--- a/apps/vs-code-designer/.eslintrc.json
+++ b/apps/vs-code-designer/.eslintrc.json
@@ -24,7 +24,9 @@
                 "*.ts",
                 "*.tsx"
             ],
-            "rules": {}
+            "rules": {
+                "react-hooks/rules-of-hooks": "off"
+            }
         },
         {
             "files": [
--- a/apps/vs-code-designer/.eslintrc.json
+++ b/apps/vs-code-designer/.eslintrc.json
@@ -1,0 +1,37 @@
+{
+    "extends": [
+        "../../.eslintrc.json"
+    ],
+    "ignorePatterns": [
+        "!**/*",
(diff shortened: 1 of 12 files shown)


hartra344 (approved):

apps/vs-code-designer/.eslintrc.json
```diff
@@ -0,0 +1,39 @@
...
+            ],
+            "rules": {}
+        },
+        {
+            "files": [
+                "*.ts",
+                "*.tsx"
+            ],
+            "rules": {
+                "react-hooks/rules-of-hooks": "off"
```
hartra344, line 28: I wouldn't advise ever turning this off, but disable it where it's not needed instead

ccastrotrejo: The `vs-code-designer` doesn't host any React code. It's just giving errors in a function that is called `useBinariesDependencies`. Eslint thinks that its a hook and its not in a react component or in a .tsx file. Thats why I decided to remove the whole rule in that project

PR #4627, opened 2024-04-16
fix(designer): html editor styles not showing in v3

AfroJoe759:

Bug fix

- Current behavior is that for the new designer, the HTML editor adds <br> tags anyplace where there is new lines which results in malformed outputs (spaces where there shouldn't be) and malformed tags resulting in styles defined in a <styles> tag to not render. The malformed HTML also appears when a user reopens a flow and checks the HTML editor.

- <br> tags aren't replacing newline (\n) characters anymore, which fixes malformed HTML outputs and allows styles that aren't defined inline to be applied, which matches the bahavior of the older designer.

N/A

The bug itself contains a more detailed example and explanation of the above, but here's a simple example of the difference before and after my change.

Link to bug: https://msazure.visualstudio.com/One/_workitems/edit/27066057/

**Before Change Screenshot**
image

**After Change Screenshot**
image


hartra344:
@Eric-B-Wu can you review this one?

Eric-B-Wu:
> @Eric-B-Wu can you review this one?

yeah im just confirming on our end this doesn't break anything I think in general we wanted to keep anything in the html editor strictly in HTML-form

Localize/lang/strings.json
Eric-B-Wu, line 1: @hartra344 do you know why these localization strings were deleted? they seem to be still in use; however, Joe mentioned that this was added on commit/push

Eric-B-Wu (approved):

Localize/lang/strings.json
hartra344: I think there's something up with them getting pushed when generated on commit

AfroJoe759: @hartra344 @Eric-B-Wu do you want me to revert that specific file to what it is in the main branch?

ek68794998:
Sorry, having trouble adding comments with the correct GitHub account. Hopefully this works.

The reason this was added was to make sure that if someone entered a newline in the raw HTML editor, that it would be retained when 1) switching back to the WYSIWYG editor, 2) when saving the flow contents, and 3) if there were multiple newlines in a row.

Let’s make sure we’ve validated these cases, and check with Kishor who can hopefully share some past newline bugs with some test cases we can run this through.

PR #4629, opened 2024-04-17
chore(all): Fix issue where strings file isn't checked in on changes during precommit

hartra344:


Localize/lang/strings.json
```diff
@@ -2014,6 +2014,7 @@
   "_iXW+2l.comment": "Chatbot input start of sentence for adding an action that the user should complete. Trailing space is intentional.",
   "_id4DBb.comment": "First part of the Copilot Get Started description for Suggested Flow section",
   "_idQjOP.comment": "Label for properties tab",
+  "_idw/7j.comment": "Export logic app text.",
```
hartra344, line 2017: I made the extra period in the description just to force it to regenerate the strings.json file for testing the fix. Does not affect what customers see.

Issue #4630, opened 2024-04-17
error message when adding Microsoft Bookings trigger: The renderComponentIntoRoot component encountered an error while loading.


ZoltanK-MSFT:
### Describe the Bug with repro steps

New workflow - add new Microsoft Bookings trigger, e.g. When a appointment is Created.
Same works with the old/legacy designer.

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Screenshots or Videos

![image](https://github.com/Azure/LogicAppsUX/assets/74300019/f1ddf31c-a365-47d9-bc04-87a167d2858a)

### Browser

Edge

Eric-B-Wu:
@ZoltanK-MSFT would you be able to send netowrk logs over to wue@microsoft.com. Also do you have any existing Microsoft booking connections or are you creating a new one?

ZoltanK-MSFT:
Discussed with Eric via MS internal channel on 18th April.

Eric-B-Wu:
there is an issue where if dynamic data passes an object, we were not parsing to get dropdown options for dynamic data and thus erroring. I'll have this fixed in my next PR

Issue #4308, opened 2024-03-04
Copy/paste not working across workflows


shahparth-lab:
### Describe the Bug with repro steps

1. copy an action from a workflow 
2. try to paste it in another workflow 
3. The paste works, but only once..
4. when you try copying another action from the different workflow, it does not copy to clipboard and pastes the previously copied action instead

### What type of Logic App Is this happening in?

Standard (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Chrome

hartra344:
Should be noted from teh video, the problem is not when pasting in another workflow but when posting in a different tab. Which workflow it's being pasted to does not play a role.

Eric-B-Wu:
this is happening because we're storing the copy-paste in local storage which doesn't seem to hot update in azure portal requiring the page to be refreshed before an operation can be pasted on another tab. This fix might need to wait until we move this logic to clipboard api

hartra344:
@Eric-B-Wu I did some digging last night, it appears that localstorage isn't what we're actually using. Portal actually fakes localstorage using indexeddb and might have a caching layer in the middle. We're definitely going to need to find another way to store this in portal. The good news though is that this opens up some possibilities with some other things since indexeddb doesn't have a limit on size/space....

Issue #3386, opened 2023-09-26
Copy Paste Functionality Improvments


Eric-B-Wu:
- Use Clipboard Paste (currently uses local storage) 
- Support Scope Actions 
- Remove tokens that have dependencies that are invalid when pasting 

AB#25273917

PR #4668, opened 2024-04-22
feat(designer): Scope Copy + Paste

Eric-B-Wu:

Feature of Scope Copy Paste. https://github.com/Azure/LogicAppsUX/issues/3386
Some other changes: 
1. Copy Paste now uses clipboard API instead of local storage. There are some pros and cons with this, the main pro is that because portal uses indexeddb to emulate local storage, this should work better. Another pro is that users can actually see what is copied + paste. Con is that we can not detect changes in the copied/ pasted value in realtime, as clipboard api will not let us detect change - therefore the paste option is always active, and if tried to paste with an invalid paste object, it will just show a popup that it's not invalid. Fixes: https://github.com/Azure/LogicAppsUX/issues/4308
3. Fixes https://github.com/Azure/LogicAppsUX/issues/4630
4. Some small custom code C# changes
5. Adds a "no items" header for combobox when there are no items

Was a feature request
Feature Now implemented

Zip file cause too large, but this one shows the copy + paste experience
[scopesuccess.zip](https://github.com/Azure/LogicAppsUX/files/15064557/scopesuccess.zip)

This one shows the benefits of clipboard api
clipboardpaste


libs/designer/src/lib/core/state/staticresultschema/staticresultsSlice.ts
```diff
@@ -25,23 +25,36 @@ export const staticResultsSlice = createSlice({
     initializeStaticResultProperties: (state, action: PayloadAction<Record<string, any>>) => {
       state.properties = action.payload;
     },
+    initScopeCopiedStaticResultProperties: (state, action: PayloadAction<Record<string, any>>) => {
+      const copiedConnections = action.payload;
+      Object.entries(copiedConnections).forEach(([nodeId, staticResultProperty]) => {
+        state.properties[nodeId + 0] = staticResultProperty;
```
rllyy97, line 31: I'm confused on the `nodeId + 0` here

hartra344 (approved):

rllyy97 (approved):
Great job with this!

rllyy97:
Comment on your `therefore the paste option is always active`, can we check the clipboard object for a valid copy/paste item when opening the (+) button menu and show the paste option based on validity?

libs/designer/src/lib/core/DesignerProvider.tsx
```diff
@@ -54,7 +55,7 @@ export const DesignerProvider = ({ key, locale = 'en', options, children }: Desi
               <div data-color-scheme={themeName} className={`msla-theme-${themeName}`} style={{ height: '100vh', overflow: 'hidden' }}>
                 <ReactQueryProvider>
                   <IntlProvider locale={locale} defaultLocale={locale} onError={onError}>
-                    <ReduxReset key={key} />
+                    <ReduxReset id={id} />
```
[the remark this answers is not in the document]
hartra344: I'm not actually sure why this change is showing up on this PR, I made this change last week. It does reset redux the same way, the problem is that key shouldn't be used as a prop, it can have problems because of how it works with react already.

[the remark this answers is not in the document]
Eric-B-Wu: yep it does at least in ladesigner, it seems localdesigner discard was never implemented? all it does is reset the dirty state. also do you know what partner team uses this?

libs/designer/src/lib/core/state/staticresultschema/staticresultsSlice.ts
Eric-B-Wu: This was just how old designer stored static results, I'm not sure why. But all it was was always just stored as nodeId + 0. So if a node was like parseJson, its staticresult property is stored in parseJson0

Eric-B-Wu: Not too big of a fan of this, but I think we need to keep it like so to maintain backwards compatibility

Eric-B-Wu:
also another con to the clipboard api, is the copied action/scope gets replaced easily. Everytime the user copies another item, it gets replaced.

hartra344:
> also another con to the clipboard api, is the copied action/scope gets replaced easily. Everytime the user copies another item, it gets replaced.

I think this is fine/expected. It has pros and cons but generally windows users can use Clipboard History (Windows + V) to paste and most Mac Users use something like RayCast or Paste for clipboard history management. So in those cases, it won't overwrite the copied value but just push it down the stack.

PR #4690, opened 2024-04-24
Feat(designer): Updates to Custom Code (boilerplate, host.json, optimizations)

Eric-B-Wu:
This is adding adding a few new features/fixes as part of custom code namely:
1. Adding boiler plate code for powershell and c sharp that will be set on initialization
2. Fixing Token picker outputs for powershell and c sharp
3. Adding Host.json changes needed for powershell
4. Cleaning up some code

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     timeout-minutes: 60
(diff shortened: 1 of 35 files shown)


hartra344:
Can you add an explanation for how this is breaking? Will this require changes for PA/ICM?

PR #4694, opened 2024-04-24
fix(Designer): Remove react query in dynamic data

Eric-B-Wu:


hartra344 (approved):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
```diff
@@ -192,75 +191,25 @@ export async function getDynamicSchema(
 
       return schema ? { ...emptySchema, ...schema } : schema;
     }
-    const queryClient = getReactQueryClient();
```
preetriti1, line 195: Why are you removing queryclient? if we do this then there will be too many duplicate calls happening from workflow

preetriti1 (changes requested):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
Eric-B-Wu: this was added a couple weeks ago and it seems like we're over caching, if we want to re-add this in the future with a better querykey we can do that. However right now it seems to be causing a lot of customers issues

preetriti1 (approved):

libs/designer/src/lib/core/utils/parameters/dynamicdata.ts
rllyy97: Eric and I were discussing this, we tested both with and without react query here and it had the same number of requests, I think one of my other optimizations covered it (not sure which one)

PR #4711, opened 2024-04-25
fix(designer): Small fix to get fileExtension on custom code action add

Eric-B-Wu:
There was a bug if a user adds a custom code file with, and doesn't edit it, it'll save with boilerplate code to a file location without the file extension


apps/Standalone/src/designer/app/AzureLogicAppsDesigner/DesignerCommandBar.tsx
```diff
@@ -95,7 +95,7 @@ export const DesignerCommandBar = ({
     const customCodeFilesWithData = getCustomCodeFilesWithData(designerState.customCode);
 
     if (!hasParametersErrors) {
-      await saveWorkflow(serializedWorkflow, customCodeFilesWithData, () => dispatch(resetDesignerDirtyState()));
+      await saveWorkflow(serializedWorkflow, customCodeFilesWithData, () => dispatch(resetDesignerDirtyState(undefined)));
```
preetriti1, line 98: Please add the argument name here, it is confusing on which argument is set as undefined and why do we have to set it undefined, shouldnt the argument assume when not sent it is undefined

Eric-B-Wu: Its because the thunk was changed to be async (_: unknown, thunkAPI: any) .There isn't an actual argument we should be providing I dont think

rllyy97 (approved):

PR #4720, opened 2024-04-26
fix(designer): More Custom Code improvements

Eric-B-Wu:
1. Updating Token Picker Results based on Demo Comments
2. Updating Boilerplate Code based on Demo Comments
3. Removed customcode initialization/add connector-specifc code and changed to be based on parameter
4. Changed Initialization to be off of CodeFile location instead of NodeId

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
@@ -18,7 +18,7 @@ import { getSplitOnOptions, getUpdatedManifestForSchemaDependency, getUpdatedMan
 import {
   addRecurrenceParametersInGroup,
   getAllInputParameters,
-  getCustomCodeFileName,
+  getCustomCodeFileNameFromParameter,
   getDependentParameters,
   getInputsValueFromDefinitionForManifest,
   getParameterFromName,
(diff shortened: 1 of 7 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
```diff
@@ -473,24 +472,18 @@ export const initializeCustomCodeDataInInputs = (inputs: NodeInputs, nodeId: str
...
-  nodeInputs: NodeInputs,
-  customCode: CustomCodeFileNameMapping
-) => {
+export const updateCustomCodeInInputs = async (parameter: ParameterInfo, customCode: CustomCodeFileNameMapping) => {
   if (isNullOrEmpty(customCode)) {
     return;
   }
-  // getCustomCodeFileName does not return the file extension because the editor view model is not populated yet
-  const fileName = getCustomCodeFileName(nodeId, nodeInputs) + fileExtension;
+  const language: EditorLanguage = parameter?.editorOptions?.language;
```
takyyon, line 479: Is the language always defined or do we need some default language as well?

rllyy97 (approved):

takyyon (approved):

libs/designer/src/lib/core/actions/bjsworkflow/initialize.ts
Eric-B-Wu: language should be defined in the manifest. Before this function in an earlier function we check if customCodeParameter.editorOptions.language is one of the languages supported. And if somehow the manifest gives us an undefined language, it won't ever get to this part of the code

PR #4745, opened 2024-05-01
fix(Designer): Fix Request Body JSON Deserialization Bug

pranaydubeymicrosoft:
**Adding context in top level comment**

Adding a typecheck for output.type to ensure it's a string as expected. The current implementation permits users to provide a custom schema where output.type is an array instead of a string. This causes an issue during deserialization pictured below. This fix will block the equals comparison when output.type is not a string value.

Bug Repro:
Create a flow/logicapp in the designer using the http trigger "When an HTTP request is received" and in the schema provide something like:
```
{
    "type": "object",
    "properties": {
        "labels": {
            "type": "array",
            "items": {
                "type": [
                    "string",
                    "null"
                ]
            }
        }
    }
}
```

Try to save, note the error:

**Power Automate**
image

image

**Logic Apps**
image

**Logic Apps Designer with fix**
image


hartra344:
Can you please add a proper explanation of what this change is trying to accomplish and why?

hartra344 (approved):

PR #4782, opened 2024-05-06
test(Templates): Adding Tests for Templates Designer / Card / State

Elaina-Lee:
For this PR: https://github.com/Azure/LogicAppsUX/pull/4761


libs/designer/src/lib/ui/templates/connections/displayConnections.tsx
```diff
@@ -2,18 +2,18 @@ import { Button } from '@fluentui/react-components';
...
   return (
     <>
       <div>Template Connections</div>
-      {Object.keys(connections).map((connection, index) => {
+      {connections.map((connection, index) => {
         return (
           <div key={index}>
             <b>
-              {index + 1}: {connection}{' '}
+              {index + 1}: {connection?.id}{' '}
```
ccastrotrejo, line 16: NIT: is there a specific reason for the empty space?

libs/designer/src/lib/ui/templates/__tests__/templateCard.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+          template1: template1Manifest,
+          template2: template2Manifest,
+        },
+      },
+    };
+    store = setupStore(minimalStoreData);
+
+    renderWithProviders(<TemplateCard templateName="template2" />, { store });
+
+    expect(screen.getByText('Create Workflow')).toBeDefined();
```
ccastrotrejo, line 99: I wonder how this will work for later implementations if we support intl

ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/__tests__/templateCard.spec.tsx
Elaina-Lee: TemplateCard already has a intl 'Create Workflow' text for the button, and the test passed!

PR #4815, opened 2024-05-08
feat(Template): Parameters Panel

Elaina-Lee:

-Added parameters panel
-Seperated validateParameterValueWithSwaggerType from workflowparametersSlice to minimize duplication, and reused it for templateSlice
-Wrote tests for validateParameterValueWithSwaggerType function
-Changed template.parameters type to record to align with how we handle parameters in designer
-Modified workflowparameterField with wrapped elements to load and catch value (with redux state) of parameters
-Wrote tests for modified workflowparameterField 

https://github.com/Azure/LogicAppsUX/assets/144840522/225090c1-13ae-44b7-9286-73fe0ee79f4c


LLotme (approved):

libs/designer-ui/src/lib/templates/templateparameters/templateparametersField.tsx
```diff
@@ -0,0 +1,146 @@
+import type { ILabelStyles, IStyle, ITextFieldStyles } from '@fluentui/react';
```
Eric-B-Wu, line 1: Is there a reason why we cant use our existing workflow parameters given that the UI is basically the same

LLotme (approved):

ccastrotrejo:
Pr is failing due to the test coverage in `lib.core.state.templates`

libs/designer-ui/src/lib/templates/templateparameters/templateparametersField.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    onChange?.({
+      name: definition.name,
+      newDefinition: { ...definition, value },
+    });
+  };
+
+  return (
+    <>
+      <div className="msla-workflow-parameter-field">
```
ccastrotrejo, line 106: NIT: I know this might be just a scratch and also the workflow parameters styles match the expected styles for this interfaces, however it would be great to name them differently or have a common styles name that is not attached to workflow-parameter wording.

libs/designer-ui/src/lib/templates/templateparameters/templateparametersField.tsx
```diff
@@ -0,0 +1,146 @@
...
+          styles={textFieldStyles}
+          onChange={onValueChange}
+          disabled={false}
+        />
+      </div>
+    </>
+  );
+};
+
+function stringifyValue(value: any): string {
```
ccastrotrejo, line 144: Maybe move this function to the utils or any library so we can reuse it. As I can see there the same in workflowparametersField.tsx, and maybe add a couple of unit tests to it

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/createWorkflowPanel.tsx
```diff
@@ -30,6 +31,7 @@ export const CreateWorkflowPanel = () => {
       </div>
 
       {connections ? <DisplayConnections connections={connections} /> : <>PLACEHOLDER: no connections to be made</>}
+      {parameters ? <DisplayParameters /> : <>PLACEHOLDER: no parameters</>}
```
ccastrotrejo, line 34: NIT: intl for placeholder?

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
```diff
@@ -0,0 +1,55 @@
...
+    description: 'Workflow Parameters Title',
+  });
+
+  const renderParameter = (item?: Template.ParameterDefinition): JSX.Element => {
+    if (!item) {
+      return <></>;
+    }
+    return (
+      <div className="msla-workflow-parameter">
+        <div>
```
ccastrotrejo, line 31: NIT: extra div?

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
```diff
@@ -0,0 +1,55 @@
...
+    id: 'X7X5ew',
+    description: 'Workflow Parameters Title',
+  });
+
+  const renderParameter = (item?: Template.ParameterDefinition): JSX.Element => {
+    if (!item) {
+      return <></>;
+    }
+    return (
+      <div className="msla-workflow-parameter">
```
ccastrotrejo, line 30: NIT: whenever rendering list of items its good to put a `key` attribute to the top parent tag, sometimes it messes it up with the re-rendering or update of fields.

libs/designer-ui/src/lib/templates/templateparameters/templateparametersField.tsx
Elaina-Lee: I tried re-using that at first, but realized I will need to add unneeded things to existing workflow parameters field if I do. Examples include:
- a description field rendering
- WorkflowParameterDefinition doesn't have description, but we'll have to add it for types purposes
- only the value field should be editable and required, meaning I have to change isEditable and isReadOnly from boolean to boolean | Record<string, boolean>
- WorkflowParameterUpdateHandler uses WorkflowParameterUpdateEvent which doesn't include 'required' field, so we will have to add it to this interface

and other formatting issues that seemed unncessary if calling WorkflowParmaretersField from Templates side.

I thought it would not make sense to add all those fields to WorkflowParametersField because they don't exist in WorkflowParameters

Elaina-Lee: What would it mean by keeping consistent with validation etc from template to designer? Validation happens at the state level, and so I had to seperate it regardless.. Also I couldn't take the whole part to the validation function since Templates only need to validate the value field. But open to suggestions!

LLotme, line 1: cc

Elaina-Lee: After offline discussion, decided to modify workflowParamaetersField and re-use this component

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
Elaina-Lee: This is needed for styling - following the same format as WorkflowParameter.tsx

ccastrotrejo (approved):

PR #4819, opened 2024-05-09
feat(designer): Adding Option to Open Token Picker by Default

pranaydubeymicrosoft:
Adding a flag focusOpensTokenPicker to designerOptions so that we can make the token editor open by default when the associated editor is selected. The behavior will align with the v1 designer experience.

### DIFF
--- a/libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
+++ b/libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
@@ -59,6 +59,7 @@ export interface SettingTokenFieldProps extends SettingProps {
   validationErrors?: string[];
   hideValidationErrors?: ChangeHandler;
   suppressCastingForSerialize?: boolean;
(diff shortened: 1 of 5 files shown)


libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -380,7 +382,7 @@ export const TokenField = ({
           labelId={labelId}
           className="msla-setting-token-editor-container"
           placeholder={placeholder}
-          basePlugins={{ tokens: showTokens }}
+          basePlugins={{ tokens: showTokens, focusOpensTokenPicker: focusOpensTokenPicker }}
```
hartra344, line 385:
```suggestion
          basePlugins={{ tokens: showTokens, focusOpensTokenPicker }}
```

hartra344 (commented):
Before marking this ready for review please include tests, both unit and/or E2E tests with standalone

PR #4821, opened 2024-05-09
Fix(Data Mapper): Disabling context menu as it causes app to crash

DanielleCogs:

removing right click menu 
 Data Mapper crashes because it is using a shared component that references Designer state
removing instead of fixing as DMV2 will not use the same context menu
does not block users, they can still delete using the delete key or from the function menu


takyyon (approved):

libs/data-mapper/src/lib/components/nodeCard/functionCard/ExpandedFunctionCard.tsx
```diff
@@ -45,10 +44,10 @@ export const ExpandedFunctionCard = (props: NodeProps<FunctionCardProps>) => {
 
   const contextMenu = useCardContextMenu();
 
-  const handleDeleteClick = () => {
-    dispatch(setSelectedItem(reactFlowId));
-    dispatch(deleteCurrentlySelectedItem());
-  };
+  // const handleDeleteClick = () => {
```
ccastrotrejo, line 47: NIT: is there a reason on why not deleting and just commenting it?

PR #4837, opened 2024-05-13
fix(Designer): Migrate some UI to Fluent v9 components

Eric-B-Wu:
We were having issues with darkmode, I'm not sure if this is limited to standalone env as it seems vscode may also have the issue. 

I'm migrating all Text componets to v9, as well as any other cases where I'm not seeing dark mode get caught


libs/designer-ui/src/lib/connectorsummarycard/index.tsx
```diff
@@ -42,7 +49,13 @@ export const ConnectorSummaryCard = (props: ConnectorSummaryCardProps) => {
         />
       </div>
       {displayRuntimeInfo ? (
-        <div className="msla-connector-summary-labels">{isBuiltIn ? <Text className="msla-psuedo-badge">{category}</Text> : null}</div>
+        <div className="msla-connector-summary-labels">
+          {isBuiltIn ? (
```
rllyy97, line 53: FV9 had an actual badge component we could convert this to

ccastrotrejo (approved):

libs/designer-ui/src/lib/html/plugins/toolbar/ColorPickerTextInput.tsx
```diff
@@ -26,7 +27,12 @@ export function TextInput({ label, value, onChange, placeholder = '', 'data-test
 
   return (
     <div className="msla-colorpicker-input-wrapper">
-      <label className="msla-colorpicker-input-label">{label}</label>
+      <label
```
rllyy97, line 30: Can we convert this to a V9 component so it just colors passively?

libs/designer-ui/src/lib/html/plugins/toolbar/DropdownColorPicker.tsx
```diff
@@ -97,7 +100,9 @@ export const DropdownColorPicker = ({
   return (
     <DropDown {...dropdownProps} disabled={disabled} stopCloseOnClickSelf buttonIconSrc={buttonIconSrc} editor={editor}>
       <div className="color-picker-wrapper" style={{ width: WIDTH }} ref={innerDivRef}>
-        <Text className="color-picker-title">{capitalizeFirstLetter(title ?? '')}</Text>
+        <Text className="color-picker-title" style={{ color: isInverted ? constants.INVERTED_TEXT_COLOR : constants.STANDARD_TEXT_COLOR }}>
```
rllyy97, line 103: I'm wondering why these aren't coloring properly by default, do these html components render outside of the FluentProvider or something?

rllyy97 (approved):

libs/designer-ui/src/lib/html/plugins/toolbar/DropdownColorPicker.tsx
Eric-B-Wu: yeah they do, which was why these weren't getting caught

libs/designer-ui/src/lib/html/plugins/toolbar/ColorPickerTextInput.tsx
Eric-B-Wu: Think this dropdown also gets rendered outside, so when using a label we won't be able to get the provider. However, Eric from PA is working/ has a PR for moving these dropdowns to fluent ui, so this won't be the case for very long

PR #4844, opened 2024-05-14
feat(designer): Adding support for dynamic input schema in dynamic parameters

preetriti1:

Feature

Currently in dynamic inputs if there is dynamic schema we were not making calls to fetch more inputs since we only put dependency on static inputs and only one level of dynamic input schema

With this feature, now if dynamic inputs have dynamic schema we will keep making dynamic data calls until there is none, there is currently no limit on this dynamic data chain.

No


apps/Standalone/src/designer/app/AzureLogicAppsDesigner/laDesigner.tsx
```diff
@@ -475,6 +475,91 @@ const getDesignerServices = (
           true /* supportsAuthenticationParameter */
         );
       },
+      getWorkflowSubSchema: (args: any) => {
```
ccastrotrejo, line 478: NIT: I know that the rest of the callbacks have any as type, but can we specify this one?

ccastrotrejo (approved):

apps/Standalone/src/designer/app/AzureLogicAppsDesigner/laDesigner.tsx
preetriti1: this is a random function which i added for testing purpose, it is not a known one hence I would not want to add it. Also I am going to use that for end to end testing.

PR #4871, opened 2024-05-21
fix(vscode): Cast and uncast condition expression for unit test

ccastrotrejo:
This pull request primarily introduces read-only functionality to the ExpressionEditor and ConditionExpression components in the `libs/designer-ui` library. It also includes changes to the serialization and deserialization of unit test definitions, and to the parameter value conversion in the `libs/designer` library.

Read-only functionality:

* Added `isReadOnly` prop to `ExpressionEditorProps` and `ExpressionEditor` function. The `isReadOnly` prop is used to set the `readOnly` property of the editor component. 
* Added `isReadOnly` prop to `GetConditionExpressionHandler` and used it in `AssertionField`. 
* Added `isReadOnly` prop to `ConditionExpressionProps` and `ConditionExpression` function. The `isReadOnly` prop is used to control the visibility of the `Callout` component and the `readOnly` property of the editor component. 

* Modified `serializeUnitTestDefinition` to create a token value segment for the assertion string and cast it to the correct type. 
* Modified `deserializeUnitTestDefinition` to parse the assertion string as an expression function.
Parameter value conversion:

* Introduced `castValueSegments` function and used it in `parameterValueToString` to cast the value segments based on the provided parameters.

### DIFF
--- a/libs/designer-ui/src/lib/expressioneditor/index.tsx
+++ b/libs/designer-ui/src/lib/expressioneditor/index.tsx
@@ -25,6 +25,7 @@ export interface ExpressionEditorProps {
   setExpressionEditorError: (error: string) => void;
   onFocus?: () => void;
   onContentChanged?(e: EditorContentChangedEventArgs): void;
+  isReadOnly?: boolean;
 }
 
 export function ExpressionEditor({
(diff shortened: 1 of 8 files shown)


ccastrotrejo (commented):
Self-reviewed

Issue #4885, opened 2024-05-23
[ICM] Search is using the browser defined language rather than the portal defined language


rllyy97:
### Describe the Bug with repro steps

1. Set your portal language to something different than your browser language.
1. Load a designer workflow
2. Try to search for an action / connector
3. Results will show in browser language rather than portal language

The expected behavior (what was done in old designer) is that the search results should be in the portal defined language.

Note: If anyone picks this up message me for the ICM

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Any

PR #4897, opened 2024-05-27
fix(designer): Fixed request locale not respecting portal language settings

rllyy97:
Fixed issue where our api calls were not respecting the selected language settings of portal.
If the user had their language set to chinese for example, their requests would not have the proper header and return results in english, when it should be requested with chinese headers.

Fixes: https://github.com/Azure/LogicAppsUX/issues/4885


rllyy97:
Not continuing with this approach

PR #4950, opened 2024-06-07
Feat(Data Mapper): Able to save data map

DanielleCogs:

none

1. **Created new service to manage calls that go to vscode and interact with the file system**
     this prevents us from passing functions down multiple components to be called

2. **Data Mapper is able to be loaded into VSCode, with limited functionality**
3. **Connections saved in the store and save using save button**

no

### DIFF
--- a/apps/vs-code-react/src/app/dataMapper/appV2.tsx
+++ b/apps/vs-code-react/src/app/dataMapper/appV2.tsx
@@ -1,3 +1,207 @@
+import { changeFetchedFunctions, changeSourceSchema, changeTargetSchema } from '../../state/DataMapSlice';
+import type { AppDispatch, RootState } from '../../state/store';
+import { VSCodeContext } from '../../webviewCommunication';
+import { DataMapperFileService} from './services/dataMapperFileService';
+import {
+  getFileNameAndPath,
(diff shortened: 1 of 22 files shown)


apps/vs-code-react/src/app/dataMapper/appV2.tsx
```diff
@@ -1,3 +1,207 @@
...
+      vscode.postMessage(msg);
+    },
+    [vscode]
+  );
+
+  const dataMapperFileService = useMemo(() => {
+    return new DataMapperFileService(sendMsgToVsix);
+  } , [sendMsgToVsix]);
+
+  // const addSchemaFromFile = (selectedSchemaFile: SchemaFile) => {
```
takyyon, line 63: can we please remove all the commented-out code from the PR? We can use this PR if we need to reference that code again?

takyyon (approved):

apps/vs-code-react/src/app/dataMapper/appV2.tsx
DanielleCogs: I feel like we're already missing out on code that we already have functionality for, so idk about this. Let's discuss during our dev sync?

PR #4967, opened 2024-06-11
feat(designer): Data structure changes to template

preetriti1:

feature - some data structure changes


Elaina-Lee (approved):

apps/Standalone/src/designer/app/SettingsSections/sourceSettings.tsx
```diff
@@ -7,7 +7,10 @@ import { ChoiceGroup, IconButton } from '@fluentui/react';
 import { useEffect, useMemo } from 'react';
 import { useDispatch } from 'react-redux';
 
-const SourceSettings = () => {
+const SourceSettings = ({
```
ccastrotrejo, line 10: NIT: can we have the `showEnvironment` and `showHistory` as part of the redux store? Instead of sending them in the props? 
So we can use them  as folllowing? 
`const showEnvironment = useShowEnvironment();`

preetriti1: Why, having them props is also fine, they dont change after first render and such props need not be in store... also this is just apps for test page

ccastrotrejo: ok, sounds good then. I was just suggesting it as it seems that that's the pattern we have been following.

PR #4979, opened 2024-06-14
fix(designer-ui): Remove problematic call to `setIsValuePlaintext`

ek68794998:
## Please check if the PR fulfills these requirements

## What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)

Bug fix

## What is the current behavior? (You can also link to an open issue here)

If you paste or type some HTML into the rich text editor which is not considered Lexical-safe, the HTML editor assumes that this is unsafe and forces the editor to switch to the code view (plaintext editor).

This is problematic from a technical point of view because it causes the editor to be deleted and recreated (firing blur events, re-renders, etc.), but also from a user point of view because it switches the user into the other mode without them being aware of it.

## What is the new behavior (if this is a feature change)?

As it turns out, the above behavior is actually completely unneeded.

- If you are typing in the rich text editor, everything you type should be Lexical-safe; that is, typing HTML there is not actually HTML, it's just plaintext (i.e., typing `<div>` is actually `&lt;div&gt;`). So, it's Lexical-safe and we don't need to force you to the code view.
- If you are typing in the code view, you are... already in the code view, so we don't need to force you to the code view.

Therefore, the new behavior is that we don't check whether to forcibly switch the user to code view when new text is entered. The behavior that remains is whether the button to switch from code view to rich text view is enabled.

## Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)

No

### Before

no-set-broken

### After

no-set-working

### DIFF
--- a/libs/designer-ui/src/lib/html/index.tsx
+++ b/libs/designer-ui/src/lib/html/index.tsx
@@ -44,7 +44,6 @@ export const HTMLEditor = ({ initialValue, onChange, ...baseEditorProps }: BaseE
       <HTMLChangePlugin
         isValuePlaintext={isValuePlaintext}
         setIsSwitchFromPlaintextBlocked={setIsSwitchFromPlaintextBlocked}
-        setIsValuePlaintext={setIsValuePlaintext}
         setValue={onValueChange}
       />
     </EditorWrapper>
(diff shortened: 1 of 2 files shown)


hartra344 (approved):
If you get a chance, can you please work with @eric-b-wu and the 2 of you get some end to end tests on this component? I feel like it has a lot of flake and constant changes and it does seem like it may be difficult to unit test, though that would also be nice.

ek68794998:
> If you get a chance, can you please work with @Eric-B-Wu and the 2 of you get some end to end tests on this component? I feel like it has a lot of flake and constant changes and it does seem like it may be difficult to unit test, though that would also be nice.

Yes, for sure. There are a lot of unit tests on the parts that _can_ be unit tested, but so much of it is built around the UI that it needs E2E for true coverage. I'll sync with him when he's around so we can figure out a plan.

PR #4982, opened 2024-06-14
feat(vscode): Auto regenerate managed api connection keys 

JoaquimMalcampo:

This introduces behavior which automatically regenerates managed api connection keys in the local settings file during project load and before flow execution.

Currently, managed api connection keys expire after 7 days of creation. There is no gesture or behavior to renew these connection keys.

The connection keys would be checked during project load and before flow execution and then regenerated if the key is expired or if its within the buffer of expiry. This is automatic and does not alert the customer when these keys are regerated.

No.

WIP


apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -123,6 +125,15 @@ async function addConnectionDataInJson(
   }
 }
 
+function isKeyExpired(jwtTokenHelper: JwtTokenHelper, connectionKey: string, bufferInHours: number): boolean {
```
ccastrotrejo, line 128: NIT: Lets create unit tests for this function

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -168,20 +179,24 @@ async function getConnectionReference(
     });
 }
 
+//change workflow file path to project root
```
ccastrotrejo, line 182: NIT: comment

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -14,8 +14,10 @@ import { addNewFileInCSharpProject } from './updateBuildFile';
 import { HTTP_METHODS, isString } from '@microsoft/logic-apps-shared';
 import type { ParsedSite } from '@microsoft/vscode-azext-azureappservice';
 import { nonNullValue } from '@microsoft/vscode-azext-utils';
+import { JwtTokenHelper, JwtTokenConstants } from '../../../../../vs-code-react/src/app/designer/services/JwtHelper';
```
ccastrotrejo, line 17: This import shouldn't be like this. Lets move the `JwtHelper` class to a library so it can be used in `vs-code-react` and `vs-code-designer` porjects. You can move the whole file to this path `libs/vscode-extension/src/lib/services`. Make sure to export it and import it in both projects and also add unit tests

apps/vs-code-designer/src/app/utils/appSettings/connectionKeys.ts
```diff
@@ -0,0 +1,54 @@
...
+    const workspaceFolder = await getWorkspaceFolder(context);
+    const projectPath = await tryGetLogicAppProjectRoot(context, workspaceFolder);
+    let azureDetails: AzureConnectorDetails;
+
+    if (projectPath) {
+      azureDetails = await getAzureConnectorDetailsForLocalProject(context, projectPath);
+      try {
+        const connectionsJson = await getConnectionsJson(projectPath);
+        if (isEmptyString(connectionsJson)) {
+          return;
```
mireedmsft, line 24: nit: should probably log something here in case this happens at runtime

mireedmsft (approved):

hartra344:
Is this ready to merge?

JoaquimMalcampo:
unit tests still in progress

libs/vscode-extension/src/lib/services/Workflow.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+        result = replaceAllOccurrences(result, `@appsetting('${settingName}')`, settingValue);
+        result = replaceAllOccurrences(result, `@{appsetting('${settingName}')}`, settingValue);
+      }
+    }
+
+    try {
+      return JSON.parse(result);
+    } catch (error) {
+      throw new Error('Failure in resolving connection parameterisation');
```
ccastrotrejo, line 34: NIT: Typo `parameterization`

libs/vscode-extension/src/lib/services/Workflow.ts
ccastrotrejo, line 1: NIT: I feel that having a class with just static functions isn't super useful, also you are not referencing any property from the class (e.g. `this.settings`) within the functions, therefore its just better to just export those functions

ccastrotrejo, line 1: Can we have some unit tests for this file as well?

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -35,17 +36,16 @@ export async function pickFuncProcess(context: IActionContext, debugConfig: vsco
   await callWithTelemetryAndErrorHandling(autoStartAzuriteSetting, async (actionContext: IActionContext) => {
     await runWithDurationTelemetry(actionContext, autoStartAzuriteSetting, async () => {
       await activateAzurite(context);
+      await verifyLocalConnectionKeys(context);
```
ccastrotrejo, line 39: NIT: can you take add another `runWithDurationTelemetry` and put the `verifyLocalConnectionKeys`. The `runWithDurationTelemetry` logs in the telemetry the time and if there is any error from whatever is getting executed within it. Or can you update the second param of the function, which is the prefix so we know its logging both of them?

ccastrotrejo (changes requested):

libs/vscode-extension/src/lib/services/Workflow.ts
JoaquimMalcampo: fixed

ccastrotrejo (changes requested):

ccastrotrejo (approved):

PR #5004, opened 2024-06-20
feat(Templates): Updating CreateWorkflow panel with latest UI + Logic

Elaina-Lee:

feature

- secondary button to be close and previous depending on tab type
- create tab to have two stages where the default second stage results in redirect based on callback
- workflow names duplicate validation, where the names are fetched every createWorkflowTab open
- workflow name to have validation according to how we handle it in portal
- review&create tab to have proper styling with Infobox
- added templateService to handle callbacks

https://github.com/Azure/LogicAppsUX/assets/144840522/d7b36e94-c200-4da5-8739-ae3302837bc4


libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
```diff
@@ -0,0 +1,52 @@
...
+  appId?: string;
+  httpClient: IHttpClient;
+  apiVersions: {
+    subscription: string;
+    gateway: string;
+  };
+  openBladeAfterCreate: () => void;
+}
+
+export class BaseTemplateService implements ITemplateService {
```
preetriti1, line 16: this should be standard template service because getting existing workflow names is goingt o be different for standard and consumption

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
```diff
@@ -59,10 +60,27 @@ export const TemplatePanel = ({ onCreateClick }: { onCreateClick: () => Promise<
     [currentPanelView, templateTitle, templateDescription, manifest?.details, intlText.CREATE_WORKFLOW, intlText.BY_MICROSOFT]
   );
   const onRenderFooterContent = useCallback(
-    () => (selectedTabProps?.footerContent ? <TemplatesPanelFooter {...selectedTabProps?.footerContent} onClose={dismissPanel} /> : null),
+    () =>
+      selectedTabProps?.footerContent ? (
+        <TemplatesPanelFooter
+          {...selectedTabProps?.footerContent}
+          secondaryButtonOnClick={selectedTabProps?.footerContent?.secondaryButtonOnClick ?? dismissPanel}
```
preetriti1, line 67: I think this should be part of tabprops because if secondary button is not prsent we do not need the click event as well

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
```diff
@@ -59,10 +60,27 @@ export const TemplatePanel = ({ onCreateClick }: { onCreateClick: () => Promise<
...
+      ) : null,
     [selectedTabProps, dismissPanel]
   );
 
+  useEffect(() => {
+    async function fetchWorkflowNames() {
+      const existingWorkflowNames = await TemplateService()?.getExistingWorkflowNames();
+
+      dispatch(setExistingWorkflowNames(existingWorkflowNames));
+    }
```
preetriti1, line 78: Use a query cache which will call templateserfvice and that method can be used as useExistingWOrkflowNames which is a react hook so we do not have to await

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.less
```diff
@@ -0,0 +1,49 @@
+.msla-templates-tab {
```
preetriti1, line 1: we have a panel.less in some templates folder can we use that?

preetriti1 (changes requested):

libs/designer-ui/src/lib/templates/templatesPanelFooter.tsx
```diff
@@ -33,12 +37,13 @@ export const TemplatesPanelFooter = ({
...
-        onClick={onClose}
+        onClick={secondaryButtonOnClick}
         style={{
           marginLeft: '8px',
         }}
+        disabled={secondaryButtonDisabled}
       >
-        {CLOSE}
+        {secondaryButtonText ?? CLOSE}
       </DefaultButton>
```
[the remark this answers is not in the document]
Elaina-Lee: We will render secondary buttons when props are not present; this is when it results in the default case of closing the panel

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
Elaina-Lee: We will render secondary buttons when props are not present; this is when it results in the default case of closing the panel. I can alternatively manually put close for every tab secondary button, but since it's so redundant for a default case, I'm wondering if we do want that

apps/Standalone/src/templates/app/TemplatesStandaloneDesigner.tsx
```diff
@@ -160,9 +148,17 @@ export const TemplatesStandaloneDesigner = () => {
...
+        connectionsData ?? {},
+        workflowAppData as WorkflowApp,
+        addConnectionDataInternal,
+        tenantId,
+        objectId,
+        canonicalLocation
+      ),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [connectionsData, settingsData, workflowAppData, tenantId, canonicalLocation]
+    [connectionsData, settingsData, workflowAppData, appId, tenantId, canonicalLocation]
```
preetriti1, line 161: appId is not needed

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -100,3 +101,9 @@ export const getConnectorResources = (intl: IntlShape) => {
     }),
   };
 };
+
+export const useExistingWorkflowNames = () => {
+  return useQuery(['getExistingWorkflowNames'], async () => {
+    return await TemplateService()?.getExistingWorkflowNames();
+  });
+};
```
preetriti1, line 109: This should be moved to queries folder to a template.ts file, like how we have it structured for designer. Not blocking this pr

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/connectionsTab.tsx
```diff
@@ -20,9 +21,10 @@ export const connectionsTab = (intl: IntlShape, dispatch: AppDispatch, nextTabId
     description: 'The tab label for the monitoring connections tab on the create workflow panel',
   }),
   description: intl.formatMessage({
-    defaultMessage: 'Connections Tab',
-    id: 'hsZ7em',
-    description: 'An accessability label that describes the connections tab',
+    defaultMessage:
+      'Configure connections to authenticate the following services and link your workflows with various services and applications, enabling seamless data integration and automation. Connections are required.',
+    id: '2zqm3d',
+    description: 'An accessability label that describes the objective of connections tab',
```
preetriti1, line 27: I thought I had this covered in the displayConnections component

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface TemplateServiceOptions {
+  baseUrl: string;
+  appId?: string;
+  httpClient: IHttpClient;
+  apiVersions: {
+    subscription: string;
+    gateway: string;
+  };
+  openBladeAfterCreate: () => void;
```
preetriti1, line 11: This should be present where we have defined the ITemplate service, if this service options is only for standard then defined in the standard template service file.

preetriti1 (approved):

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/connectionsTab.tsx
Elaina-Lee: I see it, I will be removing it from displayConnections component. This description folows the same format and styling as other tabs, so it'd be better to keep a consistent way for all.

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
Elaina-Lee: We plan to have the service for consumption as well for v2 so I put that into consideration and left it under base so consumption can import it as well. If we still think this is an issue, I will remove it in the next PR and move it under standard.

PR #5012, opened 2024-06-22
feat(designer-ui): Improve connection selector table UI

ek68794998:
## Please check if the PR fulfills these requirements

* [x] Tests for the changes have been added
* [ ] Docs have been added / updated

## What kind of change does this PR introduce?

Feature

## What is the current behavior?

Current connection selection table has a few issues:

1. It is not possible to see the user account that the connection is used for, if it uses a custom display name.
2. "Gateway" column never has content due to a typing issue.
3. Column text overflows into other columns.
4. Connection name (the technical reference name, not the display name) is always displayed despite not generally being useful information to the user.

## What is the new behavior?

1. If a connection has a user account that is not the same as the display name, it is shown underneath the display name.
2. If a connection is a gateway (and therefore not a user account connection), the gateway name is shown underneath the display name. (Meaning the "Gateway" column has been removed.)
3. Connection reference/logical name has been moved to a popover, along with the addition of connection creation date.
4. "Invalid" column now shows a green check by connections which are connected, for easy reference.
5. Connections are now sorted: selected first, then connected, then invalid.

## Does this PR introduce a breaking change?

Don't think so!

### Before

image

### After

image

image


hartra344 (approved):

rllyy97 (approved):
Looks great!  Thanks Eric

PR #5026, opened 2024-06-26
feat(templates): Add newly created connections in workflow create payload

preetriti1:
This should complete the connections work in template scenario.

- **Pending**
Integrating connections validation on traversing to review and create.


libs/designer/src/lib/ui/panel/connectionsPanel/createConnection/createConnectionWrapper.tsx
```diff
@@ -277,7 +279,7 @@ export const CreateConnectionInternal = (props: {
         let connection: Connection | undefined;
         let err: string | undefined;
 
-        const newName = await getUniqueConnectionName(connector.id, existingReferences);
+        const newName = connectionName ? connectionName : await getUniqueConnectionName(connector.id, existingReferences);
```
rllyy97, line 282: Nit: We could just use `??` here 
```suggestion
        const newName = connectionName ?? await getUniqueConnectionName(connector.id, existingReferences);
```

rllyy97 (approved):

libs/designer/src/lib/ui/panel/connectionsPanel/createConnection/createConnectionWrapper.tsx
preetriti1: Will do this in my next PR

PR #5029, opened 2024-06-27
feat(Templates): Use fuse for search operations in templates page

Elaina-Lee:

Uses fuse search with priority 1 for title and description, 2 for other details, and 3 for parameters and connections

https://github.com/Azure/LogicAppsUX/assets/144840522/f0c192e5-b0e6-44f9-bcb8-8e2956fabf57


libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -84,6 +74,56 @@ export const getFilteredTemplates = (
...
+      },
+      {
+        name: 'connections',
+        weight: 3,
+        getFn: ([_name, template]: [string, Template.Manifest]) =>
+          template.connections
+            ? Object.values(template.connections)?.reduce(
+                (acc: string[], connection) =>
+                  acc.concat([
+                    connection.connectorId, // TODO: change after PR (connection name to be stored in the state)
```
preetriti1, line 110: What do you mean by connection name getting stored in state?, we only store connection id and its display name

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -84,6 +111,15 @@ export const getFilteredTemplates = (
     });
     return hasDetailFilters;
   });
+
+  if (!filters.keyword) {
```
ccastrotrejo, line 115: Nit: Might be easier to return this at the beginning instead of waiting for all the logic to be executed

ccastrotrejo (approved):

libs/designer/src/lib/core/templates/utils/helper.ts
Elaina-Lee: We are returning the executed value which is filteredTemplateEntries (declared in line 89). This function is running two logics: filters and search (from keyword). If there's no keyword, we will still run through the filter logic, just won't go through the search logic below this line.

ccastrotrejo: ohh ok ok, makes sense. Sorry I got confused. Sounds good then!

PR #5033, opened 2024-06-27
feat(designer): Hybrid Work on Designer

Eric-B-Wu:
Implementation of hybrid support on Standalone/ Designer

### DIFF
--- a/libs/logic-apps-shared/src/designer-client-services/lib/standard/run.ts
+++ b/libs/logic-apps-shared/src/designer-client-services/lib/standard/run.ts
@@ -93,6 +93,15 @@ export class StandardRunService implements IRunService {
     }
   }
 
+  public static getProxyUrl(uri: string): { uri: string; headerPath: string } {
+    const appName = uri.split('hostruntime')[0].split('/');
+    appName.pop();
+    return {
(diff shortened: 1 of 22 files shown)


libs/logic-apps-shared/src/designer-client-services/lib/standard/run.ts
```diff
@@ -101,9 +110,22 @@ export class StandardRunService implements IRunService {
   async getRun(runId: string): Promise<Run> {
     const { apiVersion, baseUrl, httpClient, workflowName } = this.options;
 
-    const uri = `${baseUrl}/workflows/${workflowName}/runs/${runId}?api-version=${apiVersion}&$expand=properties/actions,workflow/properties`;
+    let uri = `${baseUrl}/workflows/${workflowName}/runs/${runId}?api-version=${apiVersion}&$expand=properties/actions,workflow/properties`;
 
     try {
+      if (uri.toLowerCase().includes('microsoft.app')) {
```
rllyy97, line 116: Should we use the hybrid app check method here?  It looks like that's what we're trying to do

rllyy97 (approved):

libs/logic-apps-shared/src/designer-client-services/lib/standard/run.ts
Eric-B-Wu: yep you're right, let me change that

PR #5040, opened 2024-06-28
feat(Templates): Sorting Dropdown + Functionality

Elaina-Lee:

* [X} The commit message follows our guidelines

Added sorting dropdown and sorting functionality when not searching for keywords

Also added the empty result UI

https://github.com/Azure/LogicAppsUX/assets/144840522/f8a04d76-4a34-4833-b64b-69d4aa058a85


libs/designer-ui/src/lib/templates/images/emptySearch.tsx
```diff
@@ -0,0 +1,118 @@
+export const EmptySearch: React.FC = () => {
```
preetriti1, line 1: What is this image?

Elaina-Lee: image

It is for the empty search result, this svg image I got from Bonica

libs/designer/src/lib/ui/templates/TemplatesDesigner.tsx
```diff
@@ -38,11 +55,21 @@ export const TemplatesDesigner = ({
...
-      </div>
+      {filteredTemplateNames && filteredTemplateNames?.length > 0 ? (
+        <div className="msla-templates-list">
+          {filteredTemplateNames.map((templateName: string) => (
+            <TemplateCard key={templateName} templateName={templateName} />
+          ))}
+        </div>
+      ) : (
+        <div className="msla-templates-empty-list">
+          <EmptySearch />
```
ccastrotrejo, line 66: NIT: I am just wondering if we can put some alt text or move it to an image tag just to make sure for whenever it doesn't load fine or screen readers.

ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/TemplatesDesigner.tsx
Elaina-Lee: Since it's a svg which is just a html being imported as a react component, I don't see a place to add alt text. Through research, it seems like I can put aria-label, so I will do that to the <div> in svg!

apps/Standalone/src/templates/app/TemplatesStandaloneDesigner.tsx
```diff
@@ -190,69 +190,75 @@ export const TemplatesStandaloneDesigner = () => {
...
+                margin: '20px',
+              }}
+            >
+              <TemplateFilters
+                connectors={[
                   {
-                    value: 'Workflow',
-                    displayName: 'Workflow',
+                    value: 'azureaisearch',
+                    displayName: 'Azure AI Search',
```
hartra344, line 202: While this is fine for now, I think we should figure out a way to make these not hard coded. It’s not going to scale

apps/Standalone/src/templates/app/TemplatesStandaloneDesigner.tsx
```diff
@@ -190,69 +190,75 @@ export const TemplatesStandaloneDesigner = () => {
...
-                ],
-              }}
-            />
-            <TemplatesDesigner createWorkflowCall={createWorkflowCall} />
+                ]}
+                detailFilters={{
+                  Trigger: [
+                    {
+                      value: 'Request',
+                      displayName: 'Request',
```
hartra344, line 221: Same as above, we don’t want to hard code these, we should initiate them based on what is available in the templates

libs/designer/src/lib/ui/templates/TemplatesDesigner.tsx
hartra344: Since the image is presentational only and not giving real information (empty search results should be announced on search update, make sure that happens) you can mark it as display presentation in CSS and screen readers will just skip over it. No need for alt text on this one I don’t think.

apps/Standalone/src/templates/app/TemplatesStandaloneDesigner.tsx
Elaina-Lee: This is coming in my next PR

apps/Standalone/src/templates/app/TemplatesStandaloneDesigner.tsx
Elaina-Lee: I'll bring it up in the discussion, I'm not sure how that'd be done; unless we want to loop to read all templates... Maybe through a back-end call?

PR #5049, opened 2024-06-30
feat(Templates): Connector Filter Dropdown to Fetch Service & Make Searchable

Elaina-Lee:

Let connection dropdown use searchService to fetch connectors using the unique connector Ids from all manifests
Added search in dropdown to enable search within the dropdown

https://github.com/Azure/LogicAppsUX/assets/144840522/93a3035e-90b2-4a19-b1e4-f19bbc1f2437


Elaina-Lee:
Discussion:
1. Do we want to display all 1000 connectors?
2. Selected checkboxes - do we want to bring them to the top?

libs/designer/src/lib/core/state/connection/connectionSelector.ts
```diff
@@ -39,6 +39,25 @@ export const useConnectorOnly = (connectorId: string | undefined, enabled = true
   );
 };
 
+export const useConnectorsOnly = (connectorIds?: string[]): UseQueryResult<Connector[] | undefined, unknown> => {
+  return useQuery(
+    ['apiOnly', connectorIds],
```
preetriti1, line 44: Please do not use same key we use for single connector, use apis and we can do useConnectors if this name is not already used

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
```diff
@@ -4,16 +4,30 @@ import type { AppDispatch, RootState } from '../../../core/state/templates/store
...
   const sortKey = useSelector((state: RootState) => state?.manifest?.filters?.sortKey);
   const intl = useIntl();
+  const { availableTemplates, subscriptionId, location } = useSelector((state: RootState) => ({
+    availableTemplates: state.manifest.availableTemplates ?? {},
+    subscriptionId: state.workflow.subscriptionId,
+    location: state.workflow.location,
+  }));
+  const allTemplatesUniqueConnectorIds = useMemo(() => {
+    const allConnections = Object.values(availableTemplates).flatMap((template) => Object.values(template.connections));
+    const uniqueConnectorsFromConnections = getUniqueConnectorsFromConnections(allConnections, subscriptionId, location);
```
preetriti1, line 26: You can just do .map here no, why do we need to new array again?



Just do `return getUniqueConnectorsFromConnections(allConnections, subscriptionId, location).map(connector => connector.connectorId)`

libs/designer/src/lib/core/templates/TemplatesDesignerContext.tsx
```diff
@@ -30,6 +31,7 @@ export interface TemplateServiceOptions {
   functionService?: IFunctionService;
   appServiceService?: IAppServiceService;
   connectionParameterEditorService?: IConnectionParameterEditorService;
+  searchService?: ISearchService;
```
preetriti1, line 34: It doesnt look like we are using the search service, so please remove it

preetriti1 (changes requested):

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
Elaina-Lee: I checked that the helper function that I reused (getUniqueConnectors) does not give back the unique connectors, so I used the Set to ensure the uniqueness of the ids

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
```diff
@@ -64,13 +78,17 @@ export const TemplateFilters = ({ connectors, detailFilters }: TemplateFiltersPr
         />
       </div>
       <div className="msla-templates-filters-dropdowns">
-        {connectors && (
+        {allUniqueConnectors && allUniqueConnectors.length > 0 && (
```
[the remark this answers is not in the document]
Elaina-Lee: Not at the moment, do we want to show the loading instead of the dropdown - would that UI make sense? Or should we just show the empty dropdown until the connectors load; this might be a design discussion

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
Elaina-Lee: I see that maybe this helper function was supposed to ensure the uniqueness. I think it's missing a logic; will fix it, that way I can remove this extra Set logic as well

preetriti1 (changes requested):

libs/designer/src/lib/core/state/connection/connectionSelector.ts
```diff
@@ -39,6 +39,25 @@ export const useConnectorOnly = (connectorId: string | undefined, enabled = true
   );
 };
 
+export const useConnectorsOnly = (connectorIds?: string[]): UseQueryResult<[string, Connector][] | undefined, unknown> => {
```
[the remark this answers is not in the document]
Elaina-Lee: Ah I was just trying to follow the pattern, thanks for explaining why the Only should not be there! Removed.

preetriti1 (approved):

PR #5070, opened 2024-07-03
feat(Templates): UI Sanitary Check, Ensuring Consistency

Elaina-Lee:

Templates UI Sanitary check
1. Fixed connection card random padding/margins to be consistent, ensuring bigger/smaller screens all have the same height of connectors regardless of connectors presence
2. Changed Templates Designer to grid type to accommodate more than 3 template cards accordingly
3. Fixed connector icons in the boxes breaking issue
4. Fixed styling for tags
5. Added learn more link for state type
6. Prerequisites to be markdown, letting links be clickable

Added coloring just to represent the fixed sizes of each column, the actual doesn't have rainbow coloring
image
image
image
image


Elaina-Lee:
Not ready: still working on the display connection popup logic change

libs/designer/src/lib/ui/panel/templatePanel/panel.less
```diff
@@ -152,6 +163,12 @@
                 }
             }
         }
+
+        .msla-template-overview-markdown {
+            * {
```
rllyy97, line 168: Does this rule need to be a wildcard or can we define this more explicitly?

ccastrotrejo (approved):

rllyy97 (approved):

libs/designer/src/lib/ui/panel/templatePanel/panel.less
Elaina-Lee: This is to remove the padding for the markdown library element. Adding the styling to the classname itself was not removing the vertical padding/margin, and through research I found out it's the html elements inside the markdown (internal) is causing the issue. I was unable to find out what specific element it would be since it's a third party library, but since the classname msla-template-overview-markdown is used specifically for the markdown, this shouldn't cause errors in other styling

PR #5110, opened 2024-07-11
fix(vscode): Add Parameters.json in .Net project conversion

mireedmsft:

    Bug fix to add parameters.json to the csproj file when converting to a .Net project

    parameters.json file is not properly included in the resulting .csproj file, which causes build issues when opened in VSCode or Visual Studio

    converted file opens and builds properly in VSCode

    No

    N/A

### DIFF
--- a/apps/vs-code-designer/src/app/commands/workflows/switchToDotnetProject.ts
+++ b/apps/vs-code-designer/src/app/commands/workflows/switchToDotnetProject.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import {
   connectionsFileName,
+  parametersFileName,
   funcIgnoreFileName,
   funcVersionSetting,
   hostFileName,
@@ -252,6 +253,10 @@ async function getArtifactNamesFromProject(target: vscode.Uri): Promise<Record<s
     if (file === connectionsFileName) {
       artifactDict['connections'].push(connectionsFileName);
       continue;
+    }
+
+    if (file === parametersFileName) {
+      artifactDict['connections'].push(parametersFileName);
     }
 
     if (file === 'Artifacts') {


apps/vs-code-designer/src/app/commands/workflows/switchToDotnetProject.ts
```diff
@@ -254,6 +255,10 @@ async function getArtifactNamesFromProject(target: vscode.Uri): Promise<Record<s
       continue;
     }
 
+    if (file === parametersFileName) {
+      artifactDict['connections'].push(parametersFileName);
```
mireedmsft, line 259: I didn't find any test coverage for this, am I missing something?

ccastrotrejo: No, I think there are no tests for this file.

ynauls, line 259: Is using 'connections' not going to cause a key collision in the dictionary overriding where 'connections' is already used?

ccastrotrejo: Yeah thats right, @mireedmsft  You might need to change to 'parameters'.

ccastrotrejo:
@mireedmsft You might need to make some changes in the `updateBuildFile` function in apps/vs-code-designer/src/app/commands/workflows/switchToDotnetProject.ts which also indicates which folders/files to copy

apps/vs-code-designer/src/app/commands/workflows/switchToDotnetProject.ts
mireedmsft: So in this case artifactDict is a dictionary of string to array, so adding the parameters file here just adds another element to the array at that index (and updateBuildFile already assumes that each element is an array and iterates through)

Issue #4765, opened 2024-05-03
Parameters wiped after saving an update


AndrewSCD:
### Describe the Bug with repro steps 

I made a change to a logic app which somehow resulted in all parameter fields for all SQL stored procedures being cleared, and possibly other parameters as well (I didn't check these before reverting to a previous version).

Below are the steps I took when this bug occurred.

1. Go To "Logic App Designer" 
2. Update a "Append to array variable" action on a parallel branch to only run after the previous action only when the previous action Has Timed Out or Has Failed. 
3. Save the Logic App. 

Note that on the previous version I'd saved, I'd created 4 parallel branches each with an "Append to array variable", and set the three other actions to Run After as per the above. It was only when I came in this morning and realised the last action hadn't been set correctly that I took the above actions.

Also note that after reverting to a previous version, and repeating the above steps, this bug did not occur again.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Microsoft Edge

### Additional context 

*No response*

rogerhillgsy:
I've had a very similar issue with a "Send email (v2)" action in the Logic Apps designer.
Trying to change the field I display in the body of the email. It keeps going back to the old value.


https://github.com/Azure/LogicAppsUX/assets/7746030/0ab76c0d-f2a0-42ac-9687-23ee24eda631

Eric-B-Wu:
Thank you @AndrewSCD and @rogerhillgsy for reporting these issues. I'm going to split up these 2 issues, as while they both relate to saving - they seem to reference different parts.
1. @AndrewSCD I'm trying to reproduce this issue you're seeing; however I don't seem to be able to. Do you have a part of your workflow json that you could share with me at wue@microsoft.com?
2. @rogerhillgsy there seems to be a regression in the new release where one of our partner teams had added a feature, but seemed to break this token experience. I've gone ahead and reverted their change, and hotfixed it to the new release. If you're still experiencing the issue, please let me know and I'll try addressing it.

AndrewSCD:
@Eric-B-Wu Thank you for your taking a look. After thinking about what happened, I have a theory about why this issue occurred that is not directly tied to the specific actions I took.

[...]

PR #5117, opened 2024-07-13
fix(Designer): Updated HTML Editor to support newline and carriage return characters (for dynamic content) 

spanvalkar:

This PR fixes a bug with the HTML Editor where the HTML Editor cannot support newline characters.

When a user adds dynamic content to any action with an HTML Editor and adds a newline character (for example, by pressing the 'enter' key) and then closes the action, reopening the action will cause the dynamic content to disappear. The dynamic content should stay in the action's HTML Editor, but it does not stay because the HTML Editor does not support newline characters.

[Bug 26654763](https://msazure.visualstudio.com/One/_workitems/edit/26654763): Expression gets lost when newline is used

My change enables the HTML Editor to support newline characters. With my change, a user can add newline characters to the HTML Editor for dynamic content and the newline character (and dynamic content) will be preserved. It will not spontaneously disappear (as seen in the bug linked above). This improves the user experience for LA Designer users because it ensures the HTML Editor can support newline characters.

This PR does not break any existing functionality. Rather, it only adds to the functionality of the HTML Editor by supporting newline characters.

Added a newline character after the 'a' in email:

image

Upon reopening the HTML-to-Text action, the dynamic content with newline character remains. Previously, HTML-to-Text field would appear blank (incorrectly):

image


hartra344:
@Eric-B-Wu  can you take a look at this?

@spanvalkar as discussed offline, I do want to hold off signing off on this until there are adequate tests added. This has been a painful issue in the past.

libs/designer-ui/src/lib/editor/base/utils/editorToSegment.ts
```diff
@@ -37,7 +37,8 @@ const getChildrenNodesToSegments = (node: ElementNode, segments: ValueSegment[],
 export const convertStringToSegments = (
   value: string,
   nodeMap: Map<string, ValueSegment>,
-  options?: SegmentParserOptions
+  options?: SegmentParserOptions,
+  convertSpaceToNewline?: boolean
```
ek68794998, line 41: Does it make sense to make this option part of `SegmentParserOptions`?

spanvalkar: Now that you mention it, I think it makes sense to make 'convertSpaceToNewline' a part of the SegmentParserOptions interface. I'll make the change now.

libs/designer-ui/src/lib/editor/base/utils/editorToSegment.ts
```diff
@@ -142,3 +163,19 @@ const collapseLiteralSegments = (segments: ValueSegment[]): void => {
...
   }
 };
+
+const removeNewlinesAndSpaces = (inputStr: string): string => {
+  return inputStr.replace(/\s+/g, '').replaceAll(/\n/g, '').replaceAll(/\r/g, '');
+};
+
+export const getTokenFromNodeMap = (segmentSoFar: string, nodeMap: Map<string, ValueSegment>): string | undefined => {
+  const copyOfSegment = segmentSoFar;
+  const processedId = copyOfSegment.replace(/\n/g, '');
```
ek68794998, line 173: For consistency, let's have a method called `removeNewlines` that removes both `\n` and `\r`, and use that on lines 173 and 175 here. `removeNewlinesAndSpaces` can also call that helper.

spanvalkar: Good point. I haven't gone back and optimized my changes yet, so I'll go back and make this change (among a few others I have in mind).

Eric-B-Wu:
asking for tests on the changes made here, as well as a test on a previous regression from this PR [Parameters wiped after saving an update · Issue #4765 · Azure/LogicAppsUX (github.com)](https://github.com/Azure/LogicAppsUX/issues/4765)

Eric-B-Wu (approved):

PR #5121, opened 2024-07-15
fix(designer): allow designer to revalidate assertions when it is deleted

samikay101:
[work Item Link](https://msazure.visualstudio.com/One/_workitems/edit/28547983)

# Pull Request

## What kind of change does this PR introduce?
Bug fix

## What is the current behavior?
When an assertion is deleted, the "Assertion name already exists." warning continues to display even though the assertion with the duplicate name has been removed.

## What is the new behavior?
The validation errors are cleared and revalidated when an assertion is deleted. This ensures that the "Assertion name already exists." warning is removed when the assertion with the duplicate name is deleted. Additionally, all remaining assertions are revalidated to ensure the state is consistent.

### Changes made:
1. Updated the `deleteAssertion` reducer to call `checkAssertionsErrors(state)` after deleting an assertion and its associated validation errors. This ensures that the validation errors are cleared and revalidated after the deletion.
2. Ensured that `checkAssertionsErrors` is used consistently across the reducers to handle validation after any changes (addition, deletion, or update) to assertions.

## Please Include Screenshots or Videos of the intended change:

Before: 
image

After: 
image

deleting the assertion, designer updates to remove false error warning: 
image

### DIFF
--- a/libs/designer/src/lib/core/state/unitTest/unitTestSlice.ts
+++ b/libs/designer/src/lib/core/state/unitTest/unitTestSlice.ts
@@ -134,6 +134,11 @@ const checkAssertionsErrors = (state: UnitTestState) => {
   }
 };
 
+/**
+ * Revalidates all assertions in the unit test state.
+ * This function clears existing validation errors and performs a fresh validation on all assertions.
+ * @param {UnitTestState} state - The current unit test state.
+ */
 const revalidateAllAssertions = (state: UnitTestState) => {
   state.validationErrors.assertions = {};
   Object.values(state.assertions).forEach((assertion) => {
--- a/libs/designer/src/lib/core/state/unitTest/unitTestSlice.ts
+++ b/libs/designer/src/lib/core/state/unitTest/unitTestSlice.ts
@@ -134,25 +134,6 @@ const checkAssertionsErrors = (state: UnitTestState) => {
   }
 };
 
-/**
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/state/unitTest/unitTestSlice.ts
```diff
@@ -134,6 +134,25 @@ const checkAssertionsErrors = (state: UnitTestState) => {
...
+/**
+ * Revalidates all assertions in the unit test state.
+ * This function clears existing validation errors and performs a fresh validation on all assertions.
+ * @param {UnitTestState} state - The current unit test state.
+ */
+const revalidateAllAssertions = (state: UnitTestState) => {
+  state.validationErrors.assertions = {};
+  Object.values(state.assertions).forEach((assertion) => {
+    const { name, id, assertionString } = assertion;
+    const validationErrors = {
```
mireedmsft, line 146: nit: unless I'm missing something this would have all validation results not just the errors? If so I would change this to validationResults instead of errors.

mireedmsft (approved):

ccastrotrejo (approved):
Looks good for me, just update Mike's NIT

ccastrotrejo (changes requested):
checkAssertionsErrors function might be doing the same thing

ccastrotrejo (approved):

PR #5145, opened 2024-07-18
feat(Templates): Parameters Type Validate + Minor UI Sanitary Check

Elaina-Lee:

Final clean-ups and UI changes

Before and After for connections spinner
image

Template Designer following the new design guide
image
image

Template Filters with updated Categories field
image


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -243,6 +243,20 @@ const getWorkflowParameters = (
...
+    equals(parameterType, UIConstants.WORKFLOW_PARAMETER_TYPE.STRING) ||
+    equals(parameterType, UIConstants.WORKFLOW_PARAMETER_TYPE.SECURE_STRING);
+
+  return isStringParameter
+    ? parameterValue
+    : parameterValue === ''
+      ? undefined
+      : typeof parameterValue !== 'string'
+        ? parameterValue
+        : JSON.parse(parameterValue);
```
Eric-B-Wu, line 257: I would wrap JSON.parse in a try catch block to not crash if parse fails

Elaina-Lee: I was re-using the part of the function from above function, but yes I wrapped it in try catch block and have the above function use this helper now!

Eric-B-Wu (approved):

libs/designer/src/lib/core/index.ts
```diff
@@ -35,7 +35,7 @@ export {
 export { useIsDesignerDirty, resetDesignerDirtyState } from './state/global';
 export { useAllSettingsValidationErrors } from './state/setting/settingSelector';
 export { useAllConnectionErrors } from './state/operation/operationSelector';
-export { serializeWorkflow } from './actions/bjsworkflow/serializer';
+export { serializeWorkflow, parseWorkflowParameterValue } from './actions/bjsworkflow/serializer';
 export {
```
preetriti1, line 39: There should not be a need to parse the parameter value in teh portal... our template designer should always give out correct values according to type for the parameters read during create.

PR #5169, opened 2024-07-22
fix(designer): Workflow canvas proper tab ordering

rllyy97:
## Main Changes

- Workflow canvas now has proper tab order
  - Tab order starts at the top of the workflow and works it's way down through the graph
  - When it encounters a scope node, it walks through each child before moving on to the next node
- Nodes now have a `nodeIndex` relative to the rest of the graph
  - This is what we are using for `tabIndex`es 
- Added a `useEdgesData` hook for accessing the `data` property of react flow edges

---
### Tab order visualization
image

---
### Tab walkthrough
canvas_focus_order


hartra344 (approved):

rllyy97:
I am doing one more check over possible performance impact then I will merge

rllyy97:
> I am doing one more check over possible performance impact then I will merge

I did investigation on a few small to very large workflows and was seeing no noticeable performance impact

PR #5177, opened 2024-07-23
feat(designer): Add unit testing for parseErrorMessage function

ccastrotrejo:
This pull request includes changes to the `libs/logic-apps-shared` package, specifically to the `http` utility and its associated tests. The changes primarily focus on enhancing the error handling capabilities of the `parseErrorMessage` function and ensuring its behavior is thoroughly tested.

* The `parseErrorMessage` function has been modified to ensure that if `error.data.error.message` is not available, `undefined` is returned instead of potentially returning a truthy value that isn't a message. This change ensures that the function's behavior is more predictable and consistent.

* A comprehensive suite of tests for the `parseErrorMessage` function has been added. These tests cover various scenarios to ensure the function behaves as expected when provided with different types and structures of error objects. This includes scenarios where the error object has a `message` property, is nested inside `data.error`, is a string inside `responseText`, is sent as a string, is `undefined` or `null`, doesn't have a JSON structure inside `responseText`, and when the error is an empty string, `null` or `undefined` with a default error message provided.

### DIFF
--- a/libs/logic-apps-shared/src/utils/src/lib/helpers/http.ts
+++ b/libs/logic-apps-shared/src/utils/src/lib/helpers/http.ts
@@ -14,7 +14,7 @@ export const parseErrorMessage = (error: any, defaultErrorMessage?: string): str
       return message;
     }
 
-    message = error?.data?.error?.message;
+    message = error?.data?.error?.message ?? undefined;
     if (message) {
       return message;
     }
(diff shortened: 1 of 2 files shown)


libs/logic-apps-shared/src/utils/src/lib/helpers/http.ts
```diff
@@ -14,7 +14,7 @@ export const parseErrorMessage = (error: any, defaultErrorMessage?: string): str
       return message;
     }
 
-    message = error?.data?.error?.message;
+    message = error?.data?.error?.message ?? undefined;
```
takyyon, line 17: Isn’t it undefined by default?

takyyon (approved):

PR #5190, opened 2024-07-25
fix(designer): Add mixed input editor

bnarayanared:
**What's new?**
Adding a new editor that toggles between floating action menu editor and schema editor.
This adds a toggle to toggle between the two editors.

_Toggle off loads the floating action menu editor:_
image

_Toggle on loads the schema editor:_
image

**This feature also includes:**
1. Add a local storage to save the toggle state
2. Adds a "useSchemaEditor" set to true whenever toggle is turned on. This also determines if the toggle needs to be turned on the first time the editor is loaded.


libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
+import type React from 'react';
```
phildaniels, line 1: The LAUX team may ask you to write some unit tests for this component, FYI

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onChange: ChangeHandler;
+  supportedTypes?: string[];
+  isRequestApiConnectionTrigger?: boolean;
+  label?: string;
+  readonly?: boolean;
+  useStaticInputs: boolean;
+}
+
+// Function to check if `initialValue` contains "useSchemaEditor": true
+const checkIsSchemaEditor = (initialValue: ValueSegment[]): boolean => {
```
phildaniels, line 19: You might want to put this into a helper file and export it, if you are required to write unit tests for this; by doing this you get get some more line coverage if you can test these. Right now they can only be tested indirectly via testing the component

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+      return parsedValue?.useSchemaEditor === true;
+    } catch (error) {
+      console.error('Error parsing value segment:', error);
+      return false;
+    }
+  });
+};
+
+// Function to add "useSchemaEditor": true if it's missing
+const addUseSchemaEditor = (value: ValueSegment[]): ValueSegment[] => {
```
phildaniels, line 32: Same thing here, if you write some unit tests for this it would be easier to test in isolation if the function exists in a file where it can be exported

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+  initialValue,
+  onChange,
+  supportedTypes,
+  isRequestApiConnectionTrigger,
+  label,
+  readonly,
+  useStaticInputs,
+}) => {
+  const [isSchemaEditor, setIsSchemaEditor] = useState<boolean>(() => {
+    const storedToggleValue = localStorage.getItem('toggle');
```
phildaniels, line 63: toggle is generic enough here where it might accidentally conflict with someone else trying to use a toggle at some point in local storage. I would do something like 

```typescript
const storedToggleValue = localStorage.getItem('MixedInputEditor_toggle');
```

or similar

phildaniels, line 63: toggle is generic enough here where it might accidentally conflict with someone else trying to use a toggle at some point in local storage. I would do something like 

```typescript
const storedToggleValue = localStorage.getItem('MixedInputEditor_toggle');
```

or similar

libs/designer-ui/src/lib/mixedinputeditor.tsx
phildaniels, line 19: You might want to put this into a helper file and export it, if you are required to write unit tests for this; by doing this you get get some more line coverage if you can test these. Right now they can only be tested indirectly via testing the component

libs/designer-ui/src/lib/mixedinputeditor.tsx
phildaniels, line 32: Same thing as my other comment on line 19

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+  useEffect(() => {
+    // Check if we need to update the state or localStorage
+    if (isSchemaEditor) {
+      const updatedValue = addUseSchemaEditor(initialValue);
+      // Only call onChange if there's a real update
+      if (JSON.stringify(updatedValue) !== JSON.stringify(initialValue)) {
+        onChange({ value: updatedValue });
+      }
+      // Update localStorage only if not already set
+      if (localStorage.getItem('toggle') !== 'true') {
```
phildaniels, line 82: Same thing as my comment on line 63 for this whole effect

libs/designer-ui/src/lib/mixedinputeditor.tsx
phildaniels, line 1: The LAUX team may ask you to write some unit tests for this component, FYI

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Function to check if `initialValue` contains "useSchemaEditor": true
+const checkIsSchemaEditor = (initialValue: ValueSegment[]): boolean => {
+  return initialValue.some((segment) => {
+    try {
+      const parsedValue = JSON.parse(segment.value);
+      return parsedValue?.useSchemaEditor === true;
+    } catch (error) {
+      console.error('Error parsing value segment:', error);
```
phildaniels, line 25: I generally avoid console logs like this. If it's a meaningful log that we'd like visibility on it I would use the logger service for the project. If it's for debug purposes only I would remove it

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
```diff
@@ -376,6 +377,18 @@ export const TokenField = ({
...
       );
 
+    case constants.PARAMETER.EDITOR.MIXEDINPUTEDITOR: {
+      return (
+        <MixedInputEditor
+          supportedTypes={editorOptions?.supportedTypes}
+          useStaticInputs={editorOptions?.useStaticInputs}
+          initialValue={value}
+          isRequestApiConnectionTrigger={editorOptions?.isRequestApiConnectionTrigger}
+          onChange={onValueChange || (() => {})}
```
phildaniels, line 387: The more modern appropriate operator in Javascript for what you're trying to do here is now `??` instead of `||`

Additionally, what you could do in your component instead of short circuiting here to a default, is something like

```typescript
onChange?.({ value });
```

That's how you can safely call `onChange`, even if it's `null` or `undefined`

libs/designer-ui/src/lib/mixedinputeditor.tsx
phildaniels: IE https://github.com/Azure/LogicAppsUX/pull/5190#discussion_r1691767729

libs/designer-ui/src/lib/mixedinputeditor.tsx
phildaniels: IE https://github.com/Azure/LogicAppsUX/pull/5190#discussion_r1691768602

libs/designer-ui/src/lib/mixedinputeditor.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }, [isSchemaEditor, initialValue, onChange]);
+
+  const handleToggleChange = (ev: React.MouseEvent<HTMLElement>, checked?: boolean) => {
+    setIsSchemaEditor(checked ?? false);
+  };
+
+  return (
+    <div>
+      <Toggle label="Use Schema Editor" checked={isSchemaEditor} onChange={handleToggleChange} disabled={readonly} />
```
phildaniels, line 96: Didn't catch this in my first pass on the PR, but the label should probably be localized IE

```typescript
// ...
import { useIntl } from 'react-intl';
// ...

export const MixedInputEditor: React.FC<MixedInputEditorProps> = ({
  // ...
}) => {
  // ...
  const intl = useIntl();
  // ...
  const label = intl.formatMessage({
    defaultMessage: 'Use Schema Editor',
    id: // eslint will tell you the correct id,
    description: ...,
  });
  
  // ...
  return (
    <div>
       <Toggle label={label}...
  // ...
```

phildaniels: file to place the localization can be found in https://github.com/Azure/LogicAppsUX/blob/main/Localize/lang/strings.json

libs/designer-ui/src/lib/settings/settingsection/settingTokenField.tsx
bnarayanared: Ah, thanks for the suggestion. Updated to use ?? and added the `onChange?.()` on the component

libs/designer-ui/src/lib/mixedinputeditor.tsx
bnarayanared: Done, added the localized string

libs/designer-ui/src/lib/mixedinputeditor.tsx
bnarayanared: Updated with tests now

hartra344 (approved):

PR #5202, opened 2024-07-25
feat(designer): Allow pinning a node to panel to view side-by-side with selected node

ek68794998:
## Requirement Checklist

### Manual test cases

- [x] Open & close one-node panel using mouse clicks
- [x] Open & close dual-node (pinned) panel using mouse clicks
- [x] Open & close one-node panel using keyboard
- [x] Open & close dual-node (pinned) panel using keyboard
- [x] Ensure focus is returned to selected node after panel close
- [x] Tab from designer into panel (should be covered by E2E test case)
- [x] Tab through panel controls
- [x] Use panel header: rename node, delete node, pin/unpin node
- [x] Can open "Change connection" and select a connection or cancel, if no node is pinned
- [x] Z-index is appropriately behind token pill, dialogs, modals, and dev tools pull-down menu
- [x] Can resize panel with a different minimum width for one-node or dual-node views

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior

Only one node may be selected and viewed in the node details panel at a time.

## New Behavior

In addition to selecting one node, an additional node may be pinned to the panel, allowing two nodes to be edited at once and permitting actions like copy-paste between inputs more easily.

This PR also updates the panel container to use Fluent 9 components.

## Impact of Change

Strictly speaking, there are no breaking changes. However, if any downstream hosts rely on Fluent 8 attributes or class names for the side panel, those will break.

There are a handful of state changes but they only apply to `panelV2`, so that should not be breaking. (Future PR will handle the breaking change for moving `panelV2` state logic to `panel`.)

image

image

pinned-panel-e2e


libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
```diff
@@ -1,101 +1,169 @@
...
     description: 'Button label for submitting a workflow to rerun from this action',
   });
-  useEffect(() => {
-    menuButtonRef.current?.focus();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isCollapsed]);
 
-  const isRight = headerLocation === PanelLocation.Right;
+  const unpinButtonText = intl.formatMessage({
+    defaultMessage: 'Unpin Action',
```
rllyy97, line 158: `Action` might need to be lowercase here

libs/designer/src/lib/core/state/panelV2/panelSlice.ts
```diff
@@ -97,24 +98,53 @@ export const panelSlice = createSlice({
       state.discoveryContent.selectedOperationGroupId = '';
       state.discoveryContent.isAddingTrigger = false;
     },
-    clearPanel: (state) => {
+    clearPanel: (state, action: PayloadAction<{ clearPinnedState?: boolean }>) => {
```
rllyy97, line 101: Are we able to make this action have an optional payload?
So we don't have to call `clearPanel({})` with an empty object

libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
ek68794998: I reused the same string as in designer panel:

image

But indeed the header menu uses lowercase :|

image

Not sure which we want to use..?

libs/designer/src/lib/core/state/panelV2/panelSlice.ts
ek68794998: I'm not sure if I can since I think `PayloadAction` has to be provided if specified even if I use something like `| undefined`, but let me see what I can do.

rllyy97 (approved):

libs/designer-ui/src/lib/panel/panelheader/panelheader.tsx
ek68794998: Using lowercase here, following up after to align others.

PR #5209, opened 2024-07-26
fix(all): Convert all label strings to `Standard case` except when referring to proper nouns

ek68794998:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

There are inconsistencies across the code base in terms of string casing. Sometimes, `Title Case` is used, while in other scenarios, `Standard case` is used. Per Riley's advice, we should be using `Standard case` except when referring to proper nouns (e.g., "Microsoft" or "Logic Apps").

## New Behavior

All labels (that I could find, anyway) now use `Standard case`. Additionally, usages of most acronyms such as "id", "uri", and "api" are now capitalized appropriately.

## Impact of Change

String changes only; no breaking changes unless E2E automation is referring to case-sensitive label values.

### Before

image

### After

image


libs/designer-ui/src/lib/utils/utils.ts
```diff
@@ -341,8 +341,8 @@ export function getStatusString(status: string, hasRetries: boolean): string {
 
     case Constants.STATUS.TIMEDOUT:
       return intl.formatMessage({
-        defaultMessage: 'Timed Out',
-        id: 'MsCHhQ',
+        defaultMessage: 'Timedout',
```
rllyy97, line 344: I think this should stay as two words
```suggestion
        defaultMessage: 'Timed out',
```

rllyy97 (approved):
I did not realize how many of these we had, thank you for changing these!

libs/designer-ui/src/lib/utils/utils.ts
ek68794998: Thanks for catching that! Just a typo on my side. 😅 Will fix.

PR #5219, opened 2024-07-29
fix(designer): Top margin for token picker if input field spans multiple lines

DhruvDhiman92:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
- header of the token picker is not visible when user write a super long multiline value in root input field.
image

## New Behavior
- Added top margin in the token picker so user can view the close, full-screen etc. operation and the input field when the root input field has long multi line values.

## Impact of Change

tokenPicker


libs/designer-ui/src/lib/tokenpicker/index.tsx
```diff
@@ -59,6 +59,15 @@ calloutStyles = isCopilotServiceEnabled()
     }
   : calloutStyles;
 
+const calloutStylesWithTopMargin: Partial<ICalloutContentStyles> = {
```
ek68794998, line 62: This will conflict with changes in #5218. Can you make sure that the extended styles retain all styles from `calloutStyles` itself?

Localize/lang/strings.json
```diff
@@ -2258,7 +2258,6 @@
   "_ljAOR6.comment": "Label for description of custom body Function",
   "_lkgjxD.comment": "Required string parameter for end time",
   "_lkmLHk.comment": "Toggle to use schema editor",
```
ek68794998, line 2260: You can merge from latest `main` and this file will be removed as this change has been completed in another PR.

hartra344 (approved):

PR #5225, opened 2024-07-30
fix(vscode): Async Download for Dependencies and Graceful Error Handling

mireedmsft:
## Requirement Checklist

## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

In the case of network congestion or failure while downloading dependencies we currently block the thread then timeout after 15 seconds, then block activation of the LA Standard Extension

## New Behavior

Downloads happen asynchronously, and will not block activation. Any failures to download (because of timeout or anything else) are then reported async after activation.

When network problems are resolved, the download then succeeds as expected.

## Impact of Change

No breaking changes

N/A


apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -68,17 +67,20 @@ export async function downloadAndExtractDependency(
...
-      executeCommand(ext.outputChannel, undefined, 'echo', `Downloading dependency from: ${downloadUrl}`);
-      // TODO: @mireed: change the download to be async and not block.
-      // TODO: @mireed: the NPM Request library is deprecated, change to use a different library.
-      // Get the Dependency Timout setting value, which is in seconds and convert to msec for the request library.
-      const dependencyTimeoutMsec = getDependencyTimeout() * 1000;
-      const downloadStream = request({ url: downloadUrl, timeout: dependencyTimeoutMsec }).pipe(fs.createWriteStream(dependencyFilePath));
-      downloadStream.on('finish', async () => {
-        await executeCommand(ext.outputChannel, undefined, 'echo', `Successfullly downloaded ${dependencyName} dependency.`);
+    const downloadPromise = axios({
+      method: 'get',
```
ccastrotrejo, line 71: NIT: We have a constant for this `HTTP_METHODS.GET`

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -108,15 +110,15 @@ export async function downloadAndExtractDependency(
...
-        resolve();
       });
-      downloadStream.on('error', reject);
+      writer.on('error', async (error) => {
+        throw error;
+      });
     });
   } catch (error) {
     vscode.window.showErrorMessage(`Error downloading and extracting the ${dependencyName} zip file: ${error.message}`);
     await executeCommand(ext.outputChannel, undefined, 'echo', `[ExtractError]: Remove ${targetFolder}`);
```
ccastrotrejo, line 120: Do you think it might be worth it to log some telemetry on the error?

apps/vs-code-designer/src/app/utils/binaries.ts
mireedmsft: awesome... fixed

apps/vs-code-designer/src/app/utils/binaries.ts
mireedmsft: added context parameter, and logging telemetry

ccastrotrejo (approved):

PR #5235, opened 2024-07-31
feat(vscode): Add telemetry for refreshing expired connection keys

JoaquimMalcampo:
Added telemetry for expired connection keys and information message when keys have been refreshed.

image

### DIFF
--- a/apps/vs-code-designer/src/app/utils/codeless/connection.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/connection.ts
@@ -225,7 +225,7 @@ export async function getConnectionsAndSettingsToUpdate(
       );
     } else if (
       localSettings.Values[`${referenceKey}-connectionKey`] &&
-      isKeyExpired(jwtTokenHelper, Date.now(), localSettings.Values[`${referenceKey}-connectionKey`], 192)
+      isKeyExpired(jwtTokenHelper, Date.now(), localSettings.Values[`${referenceKey}-connectionKey`], 3)
     ) {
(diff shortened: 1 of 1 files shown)


mireedmsft (approved):

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -235,11 +244,17 @@ export async function getConnectionsAndSettingsToUpdate(
         settingsToAdd,
         parametersFromDefinition
       );
+
```
ccastrotrejo, line 247: NIT: Why don't we just simply show the information message here? Instead of relying on a variable? since we don't use the variable anywhere else?

apps/vs-code-designer/src/app/utils/codeless/connection.ts
ccastrotrejo, line 1: I don't see any telemetry in the refreshing part. We are just logging it the message in the output channel
Can you please add it with the `context.telemetry.`

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/utils/codeless/connection.ts
JoaquimMalcampo: Agreed no change needed as variable in a for loop which checks if any key is refreshed.

apps/vs-code-designer/src/app/utils/codeless/connection.ts
JoaquimMalcampo: Fixed.

ccastrotrejo (approved):

PR #5255, opened 2024-08-01
fix(designer): Update selector for retry history and add useMemo

### LINKED ISSUE #5215: LA STD- unable to see retry history on HTTP actions in for each loop

Add HTTP action inside for each loop, let action fail with any 5xx or retriable error codes.

 Go to failed run and navigate to failed action, there is no retry tab at all

image

if you open the similar HTTP action outside loop it shows the retry

 and in HAR traces also we can find the retry history, but on UI its not showing

image

Standard (Portal)

New Designer

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

edge, chrome

### Additional context 

*No response*

ccastrotrejo:
This pull request includes changes to enhance the functionality and efficiency of the retry panel in the monitoring UI, as well as improvements to the workflow state selectors. The most important changes include adding a new `useEffect` hook to update retry history when the current page changes, and simplifying the selector logic for fetching retry history.

* Removed redundant code in `handlePagerChange` function to avoid unnecessary state updates.

* Simplified `useRetryHistory` selector to directly fetch retry history from `nodesMetadata`.

* Added `useEffect` import to support the new hook.

Fixes #5215 

https://github.com/user-attachments/assets/fc240009-5897-43fc-8d2d-29fb00b4928d


libs/designer-ui/src/lib/monitoring/retrypanel/index.tsx
```diff
@@ -20,6 +20,11 @@ export const RetryPanel: React.FC<RetryPanelProps> = ({ retryHistories, visible
     return calculateDuration(retryHistory.startTime, retryHistory.endTime);
   }, [retryHistory.endTime, retryHistory.startTime]);
 
+  useEffect(() => {
+    const newRetryHistory = retryHistories[currentPage - 1];
```
Elaina-Lee, line 24: Would there be a reason why we are not using useMemo for this and using useEffect to update the value here?

hartra344: I agree, that would probably be better and lead to fewer rerenders

libs/designer-ui/src/lib/monitoring/retrypanel/index.tsx
```diff
@@ -20,6 +20,11 @@ export const RetryPanel: React.FC<RetryPanelProps> = ({ retryHistories, visible
     return calculateDuration(retryHistory.startTime, retryHistory.endTime);
   }, [retryHistory.endTime, retryHistory.startTime]);
 
+  useMemo(() => {
```
Elaina-Lee, line 23: Followed up offline, but I would suggest with the correct usage of useMemo, it would look like this:

`const retryHistory = useMemo(() => retryHistories[currentPage - 1], [retryHistories, currentPage]);`

image

ccastrotrejo: Thank you so much I will make the change

Elaina-Lee (approved):

PR #5260, opened 2024-08-02
feat(designer): Initial implementation of undo-redo feature in designer

shuktika12163:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior

Users cannot undo/redo their actions in the designer.

## New Behavior

Design proposal: #5057 

- Compressed partial state history is stored in undoRedo slice. Users can specify max limit through designerOptions.hostOptions.maxStateHistorySize
- Undo/redo button can be added by hosts to link to the onclick functions "onUndoClick" and "onRedoClick". Two selectors "useCanUndo" and "useCanRedo" are provided for enabling/disabling the buttons.
- Users can undo/redo certain actions in their workflow (eg. add, delete, move nodes). There are a couple bugs that need to be addressed before parameter changes can be recorded for state history too (this will be fixed in follow-up PRs).
- By default, we currently store 0 states in state history unless host specifies a different max state count through hostOptions. This is to ensure that this new feature does not blow-up state space consumption unless host wants to enable it, plus avoid state history while bugs are being worked on. This can be updated once most of the bugs have been fixed. More logs and tests will also be added in follow-up PRs.

## Impact of Change

undoRedoInitialImplementation


libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  updateRunAfter,
+} from '../workflow/workflowSlice';
+
+export interface StateHistory {
+  past: string[];
+  future: string[];
+}
+
+// Omitted slices: designerView, designerOptions, dev, undoRedo
+export type UndoRedoPartialRootState = Pick<
```
shuktika12163, line 23: Looking for inputs here on whether some other slice could be removed too. I was unsure about "customCode" and "workflowParameters".

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  | 'panel'
+  | 'panelV2'
+  | 'settings'
+  | 'staticResults'
+  | 'tokens'
+  | 'workflow'
+  | 'workflowParameters'
+>;
+
+export const undoableActionTypes = [
```
shuktika12163, line 37: Looking for suggestions on any other actions that might need to be added here.

hartra344:
please make sure to write lots of tests for this because of how much it's touching

hartra344:
Closing as stale, please reopen when ready to review

shuktika12163:
Some data from PAuto flows on undo redo partial root state sizes without compression (in bytes). This can be used to help decide whether to compress or not (since compressing and decompressing will add some compute time) and how many max states to keep.

image

- Percentile 50: 244495.809029112 bytes (0.24 MB) per state => 100 states = 24 MB total for 50 percentile flows
- Percentile 75: 446546.914590654 bytes (0.44 MB) per state => 100 states = 44 MB total for 75 percentile flows
- Percentile 90: 866916.086735937 bytes (0.86 MB) per state => 100 states = 86 MB total for 90 percentile flows
- Percentile 95: 1342593.68540028 bytes (1.34 MB) per state => 100 states = 134 MB total for 95 percentile flows
- Percentile 99: 3130278.26487026 bytes (3.13 MB) per state => 100 states = 313 MB total for 99 percentile flows
- Percentile 99.99: 17459309.4829932 bytes (17.45 MB) per state => 100 states = 1745 MB total for 99.99 percentile flows

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
rllyy97: I think custom code and workflowparameters should both stay, and I think naming here is fine

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  addNode,
+  moveNode,
+  deleteNode,
+  addSwitchCase,
+  deleteSwitchCase,
+  addImplicitForeachNode,
```
rllyy97, line 43: Let's remove `addImplicitForeachNode` from this list and manually store a state at the end of `addForeachToNode` in `libs\designer\src\lib\core\utils\loops.ts`

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  pasteNode,
+  pasteScopeNode,
+  updateRunAfter,
```
rllyy97, line 46: Let's remove the run after actions from this

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  removeEdgeFromRunAfter,
+  addEdgeFromRunAfter,
+  updateNodeConnection.fulfilled,
+  /*
+   * TODO: Following actions to be added after bug fixes for:
+   * 1. Parameters panel is not re-rendering on undo/redo click
+   * 2. State history should be saved only when there is a change in state (clicking in/out of parameter is triggering save)
+   *
+   * Actions list:
```
rllyy97, line 55: I think we should combine all of the operation parameter / settings type actions as one state save that runs when the panel is closed if there were any edits.  We can do this in a separate PR but I think the flow should go:

1. Open panel
2. User makes edits
3. We mark the state as "dirty"
4. User tries to leave panel / we are "done" with the edits
5. Save state in history
6. Close panel

This would avoid issue #2

rllyy97 (approved):
Looks great!  Left a couple comments and I'm going to try it locally first but otherwise seems good to me

rllyy97 (changes requested):
I'm going to test it out locally then reapprove

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
shuktika12163: Ohh, I'll check with our PM on whether this is ok. In V1, we were storing a state for every parameter change (i.e. when user focused out of a parameter and there was a change in flow definition). I was thinking of looking into why the re-render isn't happening when panel is open on undo/redo click; and to add a condition on save state function to save only if there was a change from previous state.

libs/designer/src/lib/core/actions/bjsworkflow/undoRedo.ts
```diff
@@ -0,0 +1,75 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import { deflateSync, inflateSync } from 'zlib';
+import CONSTANTS from '../../../common/constants';
+import { setStateAfterUndoRedo } from '../../state/global';
+import { saveStateToHistory, updateStateHistoryOnRedoClick, updateStateHistoryOnUndoClick } from '../../state/undoRedo/undoRedoSlice';
+import type { UndoRedoPartialRootState } from '../../state/undoRedo/undoRedoTypes';
+import type { RootState } from '../../store';
+
```
shuktika12163, line 8: add try-catch around these functions to be safer in case compression fails

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
rllyy97: Ah ok I wasn't aware you had undo/redo previously, we can keep that same functionality then that's fine

rllyy97 (approved):

libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
rllyy97: Edit: Don't remove these

PR #5261, opened 2024-08-02
feat(designer): Flow perf improvements

rllyy97:
## Main Changes

- Added `onlyRenderVisibleElements` prop to ReactFlow
  - MUCH faster navigation on medium-large workflows, almost no hiccups unless zoomed out quite far
  - This makes offscreen elements inaccessible to tab navigation
  - To fix this I've added a listener on startup to infer if a user is using tab navigation for primary nav
  - If we infer user tab nav we turn this option off
- Memoization of a few components / component data
- Removed some unneeded div clutter
  - (Just trying to decrease number of rendered elements) 
- Removed uses of `getBrandColorRgbA` (was causing decent slowdown on larger node counts)
  - Added a much faster color+opacity function for the few cases we needed it
- Fixed loading state issue on `OperationCardNode` component
- Other refactoring

---

### Old Perf (Gif starts at load start)
- `~14s` from load start to idle on large test workflow
- Example has hiccups and missed inputs due to lag

08_02_perf_old

---

### New Perf (Gif starts at load start) 
- `~6s` from load start to idle on large test workflow
- Example generally smooth, a little stutter when zoomed out and lots of graphs / nodes on screen

08_02_perf_new


libs/designer-ui/src/lib/card/__test__/__snapshots__/cardfooter.spec.tsx.snap
```diff
@@ -10,6 +10,8 @@ exports[`lib/card/cardfooter > should render 1`] = `
 </div>
 `;
 
+exports[`lib/card/cardfooter > should render 2`] = `null`;
```
ccastrotrejo, line 13: Can you update the test or is there a new reason on why its rendering as null?

rllyy97: This is just the snapshot so it's generated based on the component, but I adjusted the footer so that it just returns null if there are no badges (less elements to render)

ccastrotrejo (approved):

PR #5300, opened 2024-08-06
chore(Designer): Added panel header focus test

rllyy97:
## Main Changes

Just adding a test to cover the panel header focus.
Catches the recent issue we had to patch with our portal release.

### DIFF
--- a/e2e/designer/tabFocus.spec.ts
+++ b/e2e/designer/tabFocus.spec.ts
@@ -45,5 +45,31 @@ test.describe(
       // Should be the first terminate node
       expect(await page.locator('*:focus').innerText()).toBe('Terminate');
     });
+
+    test('Should open node details panel with proper focus', async ({ page }) => {
+      const tab = async () => page.locator('*:focus').press('Tab');
+      const backTab = async () => page.locator('*:focus').press('Shift+Tab');
+
+      await page.goto('/');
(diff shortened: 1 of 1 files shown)


e2e/designer/tabFocus.spec.ts
```diff
@@ -45,5 +45,31 @@ test.describe(
...
+      // Focus should be on the collapse button
+      expect(await page.locator('*:focus').getAttribute('aria-label')).toBe('Collapse');
+      await tab();
+      // Focus should be on the node name textfield
+      expect(await page.locator('*:focus').getAttribute('value')).toBe('manual');
+      // Delete one character
+      await page.locator('*:focus').press('ArrowRight');
+      await page.locator('*:focus').press('Backspace');
+      // Focus should be on the node name textfield
+      expect(await page.locator('*:focus').getAttribute('value')).toBe('manua');
```
ccastrotrejo, line 67: should it be `manual` instead of `manua`?

rllyy97: Nope, we are testing deleting a character and keeping the focus in the textfield, up on line 65 we delete a character

ccastrotrejo: ohh ok ok, it makes sense

ccastrotrejo (approved):

PR #5303, opened 2024-08-06
chore(meta): revert copy paste action on v4.44

tbabatund:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [x] Other

## Impact of Change

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/copypaste.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/copypaste.ts
@@ -17,7 +17,7 @@ import type { ActionDefinition } from '@microsoft/logic-apps-shared/src/utils/sr
 import { initializeDynamicDataInNodes, initializeOperationMetadata } from './operationdeserializer';
 import type { NodesMetadata } from '../../state/workflow/workflowInterfaces';
 import { updateAllUpstreamNodes } from './initialize';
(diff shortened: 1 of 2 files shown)


hartra344:
What exactly is this reverting and why? Please add a description when you submit a PR.

tbabatund:
In the original bug fix https://github.com/Azure/LogicAppsUX/commit/6ab89920e86836ee7c5b8b3e7d81ad4916ebd904, when an action is copied (like Sharepoint Get -Items) , we noticed dynamic contents are not copied. Original bug addressed that issue but then exposed another issue whereby for each loop is not applied when Get-items is copied. We decided that will cause unpleasant customer experience and affect NPS score. So we thought it was best to have the fix for both issue at once. 

In make.test, create a manual flow.
ADD SP Get items action, set the parameters up.
Now add a Compose action. use Title/ID dynamic content from Get items.
Designer puts a for-each loop, which is great!
Image
Click Save/Publish
Readd title token in COmpose action. This is a bug being tracked separately.
Now, inside For-each, copy/paste the Get items action from above.
Now add Compose action, 
and pick Title token from Get items copy action.
Expecation: For each loop is applied
Reality: No for each loop is applied

tbabatund:
It was reverted in main already and we are on v4.45.

hartra344 (approved):

PR #5287, opened 2024-08-05
feat: Implementation of UiInteractionsService with custom menu component

Helennnnnnnnnnn:
## Requirement Checklist

* [x] The commit message follows our guidelines
* [x] Tests for the changes have been added (for bug fixes or features)
* [x] Docs have been added or updated (for bug fixes or features)

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Impact of Change

* [ ] **This is a breaking change.**

Design proposal:
https://github.com/Azure/LogicAppsUX/discussions/5250

## Screenshots or Videos (if applicable)

If UiInteractionsService does not pass in any array, it does not affect anything:

image

If UiInteractionsService pass a list of menu items:

nodeContextMenu

addButtonContextMenu

PR #5321, opened 2024-08-07
feat(designer): Moved edge context menu to one shared component

rllyy97:
## Main Changes

- Moved edge context menu out to a dedicated shared component
- Added `EdgeContextMenuData` to state
- Removed unused "no paste" code
- Fixed copy tooltip hiding bug

Performance marginally improved from reducing node count and hook usage, only one instance instead of one on each dropzone component.


ek68794998:
As a heads-up, this will probably conflict with #5287.

hartra344 (approved):

PR #5325, opened 2024-08-07
fix(Data Mapper): Update auto layout logic to only run when there are intersecting nodes on canvas

takyyon:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Impact of Change

Updating the Auto-layout logic so it only updates the position on first load and also only when there are intersecting function-nodes.
This prevents any auto-layout when a new function node is added (Day 0 scenario)


hartra344 (approved):
Much lack of description, there is

takyyon:
> Much lack of description, there is

Updated description to add a note and updated the title as well

PR #5354, opened 2024-08-12
fix(vscode): Fix Async Download for Dependencies

mireedmsft:
## Requirement Checklist

## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Async download for dependencies would delete the temp directory too early when a download failed. 

## New Behavior

Change to only delete the temp directory on success.

## Impact of Change

Dependencies download continues successfully when one download fails.

N/A


ccastrotrejo (approved):

apps/vs-code-designer/src/main.ts
```diff
@@ -56,7 +56,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
     runPostWorkflowCreateStepsFromCache();
 
-    await downloadExtensionBundle(activateContext);
+    try {
+      await downloadExtensionBundle(activateContext);
+    } catch (error) {
+      // log the error message the VSCode window and to telemetry.
+      const errorMessage = `Error downloading and extracting the Logic Apps Standard Extension Bundle: ${error.message}`;
```
ccastrotrejo, line 63: - Can you localize the text?
- Can you log the error in telemetry?
- I feel like the `Extension Bundle` shouldn't be uppercase

ccastrotrejo (changes requested):

apps/vs-code-designer/src/main.ts
mireedmsft: fixed extension bundle..

the more I think about it the more I like just logging to telemetry instead of to the console, so I swapped it around. Do we localize telemetry messages?

ccastrotrejo: No, telemetry is not localized

ccastrotrejo (approved):

PR #5359, opened 2024-08-12
fix(Templates): Image not Rendering in Prod + Sort By Text

Elaina-Lee:
## Requirement Checklist

## Type of Change
- let images be rendered by the url, preventing the build to not render images (no change in UX)

image


ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
```diff
@@ -110,6 +110,7 @@ export const TemplateFilters = ({ detailFilters }: TemplateFiltersProps) => {
         ))}
       </div>
       <div className="msla-templates-filters-sort">
+        <Text>Sort by</Text>
```
preetriti1, line 113: shouldnt this be localised

PR #5370, opened 2024-08-13
fix(designer): Fix Comments on nodes

Eric-B-Wu:
With the pinning of panel, comments were broken:
Before:
commentbefore

After:
commentfix

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "tabWidth": 2,
+  "useTabs": false,
+  "singleQuote": true,
+  "printWidth": 140
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,0 @@
-{
-  "tabWidth": 2,
-  "useTabs": false,
-  "singleQuote": true,
-  "printWidth": 140
-}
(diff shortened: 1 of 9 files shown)


.prettierrc
ccastrotrejo, line 1: Do we need this?

ccastrotrejo (approved):

.prettierrc
Eric-B-Wu: I think the default prettier setting are different from the one we currently use in the repo, so every time I do format, the whole file gets formatted differently. I'm thinking we have this config file, so we're all on the same page

ccastrotrejo: Sounds good, thanks for doing it!

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
ek68794998, line 1: Shouldn't this be a Biome config?

ek68794998:
Thanks for catching this!

.prettierrc
Eric-B-Wu: hmm i wonder why my biome config is not catching it locally then

PR #5386, opened 2024-08-14
fix(designer): Fix Standalone Designer Command offsetting Add Action popup

Eric-B-Wu:
Fixes an issue that we had moving edge context menu

Before:
before

After:
after

We'll need to do a similar pr in Portal repo


takyyon (approved):

rllyy97:
Are these the correct commits?  These look combobox related

Eric-B-Wu:
> Are these the correct commits? These look combobox related

sorry built off my other branch, pulled main now

Issue #5101, opened 2024-07-10
Designer doesn't load Json schema of HTTP Request trigger as defined


nobuko-yamada:
### Describe the Bug with repro steps 

Create 2 workflows as below

Defining “Description”, “Execarray” and “Mode”

![image](<https://github.com/Azure/LogicAppsUX/assets/115044612/299645ce-378f-4310-bc3b-3e2cf4186ce4>)

But there is no “description” Edit in Parent workflow below: (Only “Execarray”, “Mode”) 

![image](<https://github.com/Azure/LogicAppsUX/assets/115044612/16bbb47f-c0de-4475-82cb-304c38dce45f>)

Parent

```
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "CallChild": {
                "inputs": {
                    "body": {
                        "description": "description",
                        "execarray": "array",
                        "mode": "CDI"
                    },
                    "host": {
                        "workflow": {
                            "id": "workflow1"
                        }
                    },
                    "retryPolicy": {
                        "type": "none"
                    }
                },
                "runAfter": {},
                "type": "workflow"
            }
        },
        "contentVersion": "1.0.0.0",
        "outputs": {},
        "triggers": {
            "When_a_HTTP_request_is_received": {
                "inputs": {
                    "schema": {}
                },
                "kind": "Http",
                "type": "Request"
            }
        }
    },
    "kind": "Stateful"
}
```

Child

```
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {},
        "contentVersion": "1.0.0.0",
        "outputs": {},
        "triggers": {
            "When_a_HTTP_request_is_received": {
                "inputs": {
                    "schema": {
                        "properties": {
                            "description": {
                                "type": [
                                    "string"
                                ]
                            },
                            "execarray": {
                                "items": {
                                    "properties": {
                                        "name": {
                                            "type": "string"
                                        },
                                        "param1": {
                                            "type": "string"
                                        },
                                        "param2": {
                                            "type": "string"
                                        },
                                        "param3": {
                                            "type": "string"
                                        },
                                        "param4": {
                                            "type": "string"
                                        },
                                        "param5": {
                                            "type": "string"
                                        }
                                    },
                                    "type": "object"
                                },
[...]

PR #5409, opened 2024-08-15
fix(designer): Adding support for multiple types in the schema processor

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/5101

### DIFF
--- a/libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
+++ b/libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
@@ -88,39 +88,46 @@ export class SchemaProcessor {
     schema = this._dereferenceRefSchema(schema) as SchemaObject;
 
     let properties: SchemaProperty[];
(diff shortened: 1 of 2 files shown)


ccastrotrejo (approved):

libs/logic-apps-shared/src/parsers/lib/common/schemaprocessor.ts
```diff
@@ -88,39 +88,46 @@ export class SchemaProcessor {
...
 
     let properties: SchemaProperty[];
-    switch (schema.type) {
-      case SwaggerConstants.Types.Array: {
-        properties = this._getArrayProperties(schema);
-        break;
+    if (Array.isArray(schema.type)) {
+      properties = [];
+      for (const type of schema.type) {
+        properties = properties.concat(this.getSchemaProperties({ ...schema, type }));
```
preetriti1, line 94: can you add test for this case

PR #5412, opened 2024-08-15
chore(designer): Adding test for support for multiple types in the schema processor

Eric-B-Wu:
Adding unit test for https://github.com/Azure/LogicAppsUX/pull/5409


libs/logic-apps-shared/src/parsers/lib/common/__test__/schemaprocessor.spec.ts
```diff
@@ -1110,4 +1110,105 @@ describe('SchemaProcessor Tests', () => {
...
+      })
+    );
+
+    expect(parameters[1]).toEqual(
+      expect.objectContaining({
+        key: 'body.$.description',
+        name: 'description',
+        type: 'null',
+      })
+    );
```
preetriti1, line 1165: are you able to see these two inputs in teh UI, is the parameterId distinguishing it or how is this rendered"

Eric-B-Wu: You don't see 2 inputs in the UI, but it uses both the validation

PR #5434, opened 2024-08-20
fix: Adding PATCH method in IHttpClient

mcaupybugs:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

IHttpClient did not have PATCH method earlier. So with this PR it will have PATCH method which is required for hybrid logic apps specifically in the file system connection creation time.

## New Behavior
Adding the PATCH functionality in the HttpClient as this is being used by the fileSystem connection creation client. For hybrid logic apps this would be useful to create connection. 

## Impact of Change
IHttpClient will not support API calls with PATCH method.

### DIFF
--- a/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Services/HttpClient.ts
+++ b/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Services/HttpClient.ts
@@ -73,6 +73,26 @@ export class HttpClient implements IHttpClient {
     }
   }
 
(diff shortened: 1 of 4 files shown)


mcaupybugs:
@microsoft-github-policy-service agree [company="Microsoft"]

mcaupybugs:

hartra344:
The way this is written right now, it would have to be marked as a breaking change. Is there a reason that it's needed?

Eric-B-Wu:
Given that this is only used in one location, I'm going to ask hybrid team to move this logic into portal repo and use a corsFetch patch in only the location where its needed

PR #5439, opened 2024-08-20
fix(designer): Prevent Warning from appearing if there is no Id provided

Eric-B-Wu:


preetriti1 (approved):

preetriti1:
BTW why would we not have id set? isnt it set everywhere else?

Eric-B-Wu:
> BTW why would we not have id set? isnt it set everywhere else?

I'm seeing that sometimes under additional parameters its not set, and then the warning appears because multiple keys are undefined

PR #5442, opened 2024-08-20
fix(Designer): Adding Test to getDropdownOptionsFromOptions

Eric-B-Wu:
Creating test for getDropdownOptionsFromOptions changes in https://github.com/Azure/LogicAppsUX/pull/5421


hartra344 (approved):

libs/designer-ui/src/lib/editor/base/utils/__test__/helper.spec.ts
```diff
@@ -33,3 +33,49 @@ describe('Helper functions', () => {
...
+    const result = getDropdownOptionsFromOptions(editorOptions);
+    expect(result).toEqual(editorOptions.options.nested);
+  });
+
+  it('handles cases where displayName is not a string', () => {
+    const editorOptions = {
+      options: [
+        { key: '1', value: 'Option 1', displayName: { text: 'Option 1' } },
+        { key: '2', value: 'Option 2', displayName: ['Option 2'] },
+      ],
```
preetriti1, line 67: something is not right, value can be any type for displayName is always a string

libs/designer-ui/src/lib/editor/base/utils/__test__/helper.spec.ts
```diff
@@ -33,3 +33,49 @@ describe('Helper functions', () => {
...
+    };
+    const result = getDropdownOptionsFromOptions(editorOptions);
+    expect(result).toEqual([
+      { key: '1', value: 'Option 1', displayName: '{"text":"Option 1"}' },
+      { key: '2', value: 'Option 2', displayName: '["Option 2"]' },
+    ]);
+  });
+
+  it('returns an empty array when no options are available', () => {
+    const editorOptions = {};
```
preetriti1, line 77: again why in the editorOptions would be not have options? this is coming from swagger right... do you see any connector with such info? Even if that is the case in the place where we set editorOPtions we should set it to options: []

libs/designer-ui/src/lib/editor/base/utils/helper.ts
```diff
@@ -197,7 +197,8 @@ export const getDropdownOptionsFromOptions = (editorOptions: any): ComboboxItem[
 
   // handle cases where the displayName is not a string
   dropdownOptions = dropdownOptions.map((option) => {
-    const stringifiedDisplayName = typeof option.displayName === 'string' ? option.displayName : JSON.stringify(option.displayName);
+    const displayName = option.displayName ?? option.key ?? option;
+    const stringifiedDisplayName = typeof displayName === 'string' ? displayName : JSON.stringify(displayName);
```
preetriti1, line 201: I think you were supposed to do this for the value field not displayName

preetriti1 (changes requested):

libs/designer-ui/src/lib/editor/base/utils/__test__/helper.spec.ts
Eric-B-Wu: yeah this was a check because there was an issue where some connectors were giving us display names where they were non-string types, which was crashing our designer

libs/designer-ui/src/lib/editor/base/utils/__test__/helper.spec.ts
Eric-B-Wu: this was just a safety check in a test

libs/designer-ui/src/lib/editor/base/utils/helper.ts
Eric-B-Wu: I've seen display names come in non-string types, there was a time where it was an object, another time it was a number. Both caused the designer to crash. This is just code to insure at least we don't crash

PR #5449, opened 2024-08-21
feat(designer): Allow Copilot to reference dynamic input/data while creating expression

deschandhok24:
## Requirement Checklist

* [X] Tests for the changes have been added (for bug fixes or features)
* [X] Docs have been added or updated (for bug fixes or features)

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior

right now copilot is not given reference input data, making the ai expression builder not as great as a user experience as possible. 

## New Behavior

Allow Copilot to reference dynamic input/data while creating expression, allowing for a better user experience.

### DIFF
--- a/libs/logic-apps-shared/src/utils/src/lib/models/token.ts
+++ b/libs/logic-apps-shared/src/utils/src/lib/models/token.ts
@@ -1,5 +1,15 @@
-import type { TokenType } from '../../editor';
-import type { OpenAPIV2 } from '@microsoft/logic-apps-shared';
+import type { OpenAPIV2 } from '../../';
+
+export const TokenType = {
+  FX: 'fx',
+  ITEM: 'item',
+  ITERATIONINDEX: 'iterationIndex',
+  OUTPUTS: 'outputs',
+  PARAMETER: 'parameter',
+  VARIABLE: 'variable',
+} as const;
+
+export type TokenType = (typeof TokenType)[keyof typeof TokenType];
 
 export interface Token {
   key: string;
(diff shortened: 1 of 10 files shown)


hartra344 (approved):

libs/logic-apps-shared/src/utils/src/lib/models/token.ts
```diff
@@ -1,5 +1,15 @@
-import type { TokenType } from '../../editor';
```
preetriti1, line 1: Can this be exported from editor?

deschandhok24: I thought since this token.ts file is now in logic-apps-shared now it cant reference anything in libs/designer-ui/src/lib/editor.

libs/logic-apps-shared/src/utils/src/lib/models/token.ts
```diff
@@ -1,5 +1,15 @@
...
-import type { OpenAPIV2 } from '@microsoft/logic-apps-shared';
+import type { OpenAPIV2 } from '../../';
+
+export const TokenType = {
+  FX: 'fx',
```
preetriti1, line 4: How does this recognize both tokentype as a type and a const? Won't this cause confusion whoever is importing it? Won't it be better to name it separately

deschandhok24: This seems to be a pattern we are using throughout LAUX https://github.com/Azure/LogicAppsUX/blob/2a28e353aea28d9b540dc576431dafd4c2a3a53b/libs/designer-ui/src/lib/chatbot/components/conversationItem.ts#L44.

hartra344: @preetriti1 this does suck but it is the best way to do ENUMs in typescript without doing typescript Enums which cause problems with some compilers.

preetriti1: I see, I remember the compiler problem with enums..

preetriti1 (approved):

PR #5480, opened 2024-08-23
fix(designer): fix linebreak issues (#5392)

ccastrotrejo:
fix linebreak issues


takyyon (approved):

hartra344 (approved):

hartra344:
Is this already in main?

takyyon:
> Is this already in main?

Yup it is - https://github.com/Azure/LogicAppsUX/pull/5392

I was talking Carlos earlier today and it seems like if we start from current designer, it should be fine but if we are moving from legacy to current, that’s when the line-break breaks. 
We will put it in hotfix pipeline and have the build ready in case we need to hotfix. Otherwise normal deployment is going to begin tomorrow anyways.

PR #5488, opened 2024-08-23
Fix(designer): Nested for each loops aren't showing iterations in active flow runs.

AfroJoe759:
## Requirement Checklist

## Type of Change

* [x] Bug fix

## Current Behavior

When you are monitoring an active flow run, when the flow completes you won't be able to see the iterations of nested Appy to Each or For Each actions since the buttons don't render. The only workaround is to wait until the flow completes and re-navigate to the history page for that particular run.

In PPUX this repro's consistently, in LAUX you can repro if you refresh the flow history page for an active run when it enters the outermost forEach loop without the inner forEach loop completing. In the LAUX scenario, even clicking the refresh button in the navbar won't bring the arrow buttons back, preventing one from viewing iterations without leaving the page and coming back after the flow completes. 

**Monitoring View (Running Flow In Progress)**
image

**Monitoring View (Running Flow Completed)**
image

## New Behavior

New behavior adds a check when rendering the LoopsPager for forEach actions, which allows the buttons to render properly.

**Monitoring View (Running Flow in Progress)**
image

**Monitoring View (Running Flow Completed)**
image


libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
```diff
@@ -334,7 +334,10 @@ const ScopeCardNode = ({ data, targetPosition = Position.Top, sourcePosition = P
             nodeIndex={nodeIndex}
           />
           {showCopyCallout ? <CopyTooltip targetRef={rootRef} hideTooltip={clearCopyCallout} /> : null}
-          {isMonitoringView && normalizedType === constants.NODE.TYPE.FOREACH ? (
+          {isMonitoringView &&
```
preetriti1, line 337: How come adding more checks is making this true to viewing the loopspager but with fewer checks dit was returning null?

preetriti1 (dismissed):

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
AfroJoe759: This bug item goes over more details: 
   - https://msazure.visualstudio.com/One/_workitems/edit/28724205/
   
I discussed this with @ccastrotrejo offline, but basically what we believe is happening is that if a call fails for the repetitions or scopeRepetitions API, the data required to determine the iterations/values won't be present when the LoopsPager attempts to render. 

When this happens the buttons for navigating for iterations won't appear for the nested actions, which results in the bug above. In both PPUX and LAUX, subsequent calls can be made to the same APIs which will give the correct data, but the UI doesn't refresh with the new API calls. 

Based on what we observed, calls to the repetitions APIs for active flow runs that fail typically do so when done for actions that haven't started or are in progress (e.g. the inner for each actions). So, what I noticed is that if we waited to render until after the state was no-longer InProgress, we would have the proper information and could render the buttons as expected.

AfroJoe759: @preetriti1 Is there a specific concern you have with this change?

ccastrotrejo: Hey @AfroJoe759 I think this one might not be the approach to solve this. As Priti said, adding more checks to the decision doesn't actually renders the pager. Specially because these are `&&` checks and not `||` checks.

preetriti1: So whatever you mentioned, if this is in Progress you do not want to render the pager if this is what we want, @ccastrotrejo then lets check the data if is in progress and put it in useMemo or useEffect block so whenever status changes UI is re-rendered, you can do it either in the pager itself by passing the data as props or from state or do it here.

If something is in effect of some data is being retrieved then we should use the react hooks for re-render and it makes it reliable

AfroJoe759: @preetriti1 @ccastrotrejo sounds good, if you both are fine with the current approach I can move the logic to a useMemo or useEffect block instead.

ccastrotrejo: Ok yes. So to be extra clear about this.
What @AfroJoe759 expects to happen over here is that once outer loops completes, the inner loop pages most render.
So please do as Priti said. Figure out a way to render catch the update of it, this can be in a useEffect or useMemo, depending on the implementation

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
ccastrotrejo, line 1: Ok this works because as you introduced a new state, everytime either the metadata/isMonitoringView gets updated the whole component gets updated. I don't think thats the best idea.

- if you log the metadata, does it gets updated? (without your changes?)
- could you wrap the loopsPager component in a useMemo and see if this works? this might prevent to re-render everything

ccastrotrejo (changes requested):

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
AfroJoe759: @ccastrotrejo I'm not too familiar with useMemo(), is your ask for me to do something like this?
```  const renderLoopsPager = useMemo(() => {
     if (metadata?.runData?.status && metadata.runData.status !== 'InProgress') {
      return true;
     }

      return false;

  }, [metadata]);

// Miscellaneous Code 
// Render Component
{normalizedType === constants.NODE.TYPE.FOREACH && isMonitoringView &&  renderLoopsPager? (
            <LoopsPager metadata={metadata} scopeId={scopeId} collapsed={graphCollapsed} />
          ) : null}
  ```
  
  OR is your ask for me to do something closer to this?
  
````
  const renderLoopsPager = useMemo(() => {
     if (metadata?.runData?.status && metadata.runData.status !== 'InProgress') {
      return (<LoopsPager metadata={metadata} scopeId={scopeId} collapsed={graphCollapsed} />);
     }

      return null;

  }, [graphCollapsed, metadata, scopeId]);

// Miscellaneous Code
// Render Component
{normalizedType === constants.NODE.TYPE.FOREACH && isMonitoringView ? renderLoopsPager : null}

```

AfroJoe759: Without my change, the metaData itself does update over time (e.g. runData.status updates throughout an action's run).

ccastrotrejo: Ok, so lets implement your second implementation, Thats the one that I was thinking about

AfroJoe759: @ccastrotrejo My most recent commit should address this. Let me know if there are any other concerns.

ccastrotrejo (approved):

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
```diff
@@ -269,6 +269,14 @@ const ScopeCardNode = ({ data, targetPosition = Position.Top, sourcePosition = P
     statusRun,
   ]);
 
+  const renderLoopsPager = useMemo(() => {
+    if (metadata?.runData?.status && metadata.runData.status !== 'InProgress') {
```
preetriti1, line 273: please use equals here to make sure we are doing this case insensitively

preetriti1 (approved):

libs/designer/src/lib/ui/CustomNodes/ScopeCardNode.tsx
AfroJoe759: Updated to use the equals method

PR #5491, opened 2024-08-23
fix(designer): Add Single Token Interpolation Option For FloatingActionOutputsViewModel

pranaydubeymicrosoft:
This change adds the option to interpolate single token inputs. This is set to false by default to preserve the current functionality. The component leveraging this functionality is only used in Power Automate. This change will align the V3 behavior with V1 for the FloatingActionMenu.

Current Power Automate:

image

image

With change:

lauxinterpolationbugcrop

### DIFF
--- a/libs/designer/src/lib/core/utils/parameters/helper.ts
+++ b/libs/designer/src/lib/core/utils/parameters/helper.ts
@@ -2529,7 +2529,15 @@ const getStringifiedValueFromFloatingActionMenuOutputsViewModel = (
       if (valueSegments?.length) {
         outputValueMap[keyFromTitle] =
           // We want to transform (for example) "1" to 1, "false" to false, if the dynamically added parameter type is not 'String'
-          parameterValueWithoutCasting({ type: config.type, value: valueSegments, ...commonProperties } as any);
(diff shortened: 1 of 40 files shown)


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -2529,7 +2529,15 @@ const getStringifiedValueFromFloatingActionMenuOutputsViewModel = (
...
           // We want to transform (for example) "1" to 1, "false" to false, if the dynamically added parameter type is not 'String'
-          parameterValueWithoutCasting({ type: config.type, value: valueSegments, ...commonProperties } as any);
+          // We will only interpolate a single token if the parameter config is of type 'string'
+          parameterValueWithoutCasting(
+            {
+              type: config.type,
+              value: valueSegments,
+              ...commonProperties,
+            } as any,
+            /* shouldInterpolateSingleToken */ config.type === 'string'
```
Eric-B-Wu, line 2539: nit: use the constant from libs/designer/src/lib/common/constants.ts to replace 'string'

Eric-B-Wu (approved):

PR #5536, opened 2024-08-28
fix(designer): Fix Allow Copilot to reference dynamic input/data while creating expression

deschandhok24:
## Requirement Checklist

* [N/A] Tests for the changes have been added (for bug fixes or features)
* [N/A] Docs have been added or updated (for bug fixes or features)

## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
Passing in the tokenGroup which is currently being done is not relevant for the context for the expression assisant.

## New Behavior
Passing in the expressionGroup which is currently being done is relevant for the context for the expression assistant.

## Impact of Change

Passing in the expressionGroup instead of the tokenGroup as NL2Flow uses that instead.

### DIFF
--- a/libs/designer-ui/src/lib/tokenpicker/index.tsx
+++ b/libs/designer-ui/src/lib/tokenpicker/index.tsx
@@ -267,7 +267,7 @@ export function TokenPicker({
         ref={containerRef}
       >
         <Nl2fExpressionAssistant
-          tokenGroup={tokenGroup ?? []}
+          tokenGroup={expressionGroup ?? []}
           isFullScreen={fullScreen}
           expression={expression}
           isFixErrorRequest={expressionEditorError !== ''}


libs/designer-ui/src/lib/tokenpicker/index.tsx
```diff
@@ -267,7 +267,7 @@ export function TokenPicker({
         ref={containerRef}
       >
         <Nl2fExpressionAssistant
-          tokenGroup={tokenGroup ?? []}
+          tokenGroup={expressionGroup ?? []}
```
preetriti1, line 270: Does this contain both function expressions and dynamic content ie outputs from previous nodes?

deschandhok24: good point let me check this.

PR #5547, opened 2024-08-29
feat(designer): Editable Code View

Eric-B-Wu:
Creating support for an editable code view

Currently there are no restrictions going from designer view -> code view
From code view -> designer view we run workflow validation. If it fails to pass validation we won't allow the switch

1. Updating an existing workflow
standardDeletion

2. Replacing a workflow
standardReplace

3. Also works for consumption
consumption


apps/Standalone/src/designer/app/AzureLogicAppsDesigner/CodeView.tsx
ccastrotrejo, line 1: If setting/retrieving the code is async, should we show a loading state? I am not super aware of how long can serializing the workflow can take

ccastrotrejo (approved):

apps/Standalone/src/designer/app/AzureLogicAppsDesigner/CodeView.tsx
Eric-B-Wu: yeah the main time will be serializing/validating the workflow from backend api, for portal there will be a notification that appears to tell the user that its currently validating

preetriti1:
@Eric-B-Wu - Please make sure you have validated the following scenarios -
1. Keep the search UI open while adding a action but do not add an action and then switch to codeview. See if no transient node is added in json to make it invalid or if it changes graph structure.
2. Add an action but do not create a connection yet, see if switching to codeview and designer works and also saving.
3. Keep some nodes in transient state for example - Add Appservice in consumption but do not select an operation yet and test the switching part back and forth.
4. Go to settings and try updating runAfter to invalid values check if graph structure is not invalid while switching back.

Basically we need to see if UI has incomplete layout or details how does switching back and forth works..

Eric-B-Wu:
> @Eric-B-Wu - Please make sure you have validated the following scenarios -
> 
> 1. Keep the search UI open while adding a action but do not add an action and then switch to codeview. See if no transient node is added in json to make it invalid or if it changes graph structure.
> 2. Add an action but do not create a connection yet, see if switching to codeview and designer works and also saving.
> 3. Keep some nodes in transient state for example - Add Appservice in consumption but do not select an operation yet and test the switching part back and forth.
> 4. Go to settings and try updating runAfter to invalid values check if graph structure is not invalid while switching back.
> 
> Basically we need to see if UI has incomplete layout or details how does switching back and forth works..

Following from offline messaging :
1. Seems to work fine
2. After going to code view we won't be able to switch back to designer view due to backend validation, as the error would be: "The API connection reference name must be provided in the 'Get_fifteen_day_of_weather_alarms' action inputs."
3. This will go to code view but because the node isn't added yet, it just continues from the graph before the adding of appservice action
4. When trying to create a circular dependency, backend validation of :The template validation failed: 'The circular dependency detected in template language expressions.

Discussed with Travis/Priti, we will handle this is in the portal by first having a dialog (with option to hide in the future) to warn users that going to code view may not be able to revert back to the designer view

PR #5585, opened 2024-09-03
fix(Data Mapper): generate XSLT error message and loop connection improvement

DanielleCogs:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## New Behavior
- now have error message toast if XSLT generation fails
- fixed loop connection logic for edge case
- added larger schema files for standalone perf testing
- show only name instead of full path for output from function

## Impact of Change


libs/data-mapper-v2/src/core/state/DataMapSlice.ts
```diff
@@ -1,5 +1,6 @@
 import type { ConnectionDictionary, InputConnection } from '../../models/Connection';
 import { directAccessPseudoFunctionKey, type FunctionData, type FunctionDictionary } from '../../models/Function';
+import type { Draft } from 'immer';
```
DanielleCogs, line 3: Temporary hack for build to pass: we have to import immer somewhere https://github.com/reduxjs/redux-toolkit/issues/1806

DanielleCogs: (I removed immer in the below file)

takyyon (approved):

Issue #5530, opened 2024-08-28
Logic App Standard Designer resetting index on ForEach loops, when scrolled out of screen


PedroAlmeidaMS:
### Describe the Bug with repro steps 

Create workflow

Add ForEach loop

Test

Select any iteration other than the first

Scroll down and back up

Iteration index is reset.

\-- Legacy Designer does not present this issue.

### What type of Logic App Is this happening in? 

Standard (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "For_each": {
                "actions": {
                    "Set_counter_": {
                        "inputs": {
                            "name": "counter",
                            "value": "@items('For_each')"
                        },
                        "type": "SetVariable"
                    }
                },
                "foreach": "@body('Parse_JSON')?['loops']",
                "runAfter": {
                    "Parse_JSON": [
                        "SUCCEEDED"
                    ]
                },
                "type": "Foreach"
            },
            "Initialize_counter": {
                "inputs": {
                    "variables": [
                        {
                            "name": "counter",
                            "type": "string"
                        }
                    ]
                },
                "runAfter": {
                    "Initialize_variable": [
                        "SUCCEEDED"
                    ]
                },
                "type": "InitializeVariable"
            },
            "Initialize_variable": {
                "inputs": {
                    "variables": [
                        {
                            "name": "json",
                            "type": "string",
                            "value": "{\n\t\"loops\": [\n\t\t{\n\t\t\t\"record\": \"1\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"2\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"3\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"4\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"5\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"6\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"7\"\n\t\t},\n\t\t{\n\t\t\t\"record\": \"8\"\n\t\t}\n\t]\n}"
                        }
[...]

PR #5586, opened 2024-09-03
fix(Designer): Fixed issue with loop pager index resetting, fixed operations reloading in loops

rllyy97:
## Main Changes

- Loop pager component index now only uses the index from the redux store
  - This fixes the issue of the pager component index being reset when rerendering on scroll
- Nodes within monitoring view loops now only fetch their data once
  - If you moved a node offscreen then back, or paged away then back, it would refetch the run data
  - This has been fixed and now only fetches once, then just uses the cached data when needed

Fixes https://github.com/Azure/LogicAppsUX/issues/5530

### DIFF
--- a/libs/designer/src/lib/ui/common/LoopsPager/LoopsPager.tsx
+++ b/libs/designer/src/lib/ui/common/LoopsPager/LoopsPager.tsx
@@ -1,12 +1,12 @@
 import constants from '../../../common/constants';
 import type { AppDispatch } from '../../../core';
-import { useActionMetadata, useRunInstance } from '../../../core/state/workflow/workflowSelectors';
+import { useActionMetadata, useNodeMetadata, useRunInstance } from '../../../core/state/workflow/workflowSelectors';
 import { setRunIndex } from '../../../core/state/workflow/workflowSlice';
 import { getForeachItemsCount } from './helper';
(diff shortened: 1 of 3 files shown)


libs/designer/src/lib/ui/common/LoopsPager/LoopsPager.tsx
```diff
@@ -75,24 +76,21 @@ export const LoopsPager = ({ metadata, scopeId, collapsed }: LoopsPagerProps) =>
...
   const onClickPreviousFailed: PageChangeEventHandler = (page: PageChangeEventArgs) => {
     const { prevFailedRepetition } = findPreviousAndNextFailed(page.value - 1);
     dispatch(setRunIndex({ page: prevFailedRepetition, nodeId: scopeId }));
-    setCurrentPage(prevFailedRepetition + 1);
   };
 
-  if (currentPage > forEachItemsCount) {
-    onPagerChange({ value: forEachItemsCount });
-  }
+  // if (currentPage > forEachItemsCount + 1) {
```
ccastrotrejo, line 91: NIT: comments

ccastrotrejo (approved):

ccastrotrejo (approved):

PR #5592, opened 2024-09-04
fix(designer-ui): Optimize file picker partitioning and fix overflow

ek68794998:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Issues with file picker:

- In Power Automate, the '...' breadcrumb sometimes does not show at all. **This is due to a bug in `react-breadcrumb` which was fixed in v9.0.4, while Power Automate currently uses v9.0.3. https://github.com/microsoft/fluentui/pull/30005**
- In Azure Logic Apps, browsing to a folder that would show the '...' breadcrumb can crash the page. (Mitigated with #5576)
- Following the mitigation for the above issue, only the first and last breadcrumb will ever show in the file picker header, even if there is room to show more.
- Long file names can render outside of the header container.

image

image

## New Behavior

image
***Note:** I wanted to get `text-overflow: ellipsis` working on the header, but doing so requires adding `overflow: hidden` to the individual breadcrumb `li` element, which breaks `useOverflowMenu` calculations. Therefore, the only option is to just cut the text off.*

image

e2e-picker

### DIFF
--- a/libs/designer-ui/src/lib/picker/helpers.ts
+++ b/libs/designer-ui/src/lib/picker/helpers.ts
@@ -1,5 +1,5 @@
 import type { TreeDynamicValue } from '@microsoft/logic-apps-shared';
-import { equals, isEmptyString, isString } from '@microsoft/logic-apps-shared';
+import { equals } from '@microsoft/logic-apps-shared';
 import { PickerItemType } from './types';
 
 export const filterAndSortItems = (
@@ -26,16 +26,6 @@ export const filterAndSortItems = (
     if (!a.isParent && b.isParent) {
       return 1;
     }
-    return getTreeItemDisplayName(a).localeCompare(getTreeItemDisplayName(b));
+    return a.displayName.localeCompare(b.displayName);
   });
-};
-
-export const getTreeItemDisplayName = (item: TreeDynamicValue): string => {
-  const possibleValues = [item.displayName, item.value?.DisplayName, item.fullyQualifiedDisplayName];
-  return possibleValues.find((v) => isString(v) && !isEmptyString(v)) || '';
-};
-
-export const getTreeItemId = (item: TreeDynamicValue): string => {
-  const possibleValues = [item.value?.Id, item.value, item.fullyQualifiedDisplayName];
-  return possibleValues.find((v) => isString(v) && !isEmptyString(v)) || '';
 };
--- a/libs/designer-ui/src/lib/picker/helpers.ts
+++ b/libs/designer-ui/src/lib/picker/helpers.ts
@@ -1,5 +1,5 @@
 import type { TreeDynamicValue } from '@microsoft/logic-apps-shared';
-import { equals } from '@microsoft/logic-apps-shared';
+import { equals, isEmptyString, isString } from '@microsoft/logic-apps-shared';
 import { PickerItemType } from './types';
 
 export const filterAndSortItems = (
@@ -26,6 +26,16 @@ export const filterAndSortItems = (
     if (!a.isParent && b.isParent) {
       return 1;
     }
-    return a.displayName.localeCompare(b.displayName);
+    return getTreeItemDisplayName(a).localeCompare(getTreeItemDisplayName(b));
   });
+};
+
+export const getTreeItemDisplayName = (item: TreeDynamicValue): string => {
+  const possibleValues = [item.displayName, item.value?.DisplayName, item.fullyQualifiedDisplayName];
+  return possibleValues.find((v) => isString(v) && !isEmptyString(v)) || '';
+};
+
+export const getTreeItemId = (item: TreeDynamicValue): string => {
+  const possibleValues = [item.value?.Id, item.value, item.fullyQualifiedDisplayName];
+  return possibleValues.find((v) => isString(v) && !isEmptyString(v)) || '';
 };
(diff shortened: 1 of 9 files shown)


libs/designer-ui/src/lib/picker/helpers.ts
```diff
@@ -26,6 +26,16 @@ export const filterAndSortItems = (
     if (!a.isParent && b.isParent) {
       return 1;
     }
-    return a.displayName.localeCompare(b.displayName);
+    return getTreeItemDisplayName(a).localeCompare(getTreeItemDisplayName(b));
   });
 };
+
+export const getTreeItemDisplayName = (item: TreeDynamicValue): string => {
+  const possibleValues = [item.displayName, item.value?.DisplayName, item.fullyQualifiedDisplayName];
```
preetriti1, line 34: Why are we doing this? You should never assume the structure of data coming back from api call.... 

All dynamic tree calls be it legacy for swaggers or manifest ones they should be setting the displayName correctly because the connectors need to define at which path the display name is expected so a callback was added in the picker to fetch that.

libs/designer-ui/src/lib/picker/helpers.ts
```diff
@@ -26,6 +26,16 @@ export const filterAndSortItems = (
...
+  return possibleValues.find((v) => isString(v) && !isEmptyString(v)) || '';
+};
+
+export const getTreeItemId = (item: TreeDynamicValue): string => {
+  const possibleValues = [item.value?.Id, item.value, item.fullyQualifiedDisplayName];
```
preetriti1, line 39: Same here.

libs/logic-apps-shared/src/designer-client-services/lib/connector.ts
```diff
@@ -12,6 +12,7 @@ export interface ListDynamicValue {
 export interface TreeDynamicValue {
   value: any;
   displayName: string;
+  fullyQualifiedDisplayName?: string;
```
preetriti1, line 15: What does this mean? what is displayName and what is fullyQualifiedDisplayName?? only one is displayed in UI so stamp it in displayName before sending back to designer from service

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/connector.ts
ek68794998: See test file for examples.

libs/designer-ui/src/lib/picker/helpers.ts
ek68794998: This is part of the problem with the fact that this code base has made such liberal use of `any`; we're left to make assumptions about the structure of data, especially because `TreeDynamicValue` has no top-level ID field.

If we want `TreeDynamicValue` to be a safe, strongly typed object, I can do that, but I didn't expect a change like that to be as well-received. Just let me know if that's preferred.

preetriti1: If the problem is to put an Id to make each item unique we can do that, I dont have any objection to that.

My object is editor UI understanding the structure of the api data coming in - TreeDynamicValue needs to be strongly typed and the client services (connector service) needs to map the result from api response to the TreeDynamicValue structure.. And this is the only structure shared with UI editor, so it always knows Id -> Id, displayName -> displayName

ek68794998: Fixed, moved `id` assignment to `ConnectorService`.

libs/logic-apps-shared/src/designer-client-services/lib/connector.ts
ek68794998: Since the test file is now removed, here's an example:

image

preetriti1: Do we still need the fullyQualified display name?

libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -219,6 +219,7 @@ export const getLegacyDynamicTreeItems = async (
     return {
       value,
       displayName: (getPropertyValue(value, titlePath as string) ?? '').toString(),
+      id: value.Id,
```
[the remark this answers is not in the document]
ek68794998: Currently `id` is set up to be optional, so it should be appropriately handled downstream if not present. However, you raise a good point here, and it would prevent issues with duplicate keys if multiple have undefined IDs. I will update that.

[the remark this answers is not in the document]
ek68794998: Fixed in 25fab9bd63cae28edeab1fdd7f3e03e6eb2b930b.

libs/logic-apps-shared/src/designer-client-services/lib/connector.ts
ek68794998: Only for completeness; it is referenced but only as an untyped string:

image

preetriti1: Then is the service reading it correctly and setting it? I dont see it getting used in any UI component

ek68794998: Yeah, it's read back here:

image

`libs\designer\src\lib\ui\panel\nodeDetailsPanel\tabs\parametersTab\index.tsx`

preetriti1 (approved):

PR #5597, opened 2024-09-05
fix(designer): Adding support for html dom and lexical characters

spanvalkar:
… - addressed the issue and the bug appears to be resolved now...

## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

[Bug 28769241](https://msazure.visualstudio.com/One/_workitems/edit/28769241): [Expression Editor] Expression is changed to gibberish

Based on the findings in the bug above, creating expressions with unsafe html DOM characters ('<' or '>') or unsafe html lexical characters ("& or '"') causes errors in the LAUX Expression Editor.

## New Behavior

The Expression Editor in LAUX now supports unsafe html DOM characters ('<' or '>') and unsafe html lexical characters ("& or '"') by encoding them as safe equivalents via the existing html string parsing functions:

1. decodeSegmentValueInDomContext
2. decodeSegmentValueInLexicalContext
3. encodeSegmentValueInDomContext
4. encodeSegmentValueInLexicalContext

For context, these encode/decode functions already existed in the LAUX codebase and worked properly, which is why I considered them safe to use here for parsing unsafe html characters.

## Impact of Change

Before: The unsafe characters above would lead to parsing issues in the Expression Editor in LAUX.
image

After: The unsafe characters are now parsed properly in the Expression Editor in LAUX by properly encoding/decoding them using the existing 4 encode/decode functions listed above. 
image

image


libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
```diff
@@ -298,6 +299,51 @@ export const decodeStringSegmentTokensInLexicalContext = (value: string, nodeMap
 export const encodeStringSegmentTokensInLexicalContext = (value: string, nodeMap: Map<string, ValueSegment>): string =>
   encodeOrDecodeStringSegmentTokens(value, nodeMap, decodeSegmentValueInLexicalContext, encodeSegmentValueInLexicalContext, 'encode');
 
+export const decodeStringSegmentTokensInDomAndLexicalContext = (value: string, nodeMap: Map<string, ValueSegment>): string =>
+  encodeOrDecodeStringSegmentTokensMultipleFunctions(
```
Eric-B-Wu, line 303: I'm wondering if splitting into encodeOrDecodeStringSegmentTokensMultipleFunctions is necessary, I'm not sure if this will get used other than within this context, and instead of using multiple decoders and encoders we could encode/decode logic inside the loop like:
```
const processStringSegmentTokensInDomAndLexicalContext= (
  value: string,
  nodeMap: Map<string, ValueSegment>,
  encode: boolean
): string => {
  return [...nodeMap.keys()].reduce((newValue, key) => {
    const encodedValue = encodeSegmentValueInLexicalContext(encodeSegmentValueInDomContext(key));
    const decodedValue = decodeSegmentValueInLexicalContext(decodeSegmentValueInDomContext(key));
    return newValue.replaceAll(encode ? decodedValue : encodedValue, encode ? encodedValue : decodedValue);
  }, value);
};
```

libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
```diff
@@ -60,10 +60,10 @@ export const parseHtmlSegments = (value: ValueSegment[], options?: SegmentParser
   const nodeMap = new Map<string, ValueSegment>();
 
   const stringValue = convertSegmentsToString(value, nodeMap);
-  const encodedStringValue = encodeStringSegmentTokensInDomContext(
-    encodeStringSegmentTokensInLexicalContext(stringValue, nodeMap),
-    nodeMap
-  );
+  const encodedStringValue = encodeStringSegmentTokensInDomAndLexicalContext(stringValue, nodeMap);
+  //const midEncodedStringValue = encodeStringSegmentTokensInLexicalContext(stringValue, nodeMap);
```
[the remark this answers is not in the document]
spanvalkar: Thanks for catching this. Removing it now. I thought I got all the comments, but I guess I missed one.

libs/designer-ui/src/lib/editor/base/utils/parsesegments.ts
spanvalkar: Fair point. Using the decode/encode functions together in a loop probably makes more sense. Making this change now.

Eric-B-Wu (commented):
please add a test, and will be good to go

Issue #5441, opened 2024-08-20
Modifying Control Condition Converts String Value to Int Automatically


krissybass:
### Describe the Bug with repro steps 

1) A previously-configured condition existed with multiple condition values. In the final condition, the value was set to compare a string value to a number stored as a string ("6"). The json showed that the number was surrounded by quotes.

![image](<https://github.com/user-attachments/assets/2da9846d-ab9a-4e7b-9b0d-4eaf92adb15d>)

![image](<https://github.com/user-attachments/assets/57277e1e-fb6d-4431-86e3-226624c0c9c3>)

2) The condition was modified (but not this row) and saved.

![image](<https://github.com/user-attachments/assets/c8bddbab-3aab-4287-bf6b-6d63398e9f4a>)

3) After saving the change, the existing condition which was previously a string was converted to an int (without quotes) in the rule (workflow.json)

![image](<https://github.com/user-attachments/assets/a9890279-09ca-437e-8036-2173e5df392e>)

4) There is no visible way to tell within the condition card whether the value on the right of the condition is a string or text, so this resulted in numerous workflow failures and issues evaluating the condition against an int instead of expected string.

### What type of Logic App Is this happening in? 

Standard (VSCode)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

VS Code

### Additional context 

*No response*

PR #5622, opened 2024-09-10
fix(designer): Deserializing Workflow not properly loading stringified primitives

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/5441 where on deserialization if we have a parameter of type ANY that had a value of a primitive wrapped in quotes, it would show/save in UI without quotes

- Some minor refactoring of the deserialization logic for condition editor
- In complex array editor to hide internal array elements


libs/designer-ui/src/lib/arrayeditor/index.tsx
```diff
@@ -34,6 +34,7 @@ export interface ArrayItemSchema {
   format?: string;
   enum?: string[];
   readOnly?: boolean;
+  ['x-ms-visibility']: string;
```
preetriti1, line 37: PLease use the swagger constant instead of coding here, this is to make sure extension names are used from single place. check schemaprocessor

preetriti1 (changes requested):

preetriti1:
Please add some unit tests for the functions which is making the conversion.... include some basic scenario and the failing user scenario...
It will be helpful to avoid regressions.

libs/designer/src/lib/core/utils/__test__/segment.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    rawModeEnabled: true,
+  });
+  it('properly converts string segment to Value Segment', () => {
+    const stringValue = 'test';
+    const result = convertor.convertToValueSegments(stringValue, constants.SWAGGER.TYPE.STRING);
+    expect(result[0].value).toEqual(createLiteralValueSegment(stringValue).value);
+  });
+
+  it('properly wraps string boolean with quotes if of type any', () => {
+    const stringValue = 'true';
```
preetriti1, line 17: What happens when we want to input number or boolean will it always be enclosed in quotes?

libs/designer/src/lib/core/utils/__test__/segment.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { beforeAll, describe, expect, it } from 'vitest';
+import { createLiteralValueSegment, ValueSegmentConvertor, ValueSegmentConvertorOptions } from '../parameters/segment';
+import constants from '../../../common/constants';
+
+describe('convertToValueSegments', () => {
+  const convertor = new ValueSegmentConvertor({
```
preetriti1, line 6: Is this convertor only used in conditions? I am worried this might change other fields

libs/designer/src/lib/core/utils/parameters/segment.ts
```diff
@@ -106,12 +107,14 @@ export class ValueSegmentConvertor {
...
 
-  private _convertStringToValueSegments(value: string): ValueSegment[] {
+  private _convertStringToValueSegments(value: string, parameterType: string): ValueSegment[] {
     if (isTemplateExpression(value)) {
       const expression = ExpressionParser.parseTemplateExpression(value);
       return this._convertTemplateExpressionToValueSegments(expression);
     }
-    return [this._createLiteralValueSegment(value)];
+    const isSpecialValue = ['true', 'false', 'null'].includes(value) || /^-?\d+$/.test(value);
+    const stringValue = parameterType === constants.SWAGGER.TYPE.ANY && isSpecialValue ? `"${value}"` : value;
```
preetriti1, line 116: Please be careful of the cases when it is any field and user really wants to input non string

Eric-B-Wu: This is for the function convertStringToValueSegment, where we first do a check of the value to be of type string. If the user has a different primitive it would not go into this logic.

preetriti1 (approved):

PR #5629, opened 2024-09-10
fix(Designer): Hide Unsupported Actions/Connectors for Standard

### LINKED ISSUE #5619: Hide Unsupported Actions/Connectors for Standard

When seraching for actions, all actions including the ones that are not supported in standard shows. As a user, you will be able to pick this action and may get confused due to the errors since it will not work.

Consumption (Portal)

### Which operating system are you using? 

Windows

New Designer

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

image

image

### Browser 

All

### Additional context 

*No response*

Elaina-Lee:
## Requirement Checklist

* [X] Tests for the changes have been added (for bug fixes or features)

## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

* hides gmail connectors / api operations for standard
fixes #5619 

https://github.com/user-attachments/assets/49a32c60-aa84-43fd-993f-c54d4e4b5a78


libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
```diff
@@ -171,7 +178,8 @@ export abstract class BaseSearchService implements ISearchService {
     const uri = `/subscriptions/${subscriptionId}/providers/Microsoft.Web/locations/${location}/managedApis`;
     // const responseArray = await this.batchAzureResourceRequests(uri);
     const responseArray = await this.getAzureResourceRecursive(uri, undefined);
-    return this.moveGeneralInformation(responseArray);
+    const movedGeneralResponseArray = this.moveGeneralInformation(responseArray);
+    return this.removeUnsupportedConnectors(movedGeneralResponseArray);
```
preetriti1, line 182: Filter unsupported and then move the general information

libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
```diff
@@ -209,8 +217,7 @@ export abstract class BaseSearchService implements ISearchService {
       };
       // const response = await this.batchAzureResourceRequests(uri, queryParameters);
       const response = await this.getAzureResourceRecursive(uri, queryParameters);
-
-      return response;
+      return this.removeUnsupportedOperations(response);
```
preetriti1, line 220: Don't do it in custom apis.. those are user deployed and managed and no need to run a iteration on it when we should not filter here. 
We should only do for managed apis

libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
```diff
@@ -228,9 +235,8 @@ export abstract class BaseSearchService implements ISearchService {
       };
       // const response = await this.batchAzureResourceRequests(uri, queryParameters);
       const response = await this.getAzureResourceRecursive(uri, queryParameters);
-
       const locationFilteredResponse = response.filter((connector: any) => equals(connector.location, location));
-      return locationFilteredResponse;
+      return this.removeUnsupportedConnectors(locationFilteredResponse);
```
preetriti1, line 239: Same remove from here too

libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
```diff
@@ -248,6 +254,14 @@ export abstract class BaseSearchService implements ISearchService {
     return connectors;
   }
 
+  public removeUnsupportedOperations(operations: DiscoveryOpArray): DiscoveryOpArray {
+    return operations.filter((operation) => !this._unsupportedConnectorIds.includes(operation?.properties?.api?.id));
```
preetriti1, line 258: I would suggest to normalize the connectorIds in your list and from the api.id to do the includes.. because it is a resource I'd and even if they have different casing they would be same resource so you need to be case insensitive here.

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
Elaina-Lee: Can you help clarify 'normalizing the connectorIds in your list and from the api.id' - what would you mean by normalize here, to Lowercase those?

libs/logic-apps-shared/src/designer-client-services/lib/base/search.ts
```diff
@@ -208,9 +217,7 @@ export abstract class BaseSearchService implements ISearchService {
         $filter: `properties/trigger eq null and type eq 'Microsoft.Web/customApis/apiOperations' and ${ISE_RESOURCE_ID} eq null`,
       };
       // const response = await this.batchAzureResourceRequests(uri, queryParameters);
-      const response = await this.getAzureResourceRecursive(uri, queryParameters);
-
-      return response;
+      return await this.getAzureResourceRecursive(uri, queryParameters);
```
preetriti1, line 220: you can just do return this.getAzureResourceRecursive no need for await here

libs/logic-apps-shared/src/utils/src/lib/helpers/__test__/connectors.spec.tsx
```diff
@@ -210,4 +212,50 @@ describe('utils/src/lib/helpers/connectors', () => {
...
+        '',
+        '/serviceProviders/Sql',
+      ]);
+      expect(normalizeConnectorIds([...armConnectorIds, '', '/serviceProviders/sql'], subscriptionId, location, true)).toEqual([
+        expectedConnectorId.toLowerCase(),
+        expectedConnectorId2.toLowerCase(),
+        '',
+        '/serviceProviders/Sql'.toLowerCase(),
+      ]);
+    });
```
preetriti1, line 259: This should be in a separate describe since this is a difference function normalizeConnectorIds

preetriti1 (approved):

PR #5633, opened 2024-09-10
Feat(Data Mapper): Allow LML generation with missing func params, and surface XSLT error message

DanielleCogs:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

unable to save LML if all function inputs are not filled

## New Behavior

able to save LML with missing function params, but add error message if XSLT generation fails

## Impact of Change


libs/data-mapper-v2/src/components/commandBar/EditorCommandBar.tsx
```diff
@@ -56,10 +58,11 @@ export const EditorCommandBar = (_props: EditorCommandBarProps) => {
     (state: RootState) => state.modal.warningModalType === WarningModalState.DiscardWarning && state.modal.isOkClicked
   );
 
-  const dataMapDefinition = useMemo<string>(() => {
+  const dataMapDefinition = useMemo<MetaMapDefinition>(() => {
```
DanielleCogs, line 61: Not really using this yet, but need to get consensus from the team on how we handle this scenario long-term before removing! Don't want to duplicate any work

takyyon (approved):

PR #5661, opened 2024-09-13
fix(Designer): Additional changes for aliasing with split on

Aleks-gdb:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
Currently the aliasing for spliton outputs is not applied to schemaAlias which causes a missing alias property in dynamic schema calls on PAuto side.

## New Behavior

The change introduces a plug-in from manifest service to handle the spliton outputs conversion - the function is optional and falls back to existing logic if not provided.

## Impact of Change

Unblocks PAuto, should have no effect on Logic Apps


libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -248,6 +259,89 @@ const convertSchemaAliasesForSplitOn = (schema: OpenAPIV2.SchemaObject): void =>
...
+      dynamicState: { ...originalDynamicState, schemaAlias: schema[ExtensionProperties.Alias] },
+    };
+  }
+}
+
+function convertDynamicTreeExtension(schema: OpenAPIV2.SchemaObject): void {
+  const dynamicTreeExtension = schema[ExtensionProperties.DynamicTree];
+
+  if (!dynamicTreeExtension) {
+    return;
```
preetriti1, line 309: Shouldnt this be done at the service layer? We do not want to update the schema in designer, I know this is present in manifestService today because backend doesnt do the conversion, but this was kept in service because eventually backend needs to do this conversion and send it to client.

So please do not add this code here.

Aleks-gdb: Since the code for spliton lives in the designer, should we make the function getUpdatedManifestForSplitOn as a whole host-provided? A good chunk of the code (including incomplete aliasing code for spliton) lives here - is the ask to extract it all out to host?

Aleks-gdb: Pushed change to rely on host for updating the manifest or fall back to the current logic if not provided. This could likely be optimized but scoping the changes in the interest of unblocking PAuto on this issue ASAP.

libs/designer/src/lib/core/utils/outputs.ts
```diff
@@ -200,7 +200,12 @@ export const getUpdatedManifestForSplitOn = (manifest: OperationManifest, splitO
     }
 
     const manifestItems: OpenAPIV2.SchemaObject | undefined = clone(manifestSection.items);
-    const updatedManifestItems = isAliasPathParsingEnabled ? getUpdatedAliasInItemProperties(manifestItems) : manifestItems;
+
+    let updatedManifestItems = manifestItems;
+    if (isAliasPathParsingEnabled && updatedManifestItems) {
```
preetriti1, line 205: I think you should be checking isAliasPathParsingEnabled && manifestItems && OperationManifestService().getUpdatedAliasInOutputIt....) { 
.... 
}

libs/logic-apps-shared/src/designer-client-services/lib/operationmanifest.ts
```diff
@@ -36,6 +36,13 @@ export interface IOperationManifestService {
    * @return {Promise<any>}
    */
   getOperationManifest(connectorId: string, operationId: string): Promise<OperationManifest>;
+
+  /**
+   * Gets the transformed output property body items with aliases
+   * @arg {OpenAPIV2.SchemaObject | undefined} schemaItem - The output property body items.
+   * @return {OpenAPIV2.SchemaObject | undefined}
+   */
+  getUpdatedAliasInOutputItemProperties?(schemaItem: OpenAPIV2.SchemaObject | undefined): OpenAPIV2.SchemaObject | undefined;
```
preetriti1, line 45: PLease name this as - getUpdatedOutputsForSplitOn

This way your method can take full control ovber the manifest schema which may or maynot include for item properties.

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/operationmanifest.ts
Aleks-gdb: Not sure I follow here - is it just a matter of function rename or increasing the scope of the service function? I would rather not duplicate code in the host if not necessary unless LA is not using most of the split on logic. To my understanding the aliasing logic was the bit LA wasn't using.

libs/designer/src/lib/core/utils/outputs.ts
Aleks-gdb: Does that mean the manifest service function can replace the existing function?

libs/logic-apps-shared/src/designer-client-services/lib/operationmanifest.ts
Aleks-gdb: Renamed the function

libs/designer/src/lib/core/utils/outputs.ts
Aleks-gdb: Does logic apps not use the aliasing logic at all? Is it safe to replace with a host service plug in?

rllyy97:
Closing, PA has forked designer

PR #5672, opened 2024-09-13
chore(designer): Revert dynamic data aliasing changes in v4.86

ek68794998:
Our proposed fix (#5661) for the aliasing issues in Power Automate was not approved, so additional work is required before we can take aliasing changes.

This cherry-pick PR is mostly the same as #5621 but for v4.86. Looks like LA is using v4.85 and v4.88 but not v4.86 so this should not affect LA portal.

> This is a hotfix PR containing two reverts:
> 
> * #5537 (254a51ad644576335ea3422130cce4f79c32b06c)
> * #5360 (2e27773cd24e5a67fd33a1656a789b3ec0d8c10a)
> 
> Unfortunately, the change to aliasing has broken us in other ways that were not fixed by our initial hotfix for v4.74.1. Therefore, we are reverting that hotfix, as well as the initial dynamic data aliasing change, until we can figure out how to fix the scenario entirely.

Also reverting unit test changes made for the above PRs (#5565).

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/__test__/initialize.spec.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/__test__/initialize.spec.ts
@@ -1,4 +1,3 @@
-import { InitOperationManifestService } from '@microsoft/logic-apps-shared';
 import { getInputParametersFromManifest } from '../initialize';
 import {
   mockGetMyOffice365ProfileOpenApiManifest,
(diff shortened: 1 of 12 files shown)


ek68794998:
Ran into merge conflicts making this, so am doing a quick local test pass to make sure the revert works as expected.

Eric-B-Wu (approved):

PR #5686, opened 2024-09-16
fix(Designer): Fixed issue with `suppressDefaultNodeSelectFunctionality`

rllyy97:
## Main Changes

When the `suppressDefaultNodeSelectFunctionality` designer option is enabled we just hide the node details panel.
This was old behavior that got changed/removed during the panel refactor PA did.
The `suppressDefaultNodeSelectFunctionality` option is used in BPT to avoid opening the panel during our token selection over in BPT and this was noticed as broken upon updating.


ccastrotrejo (approved):

takyyon (approved):

ek68794998 (approved):
Thanks for catching this!

Issue #5639, opened 2024-09-11
There is no way to paste an action to a new parallel branch


ek68794998:
### Describe the Bug with repro steps 

1. Have any workflow with a trigger and two consecutive actions. 
2. Copy the last action. 
3. Click a (+) node. 

Expected: You should be able to paste the action to a new parallel branch.

 Actual: You can create a new parallel branch with a *new* action, or paste the action consecutively, but you cannot paste to a new parallel branch.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

![cant-paste-parallel](<https://github.com/user-attachments/assets/93b85678-67c0-4203-b461-e2d7495895df>)

### Browser 

Edge, Version 128.0.2739.67 (Official build) (64-bit).

### Additional context 

*No response*

PR #5689, opened 2024-09-16
feat(Designer): Support Paste Parallel Action

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior
- There is no way of pasting an action to a parallel branch

## New Behavior
- When copied operation exists, it will show the paste action with a selection for submenu which includes parallel branch option
- The user can still select the main menu selection for paste an action without having to select it from the submenu
- For parallel branch selection, it will insert the pasted operation as a parallel branch

## Impact of Change

image

https://github.com/user-attachments/assets/fa2b9976-5027-4933-8864-63a72ca1b037


libs/designer/src/lib/core/state/operation/operationSelector.ts
```diff
@@ -22,14 +22,18 @@ export const getNodeOperationData = (state: OperationMetadataState, nodeId: stri
...
+    operationMetadata: getNodeOperationMetadata(state, nodeId),
     settings: getRecordEntry(state.settings, nodeId),
     staticResult: getRecordEntry(state.staticResults, nodeId),
     actionMetadata: getRecordEntry(state.actionMetadata, nodeId),
     repetitionInfo: getRecordEntry(state.repetitionInfos, nodeId),
   };
 };
 
+export const getNodeOperationMetadata = (state: OperationMetadataState, nodeId: string) => {
+  return getRecordEntry(state.operationMetadata, nodeId);
```
Eric-B-Wu, line 34: is there any reason why this was moved into its own function?

Elaina-Lee: For copy scope action, I had to add the nodeData - but only the operation metadata part was needed so didn't want to store the rest of unused data

ccastrotrejo (approved):

Elaina-Lee:
Fixes https://github.com/Azure/LogicAppsUX/issues/5639

PR #5695, opened 2024-09-17
feat(designer): Add panel action changes to undo/redo history

shuktika12163:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior
Only workflow related actions like add, delete and move node can be undone.

## New Behavior
* Some panel related actions like user updating parameter values, user updating panel title, static results can also be undone now.
* replaceId function is now dispatched on title blur to avoid storing state on each keystroke and undoing one letter at a time.

## Impact of Change


libs/designer/src/lib/core/state/undoRedo/undoRedoSlice.ts
```diff
@@ -1,33 +1,57 @@
...
+      // Add current edited panel details to state being added to history
+      const newStateItem: StateHistoryItem = {
+        compressedState: action.payload.compressedState,
+        editedPanelTab: state.currentEditedPanelTab,
+        editedPanelNode: state.currentEditedPanelNode,
+      };
+
+      // Update current edited panel details to the one from the state we are undoing to
+      // This ensures we don't lose track of edited panel details on undo/redo click when state is updated
+      state.currentEditedPanelTab = state.past[state.past.length - 1].editedPanelTab;
```
rllyy97, line 34: Shouldn't this `editedPanelTab` already be in the panel state?  I'm confused on why we need to do this manually, if we load a compressed state the selected panel tab + node should be loaded as well

rllyy97: Discussed offline

rllyy97 (approved):

PR #5733, opened 2024-09-23
fix(designer): Adding support for output tokens in case nodes

Eric-B-Wu:
Creating support for output tokens for switch cases based on available tokens for switch

addSwitch


libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
```diff
@@ -89,6 +94,7 @@ export const ParametersTab: React.FC<PanelTabProps> = (props) => {
   const errorInfo = useOperationErrorInfo(selectedNodeId);
   const { hideUTFExpressions } = useHostOptions();
   const replacedIds = useReplacedIds();
+  const parentIdOfSwitch = useGetSwitchParentId(selectedNodeId);
```
rllyy97, line 97: We already have this info in the NodeMetadata in redux, on line 80 `nodeMetadata.parentNodeId`

rllyy97 (changes requested):

ccastrotrejo (approved):

rllyy97 (approved):

Issue #5737, opened 2024-09-24
Logic app designer not loading in VS Code


giancarloQ:
### Describe the Bug with repro steps 

Once I open the designer for the workflow .json the error below appears:

settingValue.startsWith is not a function

### What type of Logic App Is this happening in? 

Standard (VSCode)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "Check_Guru_Api_Proxy": {
                "type": "Scope",
                "actions": {
                    "HTTP_Guru_Api_Proxy_Healthcheck": {
                        "type": "Http",
                        "inputs": {
                            "uri": "@replace(appsetting('GuruApiProxyUrl'), '/proxy/', '/healthcheck')",
                            "method": "GET",
                            "headers": {
                                "x-functions-key": "@{appsetting('GuruApiProxyFunctionKey')}"
                            }
                        },
                        "runtimeConfiguration": {
                            "contentTransfer": {
                                "transferMode": "Chunked"
                            }
                        }
                    },
                    "Set_variable_canConnectGuruApi": {
                        "type": "SetVariable",
                        "inputs": {
                            "name": "canConnectGuruApiProxy",
                            "value": true
                        },
                        "runAfter": {
                            "HTTP_Guru_Api_Proxy_Healthcheck": [
                                "SUCCEEDED"
                            ]
                        }
                    }
                },
                "runAfter": {
                    "Check_Junifer_API": [
                        "SUCCEEDED",
                        "TIMEDOUT",
                        "SKIPPED",
                        "FAILED"
                    ]
                }
            },
            "Check_Junifer_API": {
                "type": "Scope",
                "actions": {
                    "HTTP_Junifer_Bearer_Token": {
                        "type": "Http",
                        "inputs": {
                            "uri": "@{appsetting('JuniferApiUrl')}token",
                            "method": "POST",
                            "authentication": {
                                "password": "@{appsetting('JuniferApiPassword')}",
                                "type": "Basic",
                                "username": "@{appsetting('JuniferApiUsername')}"
                            }
                        },
[...]

Issue #5736, opened 2024-09-24
settingValue.startsWith is not a function


schenDisid:
### Describe the issue 

Go to *workflow*, Open *Designer* and shows **settingValue.startsWith** is not a **function** error from Azure Logic Apps

### Are you using new designer or old designer 

New Designer (With action panels)

### Screenshots or Videos 

*No response*

### Browser 

- VSCode extension 

### Additional context 

*No response*

Eric-B-Wu:
This is likely happening due to an app setting not being a string, I'll create a type check for this

FlipFlop17:
I have the same issue, cant open my workflow
![image](https://github.com/user-attachments/assets/41ebed40-66c9-4344-9b8f-00cbe2cdd810)

schenDisid:
I fixed it in my workflow changing type "LogicAppsNodePort": "5000" to string because before I was putting it as a number (5000). It's weird tho, at first I put 5000 and it is working but suddenly it started showing that issue.

FlipFlop17:
never heard of LogicAppsNodePort setting o.O in your workflow ??

schenDisid:
"LogicAppsNodePort": "5000". Used to execute javascript code using node.js lib (version 16.X.X). It must be declared to be listened on port 5000 while i execute js code. That was my 'main' trouble as fast as i fixed it, it disappeared.

PR #5738, opened 2024-09-24
fix(designer): Add Type check when resolving non-string app settings

Eric-B-Wu:
Type check to make sure appsettings are strings so that we don't run into the issue:
settingValue.startsWith is not a function

https://github.com/Azure/LogicAppsUX/issues/5737
https://github.com/Azure/LogicAppsUX/issues/5736

### DIFF
--- a/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Utilities/Workflow.ts
+++ b/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Utilities/Workflow.ts
@@ -77,7 +77,7 @@ export class WorkflowUtility {
       for (const settingName of Object.keys(appsettings)) {
         const settingValue = appsettings[settingName] !== undefined ? appsettings[settingName] : '';
         // Don't replace if the setting value is a KeyVault reference
-        if (settingValue.startsWith('@Microsoft.KeyVault(')) {
+        if (typeof settingValue === 'string' && settingValue.startsWith('@Microsoft.KeyVault(')) {
           continue;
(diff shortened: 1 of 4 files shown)


apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Utilities/Workflow.ts
```diff
@@ -77,7 +77,7 @@ export class WorkflowUtility {
       for (const settingName of Object.keys(appsettings)) {
         const settingValue = appsettings[settingName] !== undefined ? appsettings[settingName] : '';
         // Don't replace if the setting value is a KeyVault reference
-        if (settingValue.startsWith('@Microsoft.KeyVault(')) {
+        if (typeof settingValue === 'string' && settingValue.startsWith('@Microsoft.KeyVault(')) {
```
hartra344, line 80: instead of typeof, you could also do this
```suggestion
        if (settingValue.startsWith && settingValue.startsWith('@Microsoft.KeyVault(')) {
```

hartra344 (approved):

apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Utilities/Workflow.ts
Eric-B-Wu: yeah I actually am thinking instead of resolving these non-string app settings, thinking we just not resolve them altogether because it might run into an error later on in the replace

PR #5741, opened 2024-09-24
feat(Templates): Templates Tab for Single/Multi Workflows

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

Added All, Workflows (Single), Accelerators (Multi) with working functaionlities

image


libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
```diff
@@ -78,6 +83,53 @@ export const TemplateFilters = ({ detailFilters }: TemplateFiltersProps) => {
...
+  ];
+
+  const onTabSelected = (e?: SelectTabEvent, data?: SelectTabData): void => {
+    if (data) {
+      const itemKey = data.value as string;
+      dispatch(
+        setDetailsFilters({
+          filterName: 'Type',
+          filters:
+            itemKey === templateDefaultTabKey
```
ccastrotrejo, line 120: I have some questions here:
- Is there any specific reason on why we storing the `displayName`? If yes I think once the user selects a tab the displayName will be overwritten by the itemKey, which is the data.value. This works fine for `Accelerator` but for `Workflow` and `all` which value is different than the displayname. Maybe I am wrong
- I think making a decision whehter is the default is just extra logic, by the end of the day the itemkey is equals to templateDefaultTabKey

ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/filters/templateFilters.tsx
Elaina-Lee: Thanks for addressing those! Appreciate these feedback and those are good points to think about -
- I can see the reason why it gets confusing whether we need the displayName here. The reason why we store displayName is because this dispatch call of setDetailsFilters is re-used from the dropdown. The dropdown items are "FilterObject" which is consisted of value and displayName. So this way, we can just pass the items without having to map to manually remove displayName. In this case, would you suggest that I modify this dispatch call to just take the values and map the filterItems to accomodate? (To clarify, displayName is not used within the state calls)
image
- Would you help me clarify why making a decision whether is the default is just extra logic? If the itemKey is 'all' (templateDefaultTabKey), then we have to remove the filters for "Type" by feeding it 'undefined'. From the logic of selections, itemKey will be one of: 'all', 'Workflow', and 'Accelerator'.
- Are you suggesting I change the key for the All case to be undefined? To clarify, the reason why I did not do this is because: if we do this, then key={undefined} and I thought this is not the best practice.
image

ccastrotrejo: ok ok no I think now it all makes sense.

1. ok ok, makes sense if we are reusing it. I was just curious about the overwritten of displayName 
2. For this one I guess I just assumed that this was the only place where the `appliedDetailFilters` was used. I guess dispatching it as `undefined` makes sense if you check whether is undefined or not to show all the templates somewhere else

Elaina-Lee: Got it, thank you. I will get these merged then!

For 2, just to clarify, we don't check if it's undefined or not to show all the templates; if a filter is unpresent, we dont apply the filter. Like if the search keyword is unpresent (undefined), then we would display all values. If the search keyword is present, we will filter out the results with those keywords.

PR #5747, opened 2024-09-25
fix(Designer): Remove extra property decoding on input serialization

Aleks-gdb:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
Currently, if the input parameter contains ~1 it is being serialized with replacing ~1 with a . which breaks runtime.

The issue happens for both Power Automate and Logic Apps. Below is logic apps view.

image
image
image

## New Behavior

This change adds an optional parameter to the function to skip the decoding that happens as the value has already been decoded before getting passed in so that the key isn't further modified by replace functions.

The value remains unchanged and LA runs successfully:
image
image

## Impact of Change

This could impact existing logic depending on the double decoding.


libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -542,7 +542,7 @@ const serializeParameterWithPath = (
     let p = result;
     while (pathSegments.length > 0) {
       const pathSegment = pathSegments.shift() as Segment;
-      const propertyKey = getAndEscapeSegment(pathSegment);
+      const propertyKey = getAndEscapeSegment(pathSegment, /** decodeSegment*/ false); // Disable decoding as it already happens with parseEx(valueKey) above
```
preetriti1, line 545: Is there an issue with how we are double decoding and it is decoding an actual property name?

preetriti1 (changes requested):

preetriti1 (approved):

PR #5754, opened 2024-09-25
feat(Templates): Add Services for Consumption to Support Different Exit Points

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

- Adding/seperating services for consumption to be able to support different behaviors
- Starting point to differentiate the exit point for openBladeAfterCreate


hartra344 (approved):

libs/logic-apps-shared/src/designer-client-services/lib/consumption/template.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseTemplateService } from '../base/template';
+
+export class ConsumptionTemplateService extends BaseTemplateService {
+  public async getExistingWorkflowNames(): Promise<string[]> {
+    return [];
+  }
```
preetriti1, line 6: What is this? We will have to implement a method that would check the validity of the name. For consumption it will be making a call to get 404 and for standard it would be from existing workflow names

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/consumption/template.ts
Elaina-Lee: Since we removed the basic tab for consumption (we won't let users type names, so this won't be used to check for duplicates), it would make sense to return an empty array. This doesn't have anything to do with the validation of the name since it's only checking for the duplicates. In this case, why should we make a call to retrieve 404?

Elaina-Lee: Understood; that is a good point, this function should not exist at all.

With this function removed, ConsumptionTemplateService has no other functions or variables at all. I made BaseTemplateService from abstract to regular class and used that in case of consumption (in the future, this can be used for both consumption & hybrid).

preetriti1 (approved):

PR #5760, opened 2024-09-26
feat(templates): Data model and store changes for multi-workflow template support

preetriti1:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

Should not change any current UX change, this is the foundation for upcoming UX changes for multi workflow templates


libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -11,10 +11,11 @@ import { ConnectorWithDetails } from '../../../../templates/connections/connecto
 import type { TemplatePanelTab } from '@microsoft/designer-ui';
 import { clearTemplateDetails } from '../../../../../core/state/templates/templateSlice';
 import Markdown from 'react-markdown';
+import { useDefaultWorkflowTemplate } from '../../../../../core/state/templates/templateselectors';
 
 export const SummaryPanel: React.FC = () => {
   const intl = useIntl();
-  const { manifest } = useSelector((state: RootState) => state.template);
+  const { manifest } = useDefaultWorkflowTemplate() ?? {};
```
rllyy97, line 18: Should we supply a default inside the `useDefaultWorkflowTemplate` hook?  I see you writing `?? {}` a lot

rllyy97 (approved):

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
preetriti1: Because initially this UI is called when there are no templates selected. I need to make sure this does not initialize when template is not selected, then i can remove this.
And this code will be changed when we implement multi workflow, this is to make sure the current single workflow is not broken with new changes.

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -8,9 +8,17 @@ export interface Manifest {
   skus: SkuType[];
   kinds?: WorkflowKindType[];
   details: Record<string, string>;
+
+  /* This is a markdown to show features for multi-workflow and details in case of single workflow */
   detailsDescription?: string;
+
   tags?: string[];
   artifacts: Artifact[];
```
Elaina-Lee, line 16: From our last discussion, we decided that artifacts will not be present for multi-workflow manifest. Should we consider making this an optional field?

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -8,9 +8,17 @@ export interface Manifest {
...
+
+  /* This consists of list of workflows listed in the multi-workflow template.
+  The key is the folder name, followed by metadata where name is default name to be used for creation */
+  workflows?: Record<string, { name: string }>;
+
   images: Record<string, string>;
   prerequisites?: string;
   parameters: Parameter[];
```
Elaina-Lee, line 24: From our last discussion, we decided that parameters/connections will not be present for multi-workflow manifest. Should we consider making these optional fields?

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -0,0 +1,214 @@
...
+import type { RootState } from '../../state/templates/store';
+import type { TemplateServiceOptions } from '../../templates/TemplatesDesignerContext';
+
+export interface WorkflowTemplateData {
+  name?: string;
+  workflowDefinition: LogicAppsV2.WorkflowDefinition | undefined;
+  manifest: Template.Manifest | undefined;
+  workflowName: string | undefined;
+  kind: string | undefined;
+  images?: Record<string, any>;
```
Elaina-Lee, line 33: Are we storing images here? Wondering why the Record values are `any` instead of `string`

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
preetriti1: No, we can have artifacts at workflow app level, hence i decided to not make that change... lets see what we get once we do zipdeploy

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
preetriti1: I just copied from previous data model, we should change it to string... will do that

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/nameStateTab.tsx
```diff
@@ -129,11 +132,11 @@ export const NameStatePanel = () => {
         ariaLabel={intlText.WORKFLOW_NAME}
         value={existingWorkflowName ?? workflowName}
         onChange={(_event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) =>
-          dispatch(updateWorkflowName(newValue))
+          dispatch(updateWorkflowName({ id: workflowId, name: newValue as string }))
```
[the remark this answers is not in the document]
preetriti1: While looking at how validations are invoked actually in store slice, we migt have to tweak this but can do it later after a discussion.

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/usePanelTabs.tsx
```diff
@@ -57,11 +57,11 @@ export const useCreateWorkflowPanelTabs = ({ onCreateClick }: { onCreateClick: (
     }
     if (!isConsumption && selectedTabId && selectedTabId !== Constants.TEMPLATE_PANEL_TAB_NAMES.BASIC) {
       if (!existingWorkflowName) {
-        dispatch(validateWorkflowName(existingWorkflowNames ?? []));
+        dispatch(validateWorkflowName({ id: id as string, existingNames: existingWorkflowNames ?? [] }));
```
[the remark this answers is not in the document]
preetriti1: We need to fix the code where we keep initializing the tabs when there is no template selected. It is because of this I had to add {} when there is no workflow.. I will fix that in a separate bug fix.. so we do not have to come to a place to doing ?? '' because this code should never hit when there is no workflow selected.
if this fails it is a bug in our code and we should fix it. There is no harm in assuming id will be a string here.

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
preetriti1: I am still not sure we should mark that as optional because again these are app level artifacts, over time when we do implementation for multi workflow we can change this, but can keep it open for now.

Issue #5201, opened 2024-07-25
New logic app is making duplicate connector swagger api calls


Xiaoyu-Huang:
### Describe the Bug with repro steps

For my team, we are using axios-cancel with request id to avoid the duplicate requests at same time. It causes issue on action panel rendering intermittently due to cancelled connector api request. I believe this might be a regression introduced starting form 4.28 as the older version we used don't have this issue.

I have verified on Azure portal by loading a random workflow. the connector API calls are called twice unnecessary. 

![image](https://github.com/user-attachments/assets/f018ee8d-25d1-489d-8eed-2bc300cd56ef)

### What type of Logic App Is this happening in?

Consumption (Portal)

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Browser

Edge

PR #5764, opened 2024-09-26
fix(Designer): Removed duplicated api requests

rllyy97:
## Main Changes

- Wrapped a lot of data request functions into queries
  - We were duplicating a lot of things like connector / swagger requests
  - Now they each should only request once, then get reused
  - Some we just had some query key issues that have been fixed
- Removed `useConnectorOnly` hook, now we just have the `useConnector` hook
- Split up the connector and swagger requests in the the connection service
  - These were just two separate requests being bundled together for some reason
  - They are now fully separate and you can call just one or the other instead of calling both and only using one or the other
- Some loading situations will now seem faster since we aren't waiting on duplicate requests

BREAKING CHANGES
- Removed `getConnectorAndSwagger` method from the Connection service
  - These were just two separate requests being bundled together for some reason
  - Somewhat replaced by the `getSwaggerFromConnector` method to just get a connector swagger 
  - Getting just the connector is still done by the `getConnector` method

Fixes https://github.com/Azure/LogicAppsUX/issues/5201


hartra344 (approved):

libs/designer/src/lib/core/queries/connections.ts
```diff
@@ -12,13 +13,9 @@ export interface ConnectorWithParsedSwagger {
 }
 
 export const getConnectorWithSwagger = async (connectorId: string): Promise<ConnectorWithParsedSwagger> => {
-  const { connector, swagger } = await getReactQueryClient().fetchQuery(['apiWithSwaggers', connectorId.toLowerCase()], async () => {
-    const { connector, swagger } = await ConnectionService().getConnectorAndSwagger(connectorId);
-    const parsedSwagger = await SwaggerParser.parse(swagger);
```
preetriti1, line 17: this is not a time consuming operation because this is not a api call, so it might not add any change

libs/designer/src/lib/core/queries/connections.ts
```diff
@@ -12,13 +13,9 @@ export interface ConnectorWithParsedSwagger {
 }
 
 export const getConnectorWithSwagger = async (connectorId: string): Promise<ConnectorWithParsedSwagger> => {
-  const { connector, swagger } = await getReactQueryClient().fetchQuery(['apiWithSwaggers', connectorId.toLowerCase()], async () => {
```
preetriti1, line 15: Since we removed the queryClient, we will have to make sure whoever calls this need to call through queryClient.. even hosts can call this.

ccastrotrejo (approved):

libs/logic-apps-shared/src/designer-client-services/lib/connection.ts
```diff
@@ -38,7 +38,7 @@ export interface CreateConnectionResult {
 
 export interface IConnectionService {
   getConnector(connectorId: string): Promise<Connector>;
-  getConnectorAndSwagger(connectorId: string): Promise<ConnectorWithSwagger>;
+  getSwaggerFromConnector(connectorId: string): Promise<OpenAPIV2.Document>;
   getSwaggerFromUri(uri: string): Promise<OpenAPIV2.Document>;
```
preetriti1, line 42: You need to mark this as breaking change so other teams can update their connection service

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/connection.ts
rllyy97: Yep I noted that this was a breaking change up in the description

libs/designer/src/lib/core/queries/connections.ts
rllyy97: The main change here was that `getConnector` and `getSwagger` will use the respective query caches, instead of having a third cache for `apiWithSwaggers` with them combined

libs/designer/src/lib/core/queries/connections.ts
rllyy97: This still going through a queryClient, the `getConnector` and `getSwagger` calls do it internally

Issue #5768, opened 2024-09-26
Foreach: Filter array action is getting into Advanced mode and Edit in basic mode is in disabled state.


vgouth:
### Describe the Bug with repro steps 

1\.Create a consumption logic app.

 2\.Click on edit and open designer.

 3\.Add http request trigger.

 4\.Add Dropbox - List files in folder action.

 5\.Add filter array action.

 6\.Add DisplayName token from Dropbox in Filter array parameter which adds implicit foreach.

 7\.In from parameter select body parameter.

 8\.Switch between code view and designer view

9. Filter Query parameter in filter array action is getting into Advanced mode and Edit in basic mode is in disabled state. 

**Expected:** Filter Query parameter in filter array action is getting into Advanced mode and Edit in basic mode should be in enable state.

**Actual :** Filter Query parameter in filter array action is getting into Advanced mode and Edit in basic mode is in disabled state.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

![edit in basic mode](<https://github.com/user-attachments/assets/dde28a71-1fe2-4d7c-85de-cdd9af57b344>)

[https://github.com/user-attachments/assets/c95c0885-6186-40e2-b606-f976a4a1e292](<https://github.com/user-attachments/assets/c95c0885-6186-40e2-b606-f976a4a1e292>)

### Browser 

Edge

### Additional context 

**Version:** 0\.2022.1.764

PR #5772, opened 2024-09-26
fix(designer): Simple Query Builder not deserializing

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/5768


hartra344 (approved):

libs/logic-apps-shared/src/utils/src/lib/helpers/functions.ts
ccastrotrejo, line 1: I know this is a really simple function, can we add some tests for it?

libs/designer-ui/src/lib/querybuilder/Group.tsx
```diff
@@ -183,7 +183,7 @@ export const Group = ({
   ];
 
   const handleUpdateNewParent = (newState: GroupItemProps | RowItemProps, currIndex: number) => {
-    const newItems = { ...groupProps };
+    const newItems = { ...groupProps, items: [...groupProps.items] };
```
ccastrotrejo, line 186: NIT: can this be `items: groupProps.items` since we are destructuring the array and creating a new array at the same time? or maybe I don't see a specific scenarios

Eric-B-Wu: This causes the array to be non-extensible which prevents elements from being added, it runs into the error of:
`Cannot add property 0, object is not extensible` and is unable to add any changes the user mades

ccastrotrejo (approved):

PR #5779, opened 2024-09-27
feat(Templates): Support Pagination + Pager UI

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

- Added pageCount constant & pageNumber in manifest state as part of filters to apply pagination logic
- Added params for Pager in designer-ui to re-use page handling logic & partial UI
- With the added params, Pager component now supports optional clickable page numbers & display total numbers

**Below is example UI when templatesCountPerPage is set to 10 to show pagination effect, currently is set to 25**

image
image
image


Elaina-Lee:
Not ready for review - attempting to fix tests

libs/designer-ui/src/lib/pager/pager.less
```diff
@@ -1,5 +1,36 @@
 @import (reference) '../variables.less';
 
+.msla-pager-v2-pagenumbers {
+  display: flex;
+  flex-direction: row;
```
ccastrotrejo, line 5: NIT: row is the default value for `display: flex`, there is no need to add it exta

ccastrotrejo (approved):

libs/designer-ui/src/lib/pager/index.tsx
```diff
@@ -186,44 +216,109 @@ export const Pager: React.FC<PagerProps> = ({
...
 
+  const pageNumbers = useMemo(() => {
+    const result = [];
+
+    // Calculate initial range around the current number
+    let rangeStart = Math.max(current - 2, min);
+    let rangeEnd = Math.min(current + 2, max);
+
+    // Adjust the range if it's less than 5 numbers
+    while (rangeEnd - rangeStart + 1 < 5) {
```
preetriti1, line 227: I am assuming 5 is the default page number you want to show.. can you keep this as a constant or take that as a prop, i am assuming based on the caller and teh area they have they can determine how much to show..
since this is UI element lets try to move this to props so it is configurable.

Elaina-Lee: Based on the portal Pager UI that we are using as a reference, it did not provide the flexibility to adjust the number of pages shown. I thought there would be a good reason for that, but since this is our custom component, I can move it as a prop with a default value of 5

PR #5792, opened 2024-10-01
feat(templates): Overview for multi workflow templates

preetriti1:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

templatemain


libs/designer/src/lib/core/state/templates/templateselectors.ts
ccastrotrejo, line 1: Github actions comments

libs/designer/src/lib/ui/templates/templateoverview.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  // TODO: Need to open new create panel for multi workflow here.
+  const footerContentProps = workflowTab(intl, dispatch, true, () => setShowCreatePanel(true), {
+    templateId: templateName ?? '',
+    workflowAppName,
+  }).footerContent;
+  return (
+    <>
+      <QuickViewPanelHeader title={title} description={description} details={info} features={detailsDescription} />
+      <div className="msla-template-overview" style={{ marginTop: '-34px' }}>
```
ccastrotrejo, line 64: Nit: can we put the `margin-top` style inside the class?

ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/templateoverview.tsx
Elaina-Lee: Nit: Would this be the best way to handle the styling? Negative marginTop and marginBottom seem to be not the best unless absolutely necessary. There should be better ways to handle this, for instance, if the marign is coming from QuickVIewPanelHeader, we should remove margin from the QuickViewPanelHeader.

You can look at CreateWorkflowPanel for reference. This removes default margin given by the react markdown component.

image

Otherwise, once merged, I can come back to this css afterwards as well.

Elaina-Lee (approved):

libs/designer/src/lib/ui/templates/templateoverview.tsx
preetriti1: I think I need to discuss this, on how we can share the single workflow components here, specifically the common tabs in panel... I can provide explicit css in header component that way we do not have to do it...
I am going to file an item and track the css changes through it, for now will add a TODO in code so we can revisit when functionality is done.

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -123,8 +128,9 @@ const loadTemplateFromResourcePath = async (templateName: string, manifest: Temp
   const templateManifest: Template.Manifest =
     manifest ?? (await import(`./../../templates/templateFiles/${templateName}/manifest.json`)).default;
```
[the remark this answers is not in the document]
preetriti1: Do you mean in the state?

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -212,3 +218,18 @@ const loadWorkflowTemplateFromManifest = async (
     return undefined;
   }
 };
+
+const getWorkflowAndManifest = async (templatePath: string, manifest: Template.Manifest | undefined) => {
+  const paths = templatePath.split('/');
+  const templateManifest: Template.Manifest =
+    manifest ?? paths.length === 2
+      ? (await import(`./../../templates/templateFiles/${paths[0]}/${paths[1]}/manifest.json`)).default
```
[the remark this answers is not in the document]
preetriti1: It wont be out of bounds no, because we are checking paths.length == 2 so it has to contain two segments isnt it?

libs/designer/src/lib/core/state/templates/templateselectors.ts
preetriti1: Fixed those tests.

PR #5797, opened 2024-10-02
feat(templates): Adjusting panel code to be reused for single and multi workflow creation

preetriti1:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other


libs/designer/src/lib/core/state/templates/templateSlice.ts
```diff
@@ -35,6 +35,26 @@ export const templateSlice = createSlice({
         state.workflows[id].workflowName = name;
       }
     },
+    validateWorkflows: (state, action: PayloadAction<{ validateName: boolean; existingNames: string[] }>) => {
```
Elaina-Lee, line 38: so this is for basic tab validation (name and kind) only right - if so, could we consider naming it more specific?

This gives me the idea that this is validating the workflows in general including the other parts (connections, parameters)

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -144,7 +144,7 @@ export const summaryTab = (
     }),
     secondaryButtonOnClick: () => {
       dispatch(closePanel());
-      if (showCreate) {
+      if (clearDetailsOnClose) {
```
Elaina-Lee, line 147: I see what this is doing here; only for the multi-workflow scenario, we clear the details. 

Before, for single workflow, we wiped the details in the state; so when we run "changeCurrentTemplateName" then "loadTemplate", it won't cause flickering since it goes from _no data_ to _loaded data_.

Since here, the state for the previously selected tab is preserved, I am worried that the flickering may happen for loading a new single template to display info for.

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
```diff
@@ -25,29 +24,24 @@ export const TemplatePanel = ({ createWorkflow, onClose, showCreate, workflowId
...
-    connections: state.workflow.connections,
-    isConsumption: state.workflow.isConsumption,
-    connectionsError: state.template.errors.connections,
-    parametersError: state.template.errors.parameters,
-    parameterDefinitions: state.template.parameterDefinitions,
+    manifest: state.template.manifest,
+    workflows: state.template.workflows,
   }));
-  const { manifest, workflowName, kind, errors, workflowDefinition } = useWorkflowTemplate(workflowId);
+  const isMultiWorkflowTemplate = useMemo(() => Object.keys(workflows).length > 1, [workflows]);
```
Elaina-Lee, line 33: Just a consideration, is it better to define it here and pass it around as params? Why are we picking this over storing this at the templateSlice state level?

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
```diff
@@ -154,7 +114,7 @@ export const TemplatePanel = ({ createWorkflow, onClose, showCreate, workflowId
       {currentPanelView === 'createWorkflow' ? (
         <CreateWorkflowPanel panelTabs={createWorkflowPanelTabs} />
       ) : currentPanelView === 'quickView' ? (
-        <QuickViewPanel workflowId={workflowId} showCreate={showCreate} />
+        <QuickViewPanel workflowId={workflowId as string} clearDetailsOnClose={showCreate} />
```
Elaina-Lee, line 117: `undefined as string` will be `"undefined"`, is this what we want? Otherwise, could we change it to `workflowId ?? ''`

Elaina-Lee (approved):
Not blockers, but concerns that could be resolved or explained

libs/designer/src/lib/core/state/templates/templateSlice.ts
preetriti1: Shall I name it validateWorkflowsNameAndKind or validateWorkflowsBasicInfo (so we are clear this is validating the basic tab details)

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
preetriti1: This is only going to clear the details when caller passes it, for multi workflow we are not clearing the details because it is being opened from templates overview page when we click on single workflow, for everything else the caller should be passing in true

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
preetriti1: No, when we are in QuickViewPanel workflowId should always be present, If it is not it is a bug and we will know that, putting workflowid to empty might mask it

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
preetriti1: templateSlice already has workflows in its cache, and isMultiWorkflow can be inferred from there, I dont think this is a costly operation and this only gets executed when template panel is being created since this is in memory, so when template card is selected or when use template is selected.

Is your concern about the repeated calculation?

libs/designer/src/lib/core/state/templates/templateSlice.ts
Elaina-Lee: Yep I think that'd be ideal, I would prefer validateWorkflowsBasicInfo (in case of any further changes) since here we are aiming to validate everything in basic tab details

libs/designer/src/lib/ui/panel/templatePanel/templatePanel.tsx
Elaina-Lee: My concern was that isMultiWorkflowTemplate is being passed as a param to places such as useCreateWorkflowPanelTabs

Elaina-Lee: I was expressing concern for that value being passed as a param when it exists or could exist in the template state

Elaina-Lee: But since it's only being passed twice as a param, if you think that's okay, wouldn't be a big deal

preetriti1: Lets change it as we need when we are into tabs implementation

PR #5800, opened 2024-10-02
feat(Designer): Introduce experimentation to enable A/B testing of features

takyyon:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Impact of Change

### DIFF
--- a/libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
+++ b/libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
@@ -1,4 +1,4 @@
-import { AssertionErrorCode, AssertionException } from 'utils/src';
+import { AssertionException, AssertionErrorCode } from '../../utils/src';
 
 export interface IExperimentationService {
   isFeatureEnabled: (featureGateName: string, type: 'extension' | 'shell') => Promise<boolean>;
--- a/libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
+++ b/libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
@@ -1,0 +1,24 @@
+import { AssertionErrorCode, AssertionException } from 'utils/src';
+
+export interface IExperimentationService {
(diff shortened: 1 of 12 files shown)


libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+let service: IExperimentationService;
+
+export const InitExperimentationServiceService = (experimentationService: IExperimentationService): void => {
+  service = experimentationService;
+};
+
+export const ExperimentationService = (): IExperimentationService => {
+  if (!service) {
```
ccastrotrejo, line 16: Just to be extra sure about this. The designer doesn't crash or throws this error if the service is not needed and not initilialized, right? for example in vscode extension

ccastrotrejo (approved):

libs/logic-apps-shared/src/designer-client-services/lib/experimentation.ts
takyyon: It shouldn't. The designer will only crash in scenarios where the experimentation service is being used but not initialized. Each client using the designer needs to initialize the service. On the side I'll also be checking to see how we can always default to a certain value value here

takyyon: Actually, I went ahead and added it explicitly for vscode as well. We need a better way to default certain service if the value remains the same for all default experiences.

takyyon: On a second thought, I went ahead and always defaulted to a false implementation of ExperimentationService so incase a certain client using the designer does not want to use the A/B test feature, they won't get it unless they explicitly add in the experimentation service

ccastrotrejo: Thanks Krrish!

PR #5803, opened 2024-10-03
fix(vscode): hybrid logic apps bug fixes 

anandgmenon:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Impact of Change

### DIFF
--- a/apps/vs-code-designer/src/app/commands/createLogicApp/createLogicAppSteps/HybridLogicAppsSteps/ContainerAppNameStep.ts
+++ b/apps/vs-code-designer/src/app/commands/createLogicApp/createLogicAppSteps/HybridLogicAppsSteps/ContainerAppNameStep.ts
@@ -11,15 +11,16 @@ import type { ContainerAppsAPIClient } from '@azure/arm-appcontainers';
 
 export class ContainerAppNameStep extends AzureWizardPromptStep<ILogicAppWizardContext> {
   public async prompt(context: ILogicAppWizardContext): Promise<void> {
-    const validationError = await ContainerAppNameStep.validateName(context, context.newSiteName);
+    const nameAvailabiltyValidationError = await this.validateNameAvailable(context, context.newSiteName);
+    const nameValidationError = ContainerAppNameStep.validateInput(context.newSiteName);
 
-    if (!validationError) {
+    if (!nameAvailabiltyValidationError && !nameValidationError) {
       return;
     }
 
     const prompt: string = localize(
       'newLogicAppName',
-      `Enter a new logic app name, ${validationError}`,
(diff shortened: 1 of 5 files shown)


apps/vs-code-designer/src/app/commands/createLogicApp/createLogicAppSteps/HybridLogicAppsSteps/ContainerAppNameStep.ts
```diff
@@ -11,24 +11,23 @@ import type { ContainerAppsAPIClient } from '@azure/arm-appcontainers';
...
       context.newSiteName,
       context.resourceGroup.name
     );
 
     context.newSiteName = (
       await context.ui.showInputBox({
         prompt,
-        validateInput: ContainerAppNameStep.validateInput,
-        asyncValidationTask: (name: string) => this.validateNameAvailable(context, name),
+        asyncValidationTask: (name: string) => ContainerAppNameStep.validateName(context, name),
```
ccastrotrejo, line 30: Is there any specific reason on why we are not validating the name again if there is a validationError?
For example:

- User first put a name that ends up in validationError above. 
- So it enters again to ask the user for the name
- How does the user will know if the new name does or doesn't exists already in the resource group?

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/commands/createLogicApp/createLogicAppSteps/HybridLogicAppsSteps/ContainerAppNameStep.ts
anandgmenon: oh I out the validateInput logic inside the `validateName` method, so it will actually do that

ccastrotrejo: Yep I noticed that, but for the new inputBox, do we validate it while the user is typing the new name? thats what I mean

anandgmenon: oh yeah I was checking that what's the difference in async and without, I see it now. I'll update

ccastrotrejo: Sure, thanks!

ccastrotrejo (approved):

Issue #5824, opened 2024-10-07
Swagger/HTTP task does not serialize


simonlaszcz:
### Describe the Bug with repro steps 

when the retry policy is 'none':

Workflow validation failed for the workflow ''. {"error":{"code":"InvalidRequestContent","message":"The request content is not valid and could not be deserialized: 'Could not find member 'retryPolicy' on object of type 'FlowTemplateAction'. Path 'properties.definition.actions.Task\_', line 1, position 1817.'."}}

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Chrome

### Additional context 

*No response*

simonlaszcz:
Any idea when this fix will be deployed?

Thanks

rllyy97:
Hi @simonlaszcz, it will be a part of our next release cycle, unfortunately we started one yesterday so the next one will be releasing between Oct. 28th and Nov. 1st

PR #5828, opened 2024-10-07
fix(Designer): Fixed HTTP+Swagger retry policy serialization issue

rllyy97:
## Main Changes

Our HTTP+Swagger action had it's inputs location set to the action root, so the retry policy was being loaded into the action root when it should be in the "actions".  This has been added to the action's swap map so the retry policy ends up in the right place.
The swapmap execution has also been moved to after the retry policy being set, so that the retry policy can be altered.

Fixes https://github.com/Azure/LogicAppsUX/issues/5824

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
@@ -306,7 +306,13 @@ const serializeManifestBasedOperation = async (rootState: RootState, operationId
   const nodeStaticResults = getRecordEntry(rootState.operations.staticResults, operationId) ?? ({} as NodeStaticResults);
   const inputPathValue = serializeParametersFromManifest(inputsToSerialize, manifest);
   const hostInfo = serializeHost(operationId, manifest, rootState);
-  const inputs = hostInfo !== undefined ? mergeHostWithInputs(hostInfo, inputPathValue) : inputPathValue;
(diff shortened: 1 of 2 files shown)


ccastrotrejo (approved):

libs/designer/src/lib/core/actions/bjsworkflow/serializer.ts
```diff
@@ -306,7 +306,13 @@ const serializeManifestBasedOperation = async (rootState: RootState, operationId
   const nodeStaticResults = getRecordEntry(rootState.operations.staticResults, operationId) ?? ({} as NodeStaticResults);
   const inputPathValue = serializeParametersFromManifest(inputsToSerialize, manifest);
   const hostInfo = serializeHost(operationId, manifest, rootState);
-  const inputs = hostInfo !== undefined ? mergeHostWithInputs(hostInfo, inputPathValue) : inputPathValue;
+  let inputs = hostInfo !== undefined ? mergeHostWithInputs(hostInfo, inputPathValue) : inputPathValue;
+  const retryPolicy = getRetryPolicy(nodeSettings);
+  if (retryPolicy) {
+    inputs.retryPolicy = retryPolicy;
+  }
+  inputs = swapInputsLocationIfNeeded(inputs, manifest.properties.inputsLocationSwapMap);
```
preetriti1, line 314: I didnt understand previously also retry policy was getting removed before swapINputs, could you explain what else needs to be happen now after swapInputs?

preetriti1 (changes requested):

PR #5848, opened 2024-10-09
feat(Templates): Updating Parameters Tab to Grid View to Align with Multi Workflow

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

Updating parameters tab to new design, aligning with multi workflow to display associated worklfow

image
**Sorted by Name**
image
**Multi workflow (Showing Associated workflows)**
image
**After the tab has been moved to review+create and back (validation for the tab triggered)**
image


Elaina-Lee:
5 mins - removing unwanted changes

libs/designer-ui/src/lib/flyout/index.tsx
```diff
@@ -25,7 +26,7 @@ const defaultTooltipHostStyles: ITooltipHostStyles = {
 };
 
 export const Flyout = React.forwardRef<{ collapseFlyout(): void }, FlyoutProps>(
-  ({ ariaLabel, iconStyle, style, tabIndex = 0, text, tooltipHostStyles, onClick, title }, ref) => {
+  ({ ariaLabel, iconStyle, style, tabIndex = 0, text, tooltipHostStyles, onClick, title, iconSize = 'lg' }, ref) => {
```
preetriti1, line 29: was lg teh default size?

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -43,6 +43,7 @@ export interface Parameter {
 
 export interface ParameterDefinition extends Parameter {
   value?: any;
+  associatedWorkflows?: string[];
```
preetriti1, line 46: Why do we have this list here? The manifest itself wont have any mention of this no?
We will have to determine based on the parameter key

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
```diff
@@ -1,37 +1,199 @@
...
+        onRenderItemColumn={onRenderItemColumn}
+        selectionMode={SelectionMode.none}
+      />
+    </div>
+  );
+};
+
+const copyAndSort = (items: Template.Parameter[], columnKey: string, isSortedDescending?: boolean): Template.Parameter[] => {
+  const keyPath =
+    columnKey === '$name' ? ['connectorDisplayName'] : columnKey === '$status' ? ['hasConnection'] : ['connection', 'displayName'];
```
preetriti1, line 189: Pease change this.

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
```diff
@@ -1,37 +1,199 @@
...
+
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <div className="msla-templates-parameters-tab">
+      <DetailsList
+        setKey="id"
```
preetriti1, line 176: Do you have id property in parameters?

preetriti1 (changes requested):

libs/designer-ui/src/lib/flyout/index.tsx
Elaina-Lee: The size that was being used was 28px, I wasn't sure what else to label this as we want something significantly smaller.

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
Elaina-Lee: We could do that, but if we go that path, then everytime the parametersTab renders, we'd have to
   for every parameter
      for every manifest
         check if this parameter name exists, then added it to associatedWorkflows to display

I thought keeping it here as part of loading parameter process will be more runtime efficient

preetriti1: We already have a way to do that.. from teh list of workflows in stores, we need to find each parameter exists in which workflow.parameterKeys... we dont have to look into manifest at all... i kept it that way for this reason

Elaina-Lee: resolved it as discussed offline

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -193,7 +192,25 @@ const loadTemplateFromResourcePath = async (templateName: string, manifest: Temp
         data.workflows[workflowPath] = workflowData.workflow;
         data.parameterDefinitions = {
           ...data.parameterDefinitions,
-          ...workflowData.parameterDefinitions,
+          ...Object.keys(workflowData.parameterDefinitions).reduce((acc: Record<string, Template.ParameterDefinition>, key: string) => {
```
preetriti1, line 195: I think this is going to run multiple times for shared parameters as well... so when you get the tests fixed, please do move this calculation after the for loop is done on the parameterDefinitions, so it only runs once for each unique parameter.

libs/designer/src/lib/ui/panel/templatePanel/__tests__/nameStateTab.spec.tsx
```diff
@@ -22,10 +22,32 @@ describe('panel/templatePanel/createWorkflowPanel/nameStateTab', () => {
...
+            kinds: [],
+            artifacts: [
+              {
+                type: 'workflow',
+                file: 'workflow.json',
+              },
+              {
+                type: 'description',
+                file: 'description.md',
+              },
```
preetriti1, line 41: does this still exist?

preetriti1 (approved):

Issue #5871, opened 2024-10-11
Connections: New connections are not being updated in Change connections without a page reload.


vgouth:
### Describe the Bug with repro steps 

1\.Create a consumption-based logic app.

 2\.Open designer.

 3, Add request trigger.

 4\.Add Microsoft Translator V2 - Get languages action/office 365/ SQL Server/ Share point/ Slack/ Dropbox (any action).

5. Create a new connection with required fields filled 
6. After new connection established, we can see it is connected to new established connection. 
7. Try and change connection. 
8. You should be able to see the new connection along with the existing connections (if any). 

**Expected:** We should see new connections and existing connections (if any) when trying to change the connections. 

**Actual:**We can see only existing connections (if any) but not new connections. But we could find the new connection once we reload the browser. It is not updating without reload.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

[https://github.com/user-attachments/assets/3f28bc74-0291-42ec-b524-0a2e1612662a](<https://github.com/user-attachments/assets/3f28bc74-0291-42ec-b524-0a2e1612662a>)

### Browser 

Edge

### Additional context 

**Version:** 2\.41007.1.5

rllyy97:
This could be related to this PR: https://github.com/Azure/LogicAppsUX/pull/5764

PR #5873, opened 2024-10-11
fix(Designer): Fixed connections query id issue

rllyy97:
## Main Changes

Adjusted query client key to be lowercase to match other instances.
The mismatched casing was causing new connections to not be included in the connections data.

Fixes https://github.com/Azure/LogicAppsUX/issues/5871


Localize/lang/strings.json
ccastrotrejo, line 1: should we keep this?

rllyy97: The strings file was added with one of Priti's PRs, not sure what it is about but it got removed by my commit tasks
https://github.com/Azure/LogicAppsUX/pull/5866

Eric-B-Wu (approved):

Localize/lang/strings.json
preetriti1: Just keep it... not sure why that got created as part of my changes I didn't update it because that was automatically included

preetriti1 (approved):

PR #5878, opened 2024-10-14
meta(Templates): Modify Download Template Script to Include Multi-Workflow

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [ ] Feature
* [ ] Other

Adding as part of script to download all workflows (folders) under the multi workflow folder
 + Minor wording changes to align to content review

image


Localize/lang/strings.json
ccastrotrejo, line 1: Maybe just uncheck this file? or is there any reason to delete all of it?

ccastrotrejo (approved):

Localize/lang/strings.json
Elaina-Lee: not sure how that got slipped in haha, thanks for catching. Removed the file

Elaina-Lee: It seems like it gets added as part of the check-in process, not sure how we can remove it

PR #5885, opened 2024-10-15
fix(Designer): Custom connector triggers now show in search

rllyy97:
## Main Changes

Adjust our custom connector operation request to include triggers

image


libs/logic-apps-shared/src/designer-client-services/lib/standard/search.ts
```diff
@@ -100,7 +100,7 @@ export class StandardSearchService extends BaseSearchService {
       const uri = `/subscriptions/${subscriptionId}/providers/Microsoft.Web/locations/${location}/apiOperations`;
       const queryParameters: QueryParameters = {
         'api-version': apiVersion,
-        $filter: `properties/trigger eq null and type eq 'Microsoft.Web/customApis/apiOperations' and ${ISE_RESOURCE_ID} eq null`,
+        $filter: `type eq 'Microsoft.Web/customApis/apiOperations' and ${ISE_RESOURCE_ID} eq null`,
```
preetriti1, line 103: We need to change both the base service and standard service

preetriti1 (dismissed):

libs/logic-apps-shared/src/designer-client-services/lib/standard/search.ts
rllyy97: Yep the file didn't get added in the commit for some reason, fixing now

ccastrotrejo (approved):

PR #5899, opened 2024-10-17
fix(designer): Fixed various bugs in undo/redo

shuktika12163:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

List of bugs being tackled here (in review/active) - each commit is associated with each fix:
https://msazure.visualstudio.com/One/_queries/query/a6e3b020-844f-409b-9bf5-5f5163419d4a/

1. Parameter panel re-render edge case: [Bug 29820667](https://msazure.visualstudio.com/One/_workitems/edit/29820667): [V3 undo redo bug bash] Undoing a change in the panel without reopening the panel does not reflect the state update

2. Delete operations:
- [Bug 29821282](https://msazure.visualstudio.com/One/_workitems/edit/29821282): [V3 undo redo bug bash] Deleting a scope adds an "invisible" undo step (i.e., you can undo one additional time after undoing deleting the scope)
- [Bug 29821013](https://msazure.visualstudio.com/One/_workitems/edit/29821013): [V3 undo redo bug bash] Removing condition or scope undo-es each remove individually

3. Optional parameter conditional visibility: [Bug 29857755](https://msazure.visualstudio.com/One/_workitems/edit/29857755): [V3 undo redo bug bash] Removing optional parameter is skipping a state in between

4. Manual trigger/floating action menu:
- [Bug 29821430](https://msazure.visualstudio.com/One/_workitems/edit/29821430): [V3 undo redo bug bash] Manual trigger text input undoes 1 keystroke at a time
- [Bug 29851270](https://msazure.visualstudio.com/One/_workitems/edit/29851270): [V3 undo redo bug bash] Floating action menu outputs editor does not save state
- [Bug 29900468](https://msazure.visualstudio.com/One/_workitems/edit/29900468): [V3 undo redo bug bash] Manual trigger undo doesn't work

5. Title update issues
- Custom people picker editor causing re-render in parameters panel causing title updates without any changes: [Bug 29820626](https://msazure.visualstudio.com/One/_workitems/edit/29820626): [V3 undo redo bug bash] Undo/redo doesn't work when there are changes in people picker editor
- Clicking in and out of title was saving a state even without any changes

## New Behavior

Above bugs have been fixed:
1. Parameter re-render fix:
parameterRerender

2. Delete operations fix:
deleteOperations

3. Optional parameter conditional visibility fix:
optionalParameter

4. Manual trigger/floating action menu fixes:
floatingActionMenu

5. People picker/title change fix:
Clicking in and out of title doesn't save a new state. It only saves state when value changes
titleFix

## Impact of Change


libs/designer/src/lib/core/state/undoRedo/undoRedoTypes.ts
```diff
@@ -19,7 +17,8 @@ import {
 export interface StateHistory {
   past: StateHistoryItem[];
   future: StateHistoryItem[];
-  stateHistoryItemIndex: number;
+  // Toggle whenever undo/redo is clicked. This is needed to re-render parameter panel on undo/redo
```
rllyy97, line 20: Do we not need the state history item index anymore?

shuktika12163: No, we just need the key to change when undo/redo is clicked so that parameter panel re-renders. Toggling the key between 0 and 1 works and avoids edge case issues due to us not updating the item index on state save (since we don't want to re-render panel every time we save a state, only when we click undo/redo)

libs/designer/src/lib/ui/settings/settingsection.tsx
```diff
@@ -332,6 +333,7 @@ const Setting = ({
     };
 
     const removeParamCallback = () => {
+      dispatch(storeStateToUndoRedoHistory({ type: updateParameterConditionalVisibility.type }));
```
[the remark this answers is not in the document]
shuktika12163: Difference in behaviour in that case:
1. If user clicks "show all" or "remove all", it'll save a state for each optional parameter, so it'll undo/redo parameters visibility one at a time instead of state before and after button clicks.
2. When user selects multiple options in the dropdown for optional parameters, it'll store state for each one and undo/redo them individually instead of multiple params together

Would this be preferable instead?

rllyy97 (approved):

PR #5956, opened 2024-10-25
fix(designer): Limit inputs/outputs content size to 2MB

### LINKED ISSUE #5938: Run History page isn't responding when actions' input/output size is large

Run History page isn't responding when actions' input/output size is large, especially the Compose action.

For example, this workflow reads 8 mb data from blob storage, and composes it in a json.

Image

After the unresponding alert shows up, the designer will crash.

Image

Previously, actions do not render large size input/ouput. They only render a button to open the full input/output on another page.

Image

Consumption (Portal)

### Which operating system are you using? 

Windows

New Designer

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

- Browser: [Edge, Chrome] 

### Additional context 

*No response*

ccastrotrejo:
This pull request includes several changes to the `libs/designer` and `libs/logic-apps-shared` libraries, primarily focused on improving the handling of undefined inputs and outputs, as well as adding size checks for content in run services. The most important changes include updating binders to return undefined inputs and outputs as-is, modifying the `getChildWorkflowIdFromInputs` function, and adding a size check for content in the `ConsumptionRunService` and `StandardRunService` classes.

*  Modified `getChildWorkflowIdFromInputs` to use a different property path and return the workflow value.
* Added a size check to return undefined if content size exceeds 2MB in `ConsumptionRunService`.
* Added a size check to return undefined if content size exceeds 2MB in `StandardRunService`.

Fixes #5938

### DIFF
--- a/libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
+++ b/libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
@@ -17,7 +17,7 @@ describe('DefaultInputsBinder', () => {
 
   it('should return an empty object when inputs are undefined', () => {
     const result = binder.bind(undefined);
-    expect(result).toEqual({});
+    expect(result).toEqual(undefined);
   });
 
   it('should handle null inputs gracefully', () => {
--- a/libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
+++ b/libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
@@ -15,7 +15,7 @@ describe('DefaultInputsBinder', () => {
     spy?.mockRestore();
   });
 
-  it('should return an empty object when inputs are undefined', () => {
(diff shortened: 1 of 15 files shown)


libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
```diff
@@ -17,7 +17,7 @@ describe('DefaultInputsBinder', () => {
 
   it('should return an empty object when inputs are undefined', () => {
```
Eric-B-Wu, line 18: we should probably change the description of these tests "should return undefined when inputs are undefined"

Eric-B-Wu (approved):

libs/designer/src/lib/core/utils/monitoring/binders/__test__/inputs/default.test.ts
ccastrotrejo: ohh sure, I will update them. Thanks!

PR #6004, opened 2024-10-31
fix(Designer): Fixed workflow ID path for getting child workflow

rllyy97:
## Main Changes

Adjusted workflow id path for getting child workflow.
The workflow id path during binding was changed, and this reference was not updated.

### DIFF
--- a/libs/designer/src/lib/ui/panel/nodeDetailsPanel/nodeDetailsPanel.tsx
+++ b/libs/designer/src/lib/ui/panel/nodeDetailsPanel/nodeDetailsPanel.tsx
@@ -224,11 +224,11 @@ export const NodeDetailsPanel = (props: CommonPanelProps): JSX.Element => {
     return undefined;
   };
 
-  const getChildWorkflowIdFromInputs = (inputs: any): string | undefined => {
-    if (!isNullOrEmpty(inputs)) {
-      const workflow = getObjectPropertyValue(inputs, ['host', 'value', 'workflow']);
+  const getChildWorkflowIdFromInputs = (childWorkflowInputs: any): string | undefined => {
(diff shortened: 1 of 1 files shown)


ccastrotrejo (approved):

Eric-B-Wu (approved):

takyyon (approved):

libs/designer/src/lib/ui/panel/nodeDetailsPanel/nodeDetailsPanel.tsx
```diff
@@ -224,11 +224,11 @@ export const NodeDetailsPanel = (props: CommonPanelProps): JSX.Element => {
     return undefined;
   };
 
-  const getChildWorkflowIdFromInputs = (inputs: any): string | undefined => {
-    if (!isNullOrEmpty(inputs)) {
-      const workflow = getObjectPropertyValue(inputs, ['host', 'value', 'workflow']);
+  const getChildWorkflowIdFromInputs = (childWorkflowInputs: any): string | undefined => {
+    if (!isNullOrEmpty(childWorkflowInputs)) {
+      const workflow = getObjectPropertyValue(childWorkflowInputs, ['host.workflow.id']);
```
Eric-B-Wu, line 229: just wondering is host.workflow.id a path? should it be done through: 
const workflow = getObjectPropertyValue(childWorkflowInputs, ['host', 'workflow', 'id']);

ccastrotrejo: The reason on why this is like this, is because when we introduced the InputsBinder we started to render and building the inputs/outputs schema based on the manifest/swagger. This is the manifest from the workflow type action
image

PR #6019, opened 2024-11-01
chore(designer): Clean Up Some Custom Editor Code

Eric-B-Wu:
Not used by Logic Apps, but there were some warnings in the files that custom editors were using, so I did a mini cleanup. There will be more cleanup we'll need to do in the future, but this also gives a starting point for potential custom editors we have in the future.

Also mini change to adding action "HTTP + Swagger" will now be named that instead of "httpwithswagger"


pnpm-lock.yaml
ccastrotrejo, line 1: NIT: is this change needed?

ccastrotrejo (approved):

pnpm-lock.yaml
Eric-B-Wu: we were having some errors being thrown due to type issues in vitest so this should fix those errors

Issue #5864, opened 2024-10-11
new designer container is not inheriting parent height


Xiaoyu-Huang:
### Describe the Bug with repro steps 

The new designer is using height 100vh, instead of inhering parent container height. It will cause styling issue for my product that msla-dsigner-tools is not rendered on page as it is using position: absolute. The root cause is that we have some other UX component like command bar at the page top.

The designer rendering doesn't have issue on Azure portal UI, it is because that the designer is rendered inside of the iframe. For power automation UX, it has override the styling by set height: 'inherit !important'

![Image](<https://github.com/user-attachments/assets/1d468622-c520-45e4-b87f-3fad211125e1>)

Designer version:

 "@microsoft/logic-apps-designer": "4.85.2",

 "@microsoft/logic-apps-shared": "4.85.2",

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

*No response*

### Screenshots or Videos 

*No response*

### Browser 

Edge

### Additional context 

*No response*

PR #6058, opened 2024-11-08
fix(Designer): Designer now inherits container size instead of being 100vh

rllyy97:
## Main Changes

Root designer component height is now set to inherit the container height, instead of being `100vh`.
Our partner teams were all manually overriding this behavior so we are just making it standard.
We will need to update portal to match.

Fixes https://github.com/Azure/LogicAppsUX/issues/5864


hartra344:
We need to make sure this is a major version bump and that all partners are communicated to

PR #6121, opened 2024-11-14
fix(designer): Prevent crashing when having object in key/value

Eric-B-Wu:
Quick Fix for key value crashing when trying to pass an object back to Fluent UI's DetailsList


libs/designer-ui/src/lib/monitoring/values/keyvaluepairs.tsx
```diff
@@ -57,14 +57,14 @@ export const KeyValuePairs: React.FC<ValueProps> = ({ displayName, value = {}, v
...
-          ariaLabel={displayName}
+          ariaLabelForGrid={displayName}
           ariaLabelForListHeader={displayName}
           checkboxVisibility={CheckboxVisibility.hidden}
           columns={columns}
           compact={true}
           isHeaderVisible={true}
           items={items}
-          layoutMode={DetailsListLayoutMode.justified}
+          layoutMode={DetailsListLayoutMode.fixedColumns}
```
takyyon, line 67: Hmmmm this is needed as well? Can you add screenshots here?

ccastrotrejo (approved):

libs/designer-ui/src/lib/monitoring/values/keyvaluepairs.tsx
Eric-B-Wu: We're unable to resize the detailsList columns in standalone (not too sure why), but with this property we don't have well-spaced column widths on load 
before:
before

after:
after

takyyon (approved):

Issue #6070, opened 2024-11-12
New Logic Apps designer adds quotation marks and back slash to the code


PeterKlapwijk:
### Describe the Bug with repro steps 

I have added multiple until actions in a Logic Apps flow using the new designer.

 Under loop until I selected data from a previous step. In the right field I entered 204.

Upon saving the flow I get multiple errors:

Workflow validation failed for the workflow ''. {"error":{"code":"InvalidTemplate","message":"The template validation failed: 'The template action 'Until\_Delete\_windowsHelloForBusinessAuthenticationMethod' at line '1' and column '8632' is not valid: \"The template language expression '*sanitized*' is not valid: the string character '\"' at position '89' is not expected.\".'."}}

Workflow validation failed for the workflow 'LAEmergencyRevokeAccessv1.0'. The template validation failed: 'The template action 'Until\_DELETE\_microsoftAuthenticatorAuthenticationMethod' at line '1' and column '10199' is not valid: "The template language expression 'equals(outputs('HTTP\_GET\_authentication\_methods')?['statusCode'],"204")' is not valid: the string character '"' at position '65' is not expected.".'.

After reviewing the code in Code view it turns out the number 204 is added as \"204\"

When I remove the quotes and back slashes, I can save the flow. When I add additional actions to the flow, the issue returns after which I need to remove the "\ again and again.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json

```

### Screenshots or Videos

![Image](<https://github.com/user-attachments/assets/0af12293-5fe3-469c-840c-c7917484a1b8>)

### Browser

Microsoft Edge.

### Additional context

*No response*

jkerai1:
noticed similar behavior when you are trying to expressions on things like \n, took me a while to realise it was adding another backslash so my expression was never being evaluated had to go into code view and change it from \\n to \n

PeterKlapwijk:
Why is this closed? The issue still occurs. Please reopen the issue!

Solarvista187:
There appears to have been some regression as this issue is still occurring. Can this issue please be reopened.

JoeBarisSpring:
[...]

Issue #3496, opened 2023-10-20
new line character broken when used in editor


SquirrelAssassin:
### Describe the Bug with repro steps 

Make a new variable. try something like a value of array(split(variables('something'), '\n'))

when you save it, the code will change it to '`\`\n'

### What type of Logic App Is this happening in? 

Standard (Portal)

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

No

### Workflow JSON 

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "actions": {
            "For_each": {
                "actions": {
                    "Parse_JSON": {
                        "inputs": {
                            "content": "@items('For_each')",
                            "schema": {
                                "properties": {
                                    "action_type": {
                                        "type": "string"
                                    },
                                    "count": {
                                        "type": "integer"
                                    },
                                    "device_id": {
                                        "type": "string"
                                    },
                                    "device_name": {
                                        "type": "string"
                                    },
                                    "dry_run": {
                                        "type": "boolean"
                                    },
                                    "executable_file_size": {
                                        "type": "integer"
                                    },
                                    "executable_name": {
                                        "type": "string"
                                    },
                                    "executable_path": {
                                        "type": "string"
                                    },
                                    "guid": {
                                        "type": "string"
                                    },
                                    "id": {
                                        "type": "string"
                                    },
                                    "occurrence_date": {
                                        "type": "string"
                                    },
                                    "reason": {
                                        "type": "string"
                                    },
                                    "tenant_id": {
                                        "type": "string"
                                    },
                                    "threat_id": {
                                        "type": "string"
                                    },
                                    "type": {
                                        "type": "string"
                                    }
                                },
                                "type": "object"
                            }
                        },
                        "type": "ParseJson"
                    },
                    "Send_Data": {
                        "inputs": {
                            "body": "@{body('Parse_JSON')}",
                            "headers": {
                                "Log-Type": "Halcyon"
                            },
                            "host": {
                                "connection": {
                                    "referenceName": "azureloganalyticsdatacollector-2"
                                }
                            },
                            "method": "post",
                            "path": "/api/logs"
                        },
                        "runAfter": {
                            "Parse_JSON": [
                                "SUCCEEDED"
                            ]
                        },
[...]

PR #6131, opened 2024-11-15
feat(designer): No longer escapes characters in token expressions

Eric-B-Wu:
By default when stringifying expressions we were escaping the characters. However, this is quite confusing to the user, especially when handling something like `array(split(variables('something'), '\n'))` to be created by: 
```
array(split(variables('something'), '
'))
```
This PR was handling "unescaping" the default behavior of tokens before serialization and then re-escaping the characters before deserializing

Fixes: https://github.com/Azure/LogicAppsUX/issues/3496

unescape

I've also tested this to work on other editors (dictionary, authentication, etc)

Fixes https://github.com/Azure/LogicAppsUX/issues/6070 where quotes were being added unncessarily
Fixed a bug where authentication editor inputs were not being defaulted to string type

### DIFF
--- a/libs/designer/src/lib/core/utils/validation.ts
+++ b/libs/designer/src/lib/core/utils/validation.ts
@@ -373,7 +373,6 @@ export function validateJSONParameter(parameterMetadata: ParameterInfo, paramete
     try {
       JSON.parse(value);
     } catch (e) {
-      console.log(e);
       if (!parameterHasOnlyTokenBinding(parameterValue)) {
         errors.push(
           intl.formatMessage({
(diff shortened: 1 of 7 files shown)


libs/designer/src/lib/core/utils/validation.ts
```diff
@@ -373,6 +373,7 @@ export function validateJSONParameter(parameterMetadata: ParameterInfo, paramete
     try {
       JSON.parse(value);
     } catch (e) {
+      console.log(e);
```
rllyy97, line 376: Is this intentional to leave in?

rllyy97 (approved):

PR #6164, opened 2024-11-21
Feat(DM): deserialization error handling

DanielleCogs:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## New Behavior

Handle error scenarios
1. Backend cannot process LML because of a syntax error (for example user edits LML and adds syntax error)
2. LML error that backend does not handle; invalid source node, invalid target node, invalid loop source node, invalid loop source node, invalid conditional node; there may be a few that we don't cover here, but I will add telemetry to see if this actually ends up catching many errors for users

## Impact of Change

### DIFF
--- a/libs/data-mapper-v2/src/mapHandling/MapDefinitionDeserializer.ts
+++ b/libs/data-mapper-v2/src/mapHandling/MapDefinitionDeserializer.ts
@@ -1,10 +1,21 @@
-import { mapNodeParams, reservedMapDefinitionKeysArray } from '../constants/MapDefinitionConstants';
-import { targetPrefix } from '../constants/ReactFlowConstants';
-import type { FunctionData } from '../models';
-import { indexPseudoFunction } from '../models';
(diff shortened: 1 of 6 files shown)


libs/data-mapper-v2/src/mapHandling/MapDefinitionDeserializer.ts
```diff
@@ -510,11 +528,11 @@ export class MapDefinitionDeserializer {
   };
 
   private isCustomValue = (value: string): boolean => {
-    return value.startsWith('"') || !Number.isNaN(Number.parseInt(value));
+    return value.startsWith('"') || !Number.isNaN(Number.parseInt(value)) || !Number.isNaN(Number.parseFloat(value));
```
DanielleCogs, line 531: Does this capture all custom value scenarios?

takyyon (approved):

Issue #6154, opened 2024-11-20
New designer does not fetch dynamic parameter input based on x-ms-dynamic-values in custom connector's swagger file


jesseck3013:
### Describe the Bug with repro steps

The parameter AdditionalFormIds is configured with x-ms-dynamic-values, but the designer does not fetch the dynamic values initially. However, after saving the workflow and refreshing the page, the issue is resolved, and the dropdown list appears. This problem does not occur in the legacy designer.

![Image](https://github.com/user-attachments/assets/f03c1a04-77b0-48c5-a443-0fbe9df68212)

Please check this [video ](https://teamforms-my.sharepoint.com/personal/anthony_teamforms_app/_layouts/15/stream.aspx?id=%2Fpersonal%2Fanthony%5Fteamforms%5Fapp%2FDocuments%2Fbug%2Din%2Dcustom%2Dconnector%2Emp4&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0&ga=1&referrer=StreamWebApp%2EWeb&referrerScenario=AddressBarCopied%2Eview%2Eccd735e1%2D6641%2D466b%2D9e8a%2Dd48b1885afac) and my swagger file.

### What type of Logic App Is this happening in?

Consumption (Portal)

### Which operating system are you using?

Windows

### Are you using new designer or old designer

New Designer

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json
{
  "swagger": "2.0",
  "info": {
    "title": "Team Forms",
    "description": "Teams Forms brings digital forms into Microsoft Teams. The software will empower your teams to build and deliver forms from within the productivity tools that they already know and trust. Unlike many other forms solutions on the market, Team Forms talks directly with your Teams SharePoint so that data captured by forms remains in your ownership and never leaves your trusted office 365 environment.",
    "version": "1.1.0",
    "termsOfService": "https://teamforms.app/terms",
    "contact": {
      "name": "Team Forms",
      "url": "https://teamforms.app",
      "email": "support@teamforms.app"
    }
  },
  "host": "api.teamforms.app",
[...]

PR #6221, opened 2024-12-05
fix(Designer): Fixed issue with array editor dynamic data

rllyy97:
## Main Changes

Currently, if a user adds an additional parameter (typed as an array) that has dynamic data, that dynamic data will not load until the user saves and reloads the workflow.
This is because when we add the additional array parameter, it decides whether or not to load a combobox based on if there are options to show.  If there are no options, it shows a string editor.  The dynamic data fetch is a part of the combobox editor, so they never load the dynamic data options.

This pr adds a new `isDynamic` prop so that the array editor knows when to load a combobox in situations that hasn't loaded the dynamic data options.  

Fixes https://github.com/Azure/LogicAppsUX/issues/6154

array_dd_fix


Eric-B-Wu (approved):

rllyy97:
Holding until January

Issue #6126, opened 2024-11-14
Copy-Paste Not Updating Dynamic Content for Action Inside For-Each Actions


pranaydubeymicrosoft:
### Describe the Bug with repro steps 

1. Create a Logic App with recurrence trigger and get items SharePoint action. 
2. Add a for-each action and iterate over the dynamic content returned from the SharePoint action. 
3. Inside the for-each action, add a compose action. Provide "Title" for input (from SharePoint dynamic content). 
4. Save Logic App. 
5. Copy the for-each action, paste it below. 
6. Actions inside the pasted for-each will have errors. 

You can paste the action in notepad and see that the nodeId reflects the new value (For\_each-copy) but the dynamic content inside is not updated accordingly:

{"nodeId":"For\_each-copy","serializedOperation":{"type":"Foreach","foreach":"@body('Get\_items')?['value']","actions":{"Compose":{"type":"Compose","inputs":"@item()?['Title']"}},"runAfter":{"Get\_items":["SUCCEEDED"]}},"allConnectionData":{},"staticResults":{},"isScopeNode":true,"mslaNode":true}

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Are you using new designer or old designer 

New Designer

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json
{"nodeId":"For_each-copy","serializedOperation":{"type":"Foreach","foreach":"@body('Get_items')?['value']","actions":{"Compose":{"type":"Compose","inputs":"@item()?['Title']"}},"runAfter":{"Get_items":["SUCCEEDED"]}},"allConnectionData":{},"staticResults":{},"isScopeNode":true,"mslaNode":true}
```

### Screenshots or Videos

![Image](<https://github.com/user-attachments/assets/95c24f2d-9feb-4127-8cf1-ef166e162f25>)

### Browser

Edge

### Additional context

*No response*

PR #6236, opened 2024-12-09
fix(designer): Fixing Implicit For Each within Copy-Paste Scope 

Eric-B-Wu:
Fixes https://github.com/Azure/LogicAppsUX/issues/6126
We require the actionName gotten from updateTokenMetadata to be used in updateScopePasteTokenMetadata
copyscopeimplict


libs/designer/src/lib/core/actions/bjsworkflow/operationdeserializer.ts
```diff
@@ -465,7 +466,7 @@ const initializeOutputTokensForOperations = (
           nodesWithData[operationId]?.settings
         )
       );
-    } catch (e) {
+    } catch {
```
takyyon, line 469: not logging the exception?

Eric-B-Wu: we had changed our tsconfigs recently so that any unused variables error, this code I didn't change, but wanted to pass our commit

takyyon: Yeah that's totally okay but I meant we should be logging the exception/error here. Will help us debug in cases where it fails

ccastrotrejo (approved):

PR #6303, opened 2024-12-23
fix(Designer): Keyboard focus move outside of 'Insert dynamic content' dialog

Elaina-Lee:
## Requirement Checklist

## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
Keyboard focus move outside of the 'Insert dynamic content' dialog when navigating using keyboard

## New Behavior
Keyboard stays within the dialog unless 'esc' key is pressed. When using mouse to navigate, it can still exit the focus when clicked outside of dialog.

## Impact of Change

[
](https://github.com/user-attachments/assets/1b8d9041-35b7-4b6c-94ba-8d8774c8351f)


Eric-B-Wu (changes requested):
I think we can look to trap focus in the tokenpicker, but it should be done in another way. With FocusTrapCallout, when users blur the tokenpicker, the focus always goes back to the editor which opened that tokenpicker. For example, if the user clicks another editor with the tokenpicker open, the focus isn't on the new editor, but rather the original editor that opened the tokenpicker

PR #6372, opened 2025-01-09
feat(Designer): Exported Run History panel

rllyy97:
## Main Changes

- Added importable run history panel for switching between runs, and between edit / monitoring views.
    - The panel is a separate component so that you can use it outside of designer, and use it to modify the data passed to designer.
    - This should be able to be dropped into our designer portal blades fairly easily, I've adjusted anything that needed to be adjusted to allow this to function properly.
- Brought Datetime util over from portal
- There's definitely more improvments / optimizations we can do here but I'm just setting up a baseline before the PR gets too big, in it's current state it's fully functional and can be imported as-is

### Switching between runs / views (Some of the loading lag is from running in standalone)
designer_run_history


libs/designer/src/lib/ui/panel/nodeSearchPanel/nodeSearchPanel.tsx
ccastrotrejo, line 1: NIT: Unneeded changes?

apps/Standalone/src/designer/app/AzureLogicAppsDesigner/laDesignerConsumption.tsx
```diff
@@ -280,23 +298,12 @@ const DesignerEditorConsumption = () => {
...
-                isReadOnly={readOnly}
-                isDarkMode={isDarkMode}
-                isDesignerView={designerView}
-                showConnectionsPanel={showConnectionsPanel}
-                enableCopilot={() => {
-                  dispatch(setIsChatBotEnabled(!showChatBot));
-                }}
-                switchViews={handleSwitchView}
-                saveWorkflowFromCode={saveWorkflowFromCode}
+            <div style={{ display: 'flex', flexDirection: 'row', height: 'inherit' }}>
```
ccastrotrejo, line 301: NIT: no need to set flexDirection as row, I think the default for display flex is row

ccastrotrejo (changes requested):

libs/designer/src/lib/ui/panel/nodeSearchPanel/nodeSearchPanel.tsx
rllyy97: I'm leaving these since they're not big enough for their own PR, it's just a general organization tweak

ccastrotrejo (approved):

PR #6384, opened 2025-01-13
fix(Data Mapper): Fix wrapping, button click and re-ordering

takyyon:
## Requirement Checklist

## Type of Change

* [ ] Bug fix
* [ ] Feature
* [ ] Other

## Impact of Change


libs/data-mapper-v2/src/components/functionConfigurationMenu/inputTab/styles.tsx
ccastrotrejo, line 1: NIT: `flex-direction: row` its default when setting the `display: flex`, usually there is no need to explicitly state it, don't know the whole context but maybe it's not needed

takyyon: Yup, just wanted to have it out there anyways in case we end up regressing this lol

DanielleCogs (approved):

takyyon:
https://github.com/Azure/LogicAppsUX/pull/6393

PR #6396, opened 2025-01-15
Fix(Data Mapper): add new files

DanielleCogs:
## Requirement Checklist

## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
due to vscode update, add new schema flow was broken

## New Behavior
use vscode api to select files

## Impact of Change


apps/vs-code-designer/src/app/commands/dataMapper/DataMapperPanelUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+        )
+      );
+      return;
+    }
+
+    // Check that the schema file dependency doesn't already exist in the Schemas folder
+    const newSchemaFilePath = path.join(pathToWorkspaceSchemaFolder, importedSchemaFileName);
+    if (!fileExistsSync(newSchemaFilePath)) {
+      copyFileSync(importedSchemaFileFullPath, newSchemaFilePath);
+    }
```
DanielleCogs, line 39: Need to eventually add 'else' clause here. Either telling the user that the file already exists, or renaming and copying over the current file. Made this bug to follow up https://msazure.visualstudio.com/One/_workitems/edit/30861689 but doesn't seem like a priority right now.

PR #6427, opened 2025-01-17
meta(vscode): Add pipeline to test and build private feature vsix

ccastrotrejo:
This pull request introduces a new GitHub Actions workflow to automate the building and testing of a private VSIX (Visual Studio Code extension) package. The workflow is designed to run tests on different Node.js versions and build the VSIX package with specific configurations.

### Why having new pipeline
 The idea is to have a pipeline in which creates a VSIX that closely resembles a release version. This pipeline requires a branch name as input, runs the tests, and builds the VSIX in GitHub. As a result, we will be able to distribute a more solid private vsix

### New GitHub Actions Workflow:
* Added a new workflow named "Private vsix build" that includes two jobs: `test` and `build`. The `test` job runs on multiple Node.js versions (18.x, 20.x, 22.x) and caches the turbo build setup. The `build` job sets up Node.js version 20, replaces placeholders with the telemetry key, and packs the VSCode Designer Extension.

### DIFF
--- a/.github/workflows/private-vsix-build.yml
+++ b/.github/workflows/private-vsix-build.yml
@@ -1,0 +1,86 @@
+name: Private vsix build
+run-name: Private vsix build - [ ${{ github.ref_name || 'branch'  }} ]
+on:
+  workflow_dispatch: # Trigger only manually  workflow_dispatch: # Trigger only manually
+
+env:
+  AI_KEY: 3cf0d6ae-3327-414a-b7c1-12f31ef45eff
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [18.x, 20.x, 22.x]
+
+    steps:
+      # checkout the repo
+      - name: 'Checkout Github Action'
+        uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+
+      - name: Cache turbo build setup
+        uses: actions/cache@v4
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-turbo-
+
+      - name: Set up Node.js version ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - uses: pnpm/action-setup@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/private-vsix-build.yml
```diff
@@ -0,0 +1,86 @@
+name: Private vsix build
+run-name: Private vsix build - [ ${{ github.ref_name || 'branch'  }} ]
+on:
+  workflow_dispatch: # Trigger only manually  workflow_dispatch: # Trigger only manually
+
+env:
+  AI_KEY: 3cf0d6ae-3327-414a-b7c1-12f31ef45eff
```
mireedmsft, line 7: is this a secret? should we be storing it in source?

mireedmsft: especially on open source github

ccastrotrejo: The AI key is not a secret, we already have it in the repo, according to the telemetry package that vscode offers is not sensitive 

The AI key is not a secret; it’s already in the repository. According to the telemetry package provided by VS Code, it is not considered sensitive information. [For more details, you can refer to the VS Code extension telemetry documentation.](https://github.com/Microsoft/vscode-extension-telemetry#readme)

mireedmsft (approved):

.github/workflows/private-vsix-build.yml
takyyon: I would say we should still add this to a variable in case we need to change it in future. Will be handy without changing the code. If it is not a secret, the value can be set as a variable instead of secret

ccastrotrejo: Update it to use a repository variable

.github/workflows/private-vsix-build.yml
```diff
@@ -0,0 +1,86 @@
+name: Private vsix build
+run-name: Private vsix build - [ ${{ github.ref_name || 'branch'  }} ]
+on:
+  workflow_dispatch: # Trigger only manually  workflow_dispatch: # Trigger only manually
+
+env:
+  AI_KEY: ${{ vars.AI_KEY }}
```
mireedmsft, line 7: even better :) awesome thanks

takyyon (approved):

Issue #6457, opened 2025-01-23
VS Code : Unable to deploy workflow from VS code to non-public clouds


vgouth:
### Describe the Bug with repro steps 

1\.Login into mooncake national cloud environment and Create a new Logic App Project

 2,Add a workflow to your local project

 3,Open the workflow.json file in designer and author a new workflow

 4\.Add http request Trigger

 5\.Add a compose action here with custom text

6. Save and run the workflow 
7. Deploy the workflow 

**Expected:** Deployment should work as expected

**Actual:**Unable to deploy the workflow getting below error. Tried alternative method of creating a logic app and then deploying the workflow. But unable to create logic app as well.

### What type of Logic App Is this happening in? 

Standard (VSCode)

### Which operating system are you using? 

Windows

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json

```

### Screenshots or Videos

![Image](<https://github.com/user-attachments/assets/65e8031c-c92d-4c10-bcb6-77ad43d723f2>)

Unable to create logic app :

![Image](<https://github.com/user-attachments/assets/ecfaaed4-aa1c-4542-98a3-14fb152872b2>)

### Browser

VS Code

### Additional context

**VSIX Version:** 5\.14.2

PR #6471, opened 2025-01-25
fix(vscode): Filter locations for new resources

ccastrotrejo:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

Fixes - #6457 

## Current Behavior

Region dropdown was showing all the available regions including the ones that Logic Apps Standard is not supported

## New Behavior

Region dropdown now filters the regions based on the ones that Logic Apps Standard is supported

## Impact of Change

Non a breaking change, will actually fix and avoid users face errors when trying to deploy to regions that are not supported

## Test Plan
 

This has been tested and will be tested by the manual vendor testing in the following test- https://msazure.visualstudio.com/One/_workitems/edit/30411354
 For future intances I will take some time and look back again on how we can implement e2e tests in the extension.

### Old behavior

Screenshot 2025-01-24 at 3 42 55 PM

### New behavior

Screenshot 2025-01-24 at 3 40 36 PM


hartra344:
Anyway we can have a test plan on this?

rllyy97 (approved):

hartra344 (approved):

ccastrotrejo:
> Anyway we can have a test plan on this?

Yes, I recently added a test to the list of manual testing by the vendors for non-public clouds and this was the first time they went through it and catched this. For future intances I will take some time and look back again on how we can implement e2e tests in the extension.

PR #6481, opened 2025-01-27
feat(Templates): Support Blank Workflow Creation for Templates

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

- Adding async support for onAddBlankWorkflow call for TemplateService
- Restructuring Templates standalone to be two separate components - standard and consumption
- Overwriting the consumption with blank workflow upon blank workflow card click

## Test Plan

onBlankWorkflow is a call within the service, and is changed to the async call. Adding e2e tests to ensure clicking on the blank card triggers the functions provided in the local templates standalone.

https://github.com/user-attachments/assets/2616f630-9747-4060-91ae-d26f98ee098a


libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
```diff
@@ -15,7 +15,7 @@ export class BaseTemplateService implements ITemplateService {
 
   public openBladeAfterCreate = (workflowName: string | undefined): void => this.options.openBladeAfterCreate(workflowName);
 
-  public onAddBlankWorkflow = (): void => this.options.onAddBlankWorkflow();
+  public onAddBlankWorkflow = (): Promise<void> | void => this.options.onAddBlankWorkflow();
```
tonytang-microsoft-com, line 18: Can a function return either void or Promise<void>, in which case I'm not sure how caller are going to handle it. Should it await or not (assuming we can't await on void, unless I'm missing something here)?

hartra344 (approved):

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
hartra344: Yes, I agree, it should always return a promise

libs/logic-apps-shared/src/designer-client-services/lib/standard/template.ts
```diff
@@ -39,7 +39,7 @@ export class StandardTemplateService extends BaseTemplateService {
       });
 
       return responseData?.value?.map((workflow: any) => workflow.name.split('/')[1]) ?? [];
-    } catch (error: any) {
+    } catch (_error: any) {
       return [];
```
ccastrotrejo, line 43: We don't need to log the error when the api fail or get the error message? So user knows that it failed?

Elaina-Lee: It was communicated that if this call returns an error in the case of no workflow present

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
Elaina-Lee: This was due to - for standard, it won't return a promise. Understood and reflected the change.

ccastrotrejo (approved):

PR #6485, opened 2025-01-28
feat(templates): Support auto selecting template

tonytang-microsoft-com:
## Type of Change

* [ ] Bug fix
* [ x] Feature
* [ ] Other

## New Behavior

Add support to auto selecting template. When this new parameter is set, the template designer will automatically select the template as if user clicked the template and create.

## Impact of Change

This is additive change so it's fully back compatible

## Test Plan

Tested locally and added CIT

image


hartra344 (approved):

ccastrotrejo (approved):

libs/designer/src/lib/ui/templates/TemplatesDesigner.tsx
```diff
@@ -17,9 +17,13 @@ export type CreateWorkflowHandler = (
   connectionsMapping: ConnectionMapping,
   parametersData: Record<string, Template.ParameterDefinition>
 ) => Promise<void>;
+
 export interface TemplatesDesignerProps {
   detailFilters: TemplateDetailFilterType;
   createWorkflowCall: CreateWorkflowHandler;
+  viewTemplate?: {
+    templateName?: string;
```
Elaina-Lee, line 25: Could we make this an non-optional field? If optional param of viewTemplate is present, we should assume templateName is present.

libs/designer/src/lib/ui/templates/templateslist.tsx
```diff
@@ -5,25 +5,46 @@ import { EmptySearch, Pager } from '@microsoft/designer-ui';
...
+import { openCreateWorkflowPanelView } from '../../core/state/templates/panelSlice';
+import { loadTemplate } from '../../core/actions/bjsworkflow/templates';
 
-export const TemplatesList = ({ detailFilters, createWorkflowCall }: TemplatesDesignerProps) => {
+export const TemplatesList = ({ detailFilters, createWorkflowCall, viewTemplate }: TemplatesDesignerProps) => {
   useEffect(() => setLayerHostSelector('#msla-layer-host'), []);
   const intl = useIntl();
   const dispatch = useDispatch<AppDispatch>();
 
+  const [locked, setLocked] = useState(false);
```
Elaina-Lee, line 23: This locked state is only being used in useEffect to check if it has been set to true. I'm wondering since it's in useEffect, if it's only being used to ensure this is run once. If so, since we are already keeping `templateLocked` in the state, we should use this for the check.

libs/designer/src/lib/ui/templates/templateslist.tsx
```diff
@@ -5,25 +5,46 @@ import { EmptySearch, Pager } from '@microsoft/designer-ui';
...
+
   const { templateName, workflows } = useSelector((state: RootState) => state.template);
   const {
+    availableTemplates,
     filteredTemplateNames,
     filters: { pageNum, detailFilters: appliedDetailFilters },
   } = useSelector((state: RootState) => state.manifest);
   const selectedTabId = appliedDetailFilters?.Type?.[0]?.value;
 
+  useEffect(() => {
```
Elaina-Lee, line 33: These logic should reside in TemplatesDataProvider to keep related functionality centralized. Since this modifies the state determining which templates are displayed, I would suggest it makes the most to keep similar logic together for better organization and coherence.

libs/designer/src/lib/ui/templates/templateslist.tsx
```diff
@@ -5,25 +5,46 @@ import { EmptySearch, Pager } from '@microsoft/designer-ui';
...
+        setLocked(true);
+        dispatch(lockTemplate(viewTemplate.templateName));
+        dispatch(loadTemplate({ preLoadedManifest: templateManifest }));
+
+        if (Object.keys(templateManifest?.workflows ?? {}).length === 0) {
+          dispatch(openCreateWorkflowPanelView());
+        }
+      }
+    }
+  }, [dispatch, availableTemplates, viewTemplate, filteredTemplateNames]);
```
Elaina-Lee, line 46: In case where we don't list all the variables used within useEffect in the dependency list intentionally, we put `// eslint-disable-next-line react-hooks/exhaustive-deps` on the top of the dependency list

libs/designer/src/lib/core/state/templates/templateSlice.ts
```diff
@@ -5,6 +5,7 @@ import { getCurrentWorkflowNames, validateConnectionsValue, validateParameterVal
 import { initializeTemplateServices, loadTemplate, validateWorkflowName, type TemplatePayload } from '../../actions/bjsworkflow/templates';
 
 export interface TemplateState extends TemplatePayload {
+  templateNameLocked?: boolean;
```
Elaina-Lee, line 8: I would like to suggest to renaming this to `isTemplateNameLocked`

Elaina-Lee (changes requested):

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/createWorkflowPanel.tsx
```diff
@@ -69,6 +71,10 @@ export const CreateWorkflowPanel = ({ createWorkflow, onClose, clearDetailsOnClo
   };
 
   const dismissPanel = useCallback(() => {
+    if (templateNameLocked) {
```
Elaina-Lee, line 74: This value is not added to the dependency list of useCallback

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/usePanelTabs.tsx
```diff
@@ -177,6 +180,7 @@ export const useCreateWorkflowPanelTabs = ({
         nextTabId: parametersExist ? Constants.TEMPLATE_PANEL_TAB_NAMES.PARAMETERS : Constants.TEMPLATE_PANEL_TAB_NAMES.REVIEW_AND_CREATE,
         hasError: !!connectionsError,
         isCreating,
+        templateNameLocked,
```
Elaina-Lee, line 183: templateNameLocked is not added in any of the useMemo dependency list for tab items

libs/designer/src/lib/ui/templates/templateslist.tsx
```diff
@@ -5,25 +5,46 @@ import { EmptySearch, Pager } from '@microsoft/designer-ui';
...
   } = useSelector((state: RootState) => state.manifest);
   const selectedTabId = appliedDetailFilters?.Type?.[0]?.value;
 
+  useEffect(() => {
+    if (viewTemplate?.templateName && filteredTemplateNames?.includes(viewTemplate.templateName)) {
+      const templateManifest = availableTemplates?.[viewTemplate.templateName];
+      if (templateManifest && !locked) {
+        setLocked(true);
+        dispatch(lockTemplate(viewTemplate.templateName));
+        dispatch(loadTemplate({ preLoadedManifest: templateManifest }));
```
Elaina-Lee, line 39: we need to pass isCustomTemplate boolean here, otherwise this will default to assuming it is present in github. Since we are checking over filteredTemplateNames, this could include custom templates.

tonytang-microsoft-com:
aborting the PR since Elaina has forked the branch and made more changes.

PR #6498, opened 2025-01-28
feat(Templates): Support auto selecting template

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## New Behavior

Add support to auto selecting template. When this new parameter is set, the template designer will automatically select the template as if user clicked the template and create.

## Impact of Change

This is additive change so it's fully back compatible
**built on top of https://github.com/Azure/LogicAppsUX/pull/6485 with additional changes**

## Test Plan

Tested locally and added CIT

image

### DIFF
--- a/libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
+++ b/libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
@@ -19,8 +19,10 @@ import {
 import { useAreServicesInitialized } from '../state/templates/templateselectors';
 import type { ConnectionReferences } from '../../common/models/workflow';
 import { getFilteredTemplates } from './utils/helper';
-import { initializeTemplateServices } from '../actions/bjsworkflow/templates';
+import { initializeTemplateServices, loadTemplate } from '../actions/bjsworkflow/templates';
 import type { Template } from '@microsoft/logic-apps-shared';
+import { lockTemplate } from '../state/templates/templateSlice';
+import { openCreateWorkflowPanelView } from '../state/templates/panelSlice';
 
 export interface TemplatesDataProviderProps {
   isConsumption: boolean | undefined;
@@ -29,12 +31,24 @@ export interface TemplatesDataProviderProps {
   services: TemplateServiceOptions;
   connectionReferences: ConnectionReferences;
   customTemplates?: Record<string, Template.Manifest>;
+  viewTemplate?: {
+    templateName: string;
+  };
   children?: React.ReactNode;
 }
 
-const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowName, children }: TemplatesDataProviderProps) => {
+const DataProviderInner = ({
+  customTemplates,
+  isConsumption,
+  existingWorkflowName,
+  viewTemplate,
+  children,
+}: TemplatesDataProviderProps) => {
   const dispatch = useDispatch<AppDispatch>();
-  const { githubTemplateNames, availableTemplates, filters } = useSelector((state: RootState) => state?.manifest);
+  const { githubTemplateNames, filteredTemplateNames, availableTemplates, customTemplateNames, filters } = useSelector(
+    (state: RootState) => state?.manifest
+  );
+  const { isTemplateNameLocked } = useSelector((state: RootState) => state.template);
 
   useEffect(() => {
     dispatch(loadGithubManifestNames());
@@ -70,6 +84,22 @@ const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowNam
       dispatch(setExistingWorkflowName(existingWorkflowName));
     }
   }, [dispatch, existingWorkflowName]);
+
+  useEffect(() => {
+    if (viewTemplate?.templateName && filteredTemplateNames?.includes(viewTemplate.templateName) && !isTemplateNameLocked) {
+      const templateManifest = availableTemplates?.[viewTemplate.templateName];
+      if (templateManifest) {
+        dispatch(lockTemplate(viewTemplate.templateName));
+        dispatch(
+          loadTemplate({ preLoadedManifest: templateManifest, isCustomTemplate: customTemplateNames?.includes(viewTemplate.templateName) })
+        );
+
+        if (Object.keys(templateManifest?.workflows ?? {}).length === 0) {
+          dispatch(openCreateWorkflowPanelView());
+        }
+      }
+    }
+  }, [dispatch, availableTemplates, viewTemplate, isTemplateNameLocked, filteredTemplateNames, customTemplateNames]);
 
   return <>{children}</>;
 };
--- a/libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
+++ b/libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
@@ -45,9 +45,7 @@ const DataProviderInner = ({
   children,
 }: TemplatesDataProviderProps) => {
   const dispatch = useDispatch<AppDispatch>();
-  const { githubTemplateNames, filteredTemplateNames, availableTemplates, customTemplateNames, filters } = useSelector(
-    (state: RootState) => state?.manifest
-  );
+  const { githubTemplateNames, availableTemplates, filters } = useSelector((state: RootState) => state?.manifest);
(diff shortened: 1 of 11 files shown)


libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
```diff
@@ -71,6 +85,22 @@ const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowNam
...
+        dispatch(
+          loadTemplate({ preLoadedManifest: templateManifest, isCustomTemplate: customTemplateNames?.includes(viewTemplate.templateName) })
+        );
+
+        if (Object.keys(templateManifest?.workflows ?? {}).length === 0) {
+          dispatch(openCreateWorkflowPanelView());
+        }
+      }
+    }
+  }, [dispatch, availableTemplates, viewTemplate, isTemplateNameLocked, filteredTemplateNames, customTemplateNames]);
```
tonytang-microsoft-com, line 102: Maybe remove customTemplateNames, as we don't support custom template? Even if we added support later, I'm not sure if we want to allow caller to pre select the template as the template name should be a well known name and I assume custom template name could change. But we could add the support later if needed. Just trying to remove unncessary depends.

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/createWorkflowPanel.tsx
```diff
@@ -69,14 +71,18 @@ export const CreateWorkflowPanel = ({ createWorkflow, onClose, clearDetailsOnClo
...
+
     dispatch(closePanel());
 
     if (clearDetailsOnClose) {
       dispatch(clearTemplateDetails());
     }
 
     onClose?.();
-  }, [clearDetailsOnClose, dispatch, onClose]);
+  }, [isTemplateNameLocked, clearDetailsOnClose, dispatch, onClose]);
```
tonytang-microsoft-com, line 85: It's okay to add this as a dep, but worth noting that it should never change in our scenario.

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
```diff
@@ -71,6 +85,22 @@ const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowNam
     }
   }, [dispatch, existingWorkflowName]);
 
+  useEffect(() => {
+    if (viewTemplate?.templateName && filteredTemplateNames?.includes(viewTemplate.templateName) && !isTemplateNameLocked) {
+      const templateManifest = availableTemplates?.[viewTemplate.templateName];
+      if (templateManifest) {
+        dispatch(lockTemplate(viewTemplate.templateName));
+        dispatch(
+          loadTemplate({ preLoadedManifest: templateManifest, isCustomTemplate: customTemplateNames?.includes(viewTemplate.templateName) })
```
tonytang-microsoft-com, line 94: I haven't tested custom template and not planning to support it. Do you think we could block this scenario? I'm not even sure if there will be a scenario ever.

tonytang-microsoft-com (approved):

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
Elaina-Lee: If that's the case, I'll be changing the check to `githubTemplateNames?.includes(viewTemplate.templateName)`

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/createWorkflowPanel.tsx
Elaina-Lee: What would you mean that it would never change? The initial value of isTemplateNameLocked should be false; at the time of initialization, the value of isTemplateNameLocked gets updated to true.

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
Elaina-Lee: Will be resolving as the same comment as above

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
```diff
@@ -71,6 +83,20 @@ const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowNam
...
   }, [dispatch, existingWorkflowName]);
 
+  useEffect(() => {
+    if (viewTemplate?.templateName && githubTemplateNames?.includes(viewTemplate.templateName) && !isTemplateNameLocked) {
+      const templateManifest = availableTemplates?.[viewTemplate.templateName];
+      if (templateManifest) {
+        dispatch(lockTemplate(viewTemplate.templateName));
+        dispatch(loadTemplate({ preLoadedManifest: templateManifest, isCustomTemplate: false }));
+
+        if (Object.keys(templateManifest?.workflows ?? {}).length === 0) {
```
preetriti1, line 93: Why is workflows empty?

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
```diff
@@ -29,12 +31,22 @@ export interface TemplatesDataProviderProps {
...
 }
 
-const DataProviderInner = ({ customTemplates, isConsumption, existingWorkflowName, children }: TemplatesDataProviderProps) => {
+const DataProviderInner = ({
+  customTemplates,
+  isConsumption,
+  existingWorkflowName,
+  viewTemplate,
+  children,
+}: TemplatesDataProviderProps) => {
```
preetriti1, line 46: Why would we be loading all templates data when the ask from portal blade is to only open one single template? This is unnecessarily going to increase the panel load time.. that is why I was saying why do you want to lock the template

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/basicsTab.tsx
```diff
@@ -48,6 +48,6 @@ export const basicsTab = (
         dispatch(clearTemplateDetails());
       }
     },
-    secondaryButtonDisabled: isCreating,
+    secondaryButtonDisabled: isCreating || isTemplateNameLocked,
```
preetriti1, line 51: did we decide to show the option of stateful and stateless selection and that is the reason to show the basics tab?

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/reviewCreateTab.tsx
```diff
@@ -214,6 +215,6 @@ export const reviewCreateTab = (
         }
       }
     },
-    secondaryButtonDisabled: isCreating,
+    secondaryButtonDisabled: isCreating || (!previousTabId && isTemplateNameLocked),
```
preetriti1, line 218: Why do we have this logic in all secondary buttons related to template locked? in reviewCreateTab previousTabId is always true right?

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/basicsTab.tsx
Elaina-Lee: Yes, that is the reason we show the basics tab

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
Elaina-Lee: They are looking for the least amount of code change for this purpose + the background of the template list view.....
I understand that we discussed about the concern of locking the template, but we had multiple meetings involving Divya where she agreed to the use case of this having to be the case..
@tonytang-microsoft-com would you have thoughts on this?

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
Elaina-Lee: With our current logic, we check if the manifest workflows is unpresent as a way of checking for single workflow or multi workflow (until before manifest restructure).

hartra344 (approved):

PR #6508, opened 2025-01-29
fix(vscode): Remove cwd from workspace path

ccastrotrejo:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Since we opted in for the default workspace behavior approach, the extra directory in the "CWD" property when trying to run and debug the logic app was causing issues and the logic app wasn't able to run in vscode

## New Behavior

The run and debug task is now able to handle the workspace behavior and now the logic apps can be run

## Impact of Change

This is not a breaking change

## Test Plan

 

We figure out this issue was happening from the development team, this issue didn't make it to production. However this issue would have been a blocker to release since as for now this is tested by manual vendor testing.
For future references, I will see if I can test this in unit tests,  in the meantime I am still takin a look on how to implement e2e tests for the extension


apps/vs-code-designer/src/app/commands/createNewProject/createProjectSteps/ScriptProjectCreateStep.ts
```diff
@@ -14,7 +14,6 @@ import {
   logicAppKind,
   vscodeFolderName,
   workerRuntimeKey,
-  workflowSubscriptionIdKey,
```
mireedmsft, line 17: Is this related? or just a drive by fix?

mireedmsft (approved):

apps/vs-code-designer/src/app/commands/createNewProject/createProjectSteps/ScriptProjectCreateStep.ts
ccastrotrejo: While it isn’t entirely related, I noticed that we had introduced that change as well. I decided to revert it to the previous implementation we were using.

hartra344 (approved):

takyyon (approved):

PR #6525, opened 2025-01-31
fix(vscode): Correct toPascalCase regex

andrew-eldridge:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Current regex fails on test string "_example_string" with leading underscores

## New Behavior

New regex accounts for leading underscores

## Impact of Change

## Test Plan

Addressed in existing unit tests

N/A


hartra344 (changes requested):
you mention that it's addressed in existing unit tests? In that case, what is prompting the fix? If the test was succeeding before and still succeeding now, but we fixed a real issue then it sounds like we need another unit test. Can you please add that?

hartra344 (approved):

PR #6534, opened 2025-02-03
feat(designer): Introduce collapse node route

ccastrotrejo:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior

There is no current behavior for this. However, customers with lots of actions will need to scroll across the canvas to find actions.

## New Behavior

New behavior will allow users to collapse actions, making the flow smaller in the canvas, therefore users will be able to see more of the flow at once. This also improves performance by not rendering the actions that are collapsed

## Impact of Change

## Test Plan

 

All new react components have unit test.
Added e2e tests

#### Scenario with scope actions

https://github.com/user-attachments/assets/c8666033-9c6a-4946-b86a-740fca91fdb5

#### Scenario with 200 actions

https://github.com/user-attachments/assets/2d3053dd-a559-4689-8eaa-0460e1def184

#### Update: When clicked in collapsed, now the collapsed node is focused

https://github.com/user-attachments/assets/91fd3fae-9a49-40ec-ae03-5c6a15eb97b3


hartra344:
Can we also fix it so that when an action is collapsed, we center the graph onto that action rather than having it move around unpredictibly?

PR #6553, opened 2025-02-05
feat(vscode): Use listMockableOperations API for create blank unit test

andrew-eldridge:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior

Currently, mockable action and trigger values are hard-coded constants

## New Behavior

Fetch mockable operations using listMockableOperations API

## Impact of Change

## Test Plan

Added unit test for new function getMockableOperations

N/A


apps/vs-code-designer/src/app/commands/workflows/unitTest/saveBlankUnitTest.ts
```diff
@@ -320,30 +321,65 @@ async function generateBlankCodefulUnitTest(
...
+
+    const resContentType = response.headers['content-type'];
+    if (!resContentType.includes('application/json')) {
+      throw new Error(localize('invalidResponseType', 'Expected json response but received {0}', resContentType));
+    }
+
+    response.data.forEach((mockableOperation: string) => mockableOperationTypes.add(mockableOperation.toUpperCase()));
+  } catch (apiError: any) {
+    // eslint-disable-next-line import/no-named-as-default-member
+    if (axios.isAxiosError(apiError)) {
```
ccastrotrejo, line 355: Can we log the error in telemetry?

apps/vs-code-designer/src/app/commands/workflows/unitTest/saveBlankUnitTest.ts
```diff
@@ -320,30 +321,65 @@ async function generateBlankCodefulUnitTest(
...
+      ext.outputChannel.appendLog(
+        localize(
+          'apiCallFailed',
```
ccastrotrejo, line 358: Which api call failed?

apps/vs-code-designer/src/app/commands/workflows/unitTest/saveBlankUnitTest.ts
```diff
@@ -320,30 +321,65 @@ async function generateBlankCodefulUnitTest(
...
+
+  // Log API details and initiate call
+  ext.outputChannel.appendLog(localize('initiatingApiCall', 'Initiating List Mockable Operations API call...'));
+
+  try {
+    const response = await axios.get(apiUrl);
+
+    ext.outputChannel.appendLog(localize('apiCallSuccessful', 'API call successful, processing response...'));
+
+    const resContentType = response.headers['content-type'];
```
ccastrotrejo, line 347: Is this a possible scenario? can the response from the api call return a non `application/json` content-type header? Does that mean that the api is not reliable and the return values can vary?

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/commands/workflows/unitTest/saveBlankUnitTest.ts
andrew-eldridge: This shouldn't be the case, can probably safely remove this check

ccastrotrejo: That would be great, thanks @andrew-eldridge!!

andrew-eldridge: removed Content-Type check

apps/vs-code-designer/src/app/commands/workflows/unitTest/saveBlankUnitTest.ts
andrew-eldridge: updated logs to be more descriptive

ccastrotrejo (approved):

andrew-eldridge:
> This should still contain a test plan please for at least some unit testing

Added unit test for new function

ccastrotrejo (approved):

PR #6554, opened 2025-02-05
feat(templates): Adding TemplateView component to view single template without gallery

preetriti1:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Change
Adding a new component to just view a single template with multiple overrides for basics and parameters tab from the host.

## Test Plan
Will be writing e2e tests for local templates view in next iteration. Also tested this manually from standalone test page both for standard and consumption.

singletemplates


libs/designer/src/lib/ui/templates/templateoverview.tsx
```diff
@@ -17,7 +17,17 @@ import { workflowTab } from '../panel/templatePanel/quickViewPanel/tabs/workflow
 import { clearTemplateDetails } from '../../core/state/templates/templateSlice';
 import { CreateWorkflowPanel } from '../panel/templatePanel/createWorkflowPanel/createWorkflowPanel';
 
-export const TemplateOverview = ({ createWorkflow }: { createWorkflow: CreateWorkflowHandler }) => {
+export const TemplateOverview = ({
+  createWorkflow,
+  panelWidth,
```
tonytang-microsoft-com, line 22: Do we only support 100% and 50% for the two scenarios we are planning to support? If this is a public contract, is it possible to have a boolean for the two supported scenarios, just to make the public API easier to usage.

libs/designer/src/lib/ui/templates/templateview.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { setLayerHostSelector, Spinner, SpinnerSize, Text } from '@fluentui/react';
+import { CreateWorkflowPanel } from '../panel/templatePanel/createWorkflowPanel/createWorkflowPanel';
+import { QuickViewPanel } from '../panel/templatePanel/quickViewPanel/quickViewPanel';
+import { openCreateWorkflowPanelView, openQuickViewPanelView } from '../../core/state/templates/panelSlice';
+import { useIntl } from 'react-intl';
+
+export interface TemplateViewProps {
+  createWorkflow: CreateWorkflowHandler;
+  showSummary?: boolean;
+  showCloseButton?: boolean;
```
tonytang-microsoft-com, line 16: Do we need these parameters? I thought for view template scenario, these values are fixed, so could we not expose them?

libs/designer/src/lib/core/state/templates/manifestSlice.ts
```diff
@@ -7,6 +7,23 @@ import type { FilterObject } from '@microsoft/designer-ui';
 export const templatesCountPerPage = 25;
 const initialPageNum = 0;
 
+interface ContentInfo<T> {
+  value: T;
+  canOverride?: boolean;
```
tonytang-microsoft-com, line 12: nit: Seems the term override used here is different from the meaning in basicOverride. I read here it means if cx can override it instead of the caller, so can we rename it to isReadOnly?: boolean?

tonytang-microsoft-com: Or name is canUserOverride:: boolean (just any different name is fine) if you want to have a different behavior for the default value.

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/usePanelTabs.tsx
```diff
@@ -23,7 +23,13 @@ import { closePanel } from '../../../../core/state/templates/panelSlice';
 export const useCreateWorkflowPanelTabs = ({
   isMultiWorkflowTemplate,
   createWorkflow,
-}: { createWorkflow: CreateWorkflowHandler; isMultiWorkflowTemplate: boolean }): TemplatePanelTab[] => {
+  showCloseButton = true,
+}: {
+  createWorkflow: CreateWorkflowHandler;
+  isMultiWorkflowTemplate: boolean;
+  showCloseButton?: boolean;
+  onClosePanel?: () => void;
```
Elaina-Lee, line 31: This `onClosePanel` props that is defined here is not used

Elaina-Lee (changes requested):

libs/designer/src/lib/core/state/templates/manifestSlice.ts
Elaina-Lee: @tonytang-microsoft-com The reason why we chose the term "override" here is because when provided, we are overriding the current value in templates. The provided ViewTemplateDetails value is not the "default" value since the actual "default" value may be present as part of templates. Since this is part of our redux store, we are trying to keep it consistent with our other structures.

libs/designer/src/lib/ui/templates/templateoverview.tsx
Elaina-Lee: At the moment, the reason why we have this is not to provide flexibility but is a workaround for overflowing. As you may see, we have used such values as 98%. We have an approximate idea of what the panel size should always look like, and I'm unsure whether we want to give it to the public to decide. With unverified numbers provided, the panel may not look ideal (overflow, not fitting all UI, etc).

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/basicsTab.tsx
```diff
@@ -47,6 +48,8 @@ export const basicsTab = (
       if (shouldClearDetails) {
         dispatch(clearTemplateDetails());
       }
+
+      onClose?.();
```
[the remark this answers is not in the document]
preetriti1: I see.... Then I need to verify where if the secondary button is close to call on close()

We have anyways removed the default dismiss and close button so on dismiss won't get called for this scenario

apps/Standalone/src/templates/state/WorkflowLoader.ts
```diff
@@ -13,6 +13,7 @@ export interface WorkflowLoadingState {
   isLocal: boolean;
   theme: ThemeType;
   language: string;
+  templatesView: string;
```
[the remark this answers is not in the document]
preetriti1: Ooh this also contains the template ids meaning names hence this is strong.... Default is gallery

libs/designer/src/lib/core/templates/TemplatesDataProvider.tsx
```diff
@@ -41,12 +45,6 @@ const DataProviderInner = ({ customTemplates, isConsumption, children }: Templat
     }
   }, [dispatch, githubTemplateNames]);
 
-  useEffect(() => {
-    if (customTemplates) {
-      dispatch(setCustomTemplates(customTemplates));
-    }
-  }, [dispatch, customTemplates]);
-
   useEffect(() => {
```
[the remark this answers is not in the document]
preetriti1: I see.. when we change to local in standalone this can change... Let me separate it out

libs/designer/src/lib/core/state/templates/manifestSlice.ts
preetriti1: I just changed the name to isEditable to make it more readable... we can keep the default behavior as disabled unless specified. [This PR is not covering that logic yet]

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -16,19 +16,35 @@ export const getQuickViewTabs = (
   dispatch: AppDispatch,
   workflowId: string,
   showCreate: boolean,
-  { templateId, workflowAppName, isMultiWorkflow }: Template.TemplateContext
+  { templateId, workflowAppName, isMultiWorkflow }: Template.TemplateContext,
+  onCloseButtonClick?: () => void
```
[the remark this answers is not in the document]
preetriti1: I missed passing the prop, now fixed all close button paths

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/parametersTab.tsx
```diff
@@ -14,7 +14,8 @@ export const ParametersPanel: React.FC = () => {
 export const parametersTab = (
   intl: IntlShape,
   dispatch: AppDispatch,
-  { isCreating, shouldClearDetails, previousTabId, hasError }: CreateWorkflowTabProps
+  { isCreating, shouldClearDetails, previousTabId, hasError }: CreateWorkflowTabProps,
+  onClose?: () => void
```
Elaina-Lee, line 18: onClose is present everytime CreateWorkflowTabProps is present. I would suggest that onClose be part of CreateWorkflowTabProps

Elaina-Lee (approved):

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -130,7 +130,8 @@ export const summaryTab = (
   dispatch: AppDispatch,
   workflowId: string,
   clearDetailsOnClose: boolean,
-  { templateId, workflowAppName, isMultiWorkflow }: Template.TemplateContext
+  { templateId, workflowAppName, isMultiWorkflow }: Template.TemplateContext,
+  onClose?: () => void
```
tonytang-microsoft-com, line 134: Do you expect the 'disable close' capability to be added separately ?

libs/designer/src/lib/core/state/templates/manifestSlice.ts
```diff
@@ -7,6 +7,23 @@ import type { FilterObject } from '@microsoft/designer-ui';
...
+export interface ViewTemplateDetails {
+  id: string;
+  basicsOverride?: Record<
+    string,
+    {
+      name?: ContentInfo<string>;
+      kind?: ContentInfo<string>;
+    }
+  >;
+  parametersOverride?: Record<string, ContentInfo<any>>;
```
tonytang-microsoft-com, line 24: It may be worth to have a doc outlining the contract for different type that is supported. From what I found in template repo, if the parameter type is int, the value should be a string like "123", so assuming here we want to stick to this contract to make it work. If this is not the case, maybe call out the types we tested and supported, while we fix this.

libs/designer/src/lib/core/state/templates/manifestSlice.ts
```diff
@@ -7,6 +7,23 @@ import type { FilterObject } from '@microsoft/designer-ui';
...
 const initialPageNum = 0;
 
+interface ContentInfo<T> {
+  value: T;
+  isEditable?: boolean;
+}
+
+export interface ViewTemplateDetails {
+  id: string;
+  basicsOverride?: Record<
```
tonytang-microsoft-com, line 17: This only applies to non accesslator template? Do we want to make this a bit more future proof, or maybe there's some other story in mind?

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/usePanelTabs.tsx
```diff
@@ -166,57 +175,75 @@ export const useCreateWorkflowPanelTabs = ({
...
         hasError: Object.values(workflows).some((workflowData) => workflowData.errors.kind || workflowData.errors.workflow),
         isCreating,
-      }),
-    }),
-    [intl, dispatch, isMultiWorkflowTemplate, connectionsExist, parametersExist, workflows, isCreating]
-  );
+      },
+      onClosePanel
+    );
+    tab.footerContent.secondaryButtonDisabled = !showCloseButton;
```
tonytang-microsoft-com, line 181: I think this is a bit confusing for the author who write basicsTab method above. This property is overriden after the author returned this property causing the behavior is not what the author expected. Also this line assumed that the author uses the second button for the close button. I would suggest to pass showCloseButton to be passed to the basicsTab method and handled over there.

libs/designer/src/lib/ui/templates/templateview.tsx
tonytang-microsoft-com: I see, sounds good.

libs/designer/src/lib/core/state/templates/manifestSlice.ts
preetriti1: We should ideally fix the parameter value either here or in parameters UI, it can be done as part of impoementing the overrides

libs/designer/src/lib/core/state/templates/manifestSlice.ts
preetriti1: this applies to all templates even accelerators, we can specify which workflow to override here

PR #6575, opened 2025-02-10
feat(Templates): Support ViewTemplate data Override

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## New Behavior

- overrides basics & parameters info when present from viewTemplateData
- Moved ViewTemplateDetails interface to template model file for reusability
- Using viewTemplateDeatails only when currentTemplateName === viewTemplateDetails.id (extra protection)
- kind is set to viewTemplateDetails kind only when following is present from the template manifest

## Impact of Change

## Test Plan

e2e tests on prefilled data on viewTemplate selection

https://github.com/user-attachments/assets/3a19e529-bc88-49c7-a0ca-b3f25e9039df

### DIFF
--- a/apps/Standalone/src/templates/app/LocalTemplates.tsx
+++ b/apps/Standalone/src/templates/app/LocalTemplates.tsx
@@ -72,9 +72,30 @@ export const LocalTemplates = () => {
         services={services}
         isConsumption={isConsumption}
         isCreateView={!isConsumption}
-        viewTemplate={isSingleTemplateView ? { id: templatesView } : undefined}
         customTemplates={localManifests}
         existingWorkflowName={undefined}
+        viewTemplate={
+          isSingleTemplateView
+            ? {
+                id: templatesView,
+                parametersOverride: {
+                  'OpenAIEmbeddingModel_#workflowname#': { value: 'overriden-default-editable' },
+                  'OpenAIChatModel_#workflowname#': { value: 'overriden-default-non-editable', isEditable: false },
+                  'LogicMessage_#workflowname#': { value: 'overriden-default-non-editable', isEditable: false },
+                },
+                basicsOverride: {
+                  ['SimpleParametersOnly']: {
+                    name: { value: 'overriden-name', isEditable: false },
+                    kind: { value: 'stateful', isEditable: false },
+                  },
+                  ['Workflow1']: {
+                    name: { value: 'overriden-name', isEditable: false },
+                    kind: { value: 'stateful', isEditable: false },
+                  },
+                },
+              }
+            : undefined
+        }
       >
         <div
           style={{
--- a/apps/Standalone/src/templates/app/LocalTemplates.tsx
+++ b/apps/Standalone/src/templates/app/LocalTemplates.tsx
@@ -84,7 +84,7 @@ export const LocalTemplates = () => {
                   'LogicMessage_#workflowname#': { value: 'overriden-default-non-editable', isEditable: false },
                 },
                 basicsOverride: {
-                  ['SimpleParametersOnly']: {
+                  ['default']: {
                     name: { value: 'overriden-name', isEditable: false },
                     kind: { value: 'stateful', isEditable: false },
                   },
--- a/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
@@ -235,12 +235,17 @@ const loadTemplateFromResourcePath = async (
       }
     }
   } else {
-    const workflowId = templateName;
-    const workflowData = await loadWorkflowTemplateFromManifest(workflowId, templateName, manifest, isCustomTemplate, viewTemplateData);
+    const workflowData = await loadWorkflowTemplateFromManifest(
+      /* workflowId */ templateName,
+      templateName,
+      manifest,
+      isCustomTemplate,
+      viewTemplateData
+    );
 
     if (workflowData) {
       data.workflows = {
-        [workflowId]: workflowData.workflow,
+        [templateName]: workflowData.workflow,
       };
       data.parameterDefinitions = workflowData.parameterDefinitions;
       data.connections = workflowData.connections;
--- a/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
@@ -271,7 +271,7 @@ const loadWorkflowTemplateFromManifest = async (
     const parameterDefinitions = templateManifest.parameters?.reduce((result: Record<string, Template.ParameterDefinition>, parameter) => {
       result[parameter.name] = {
         ...parameter,
-        value: viewTemplateData?.parametersOverride?.[parameter.name]?.value ?? parameter.default,
+        value: viewTemplateData?.parametersOverride?.[parameter.name]?.value?.toString() ?? parameter.default,
         associatedWorkflows: [templateManifest.title],
       };
       return result;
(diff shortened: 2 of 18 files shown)


libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -57,3 +57,20 @@ export interface TemplateContext {
   workflowAppName?: string;
   isMultiWorkflow: boolean;
 }
+
+interface ContentInfo<T> {
```
tonytang-microsoft-com, line 61: can we export this?

libs/designer/src/lib/ui/templates/basics/multiworkflow.tsx
```diff
@@ -79,7 +82,9 @@ export const MultiWorkflowBasics = () => {
     Object.values(workflows).map((workflow) => ({
       id: workflow.id,
       name: workflow.workflowName,
+      isNameEditable: basicsOverrideByWorkflow?.[workflow.id]?.name?.isEditable ?? true,
       kind: workflow.kind ?? (workflow.manifest.kinds?.length ? workflow.manifest.kinds[0] : WorkflowKind.STATEFUL),
+      isKindEditable: basicsOverrideByWorkflow?.[workflow.id]?.kind?.isEditable ?? true,
```
tonytang-microsoft-com, line 87: When kind.value is provided and isEditable is missing, we should default to false.

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
```diff
@@ -136,10 +137,11 @@ export const DisplayParameters = () => {
         return (
           <TextField
             className="msla-templates-parameters-values"
-            data-testid="msla-templates-parameter-value"
-            id="msla-templates-parameter-value"
+            data-testid={`msla-templates-parameter-value-${item.name}`}
+            id={`msla-templates-parameter-value-${item.name}`}
             aria-label={item.value}
             value={item.value}
+            disabled={parametersOverride?.[item.name]?.isEditable === false}
```
tonytang-microsoft-com, line 144: same here. when value of foo is overriden like below, the parameter should be not editable. (The default value of a boolean is false, so when omitted, it's the same semantic as if the value is set to false)

parametersoverride: {
    "foo": {
        value: "value"
    }
}

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -126,14 +125,19 @@ export const initializeTemplateServices = createAsyncThunk(
...
+      preLoadedManifest,
+      viewTemplateDetails,
+      isCustomTemplate = false,
+    }: { preLoadedManifest: Template.Manifest | undefined; viewTemplateDetails?: Template.ViewTemplateDetails; isCustomTemplate?: boolean },
     thunkAPI
   ) => {
     const currentState: RootState = thunkAPI.getState() as RootState;
-    const currentTemplateResourcePath = currentState.template.templateName;
+    const currentTemplateName = currentState.template.templateName;
+    const viewTemplateData = currentTemplateName === viewTemplateDetails?.id ? viewTemplateDetails : undefined;
```
tonytang-microsoft-com, line 137: curious in what scenario will current template name !== viewTemplateDetails.id ? I expect if viewTemplateDetails is provided, id should always match current.

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
tonytang-microsoft-com: Can you elaborate on this @preetriti1 ? This is a public contract should be able to be referenced right?

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
tonytang-microsoft-com: IMHO, have the boolean type default to false is the natural contract that people are following, so I think it's better to keep following this pattern. If the you want to make it such the lack of the property means it's not editable, I'd suggest to rename this parameter to isReadOnly.

If we want to have default value to be true, it needs to be call out in the doc as a minimum, because it's the opposite of the most easy to understand contract.

Elaina-Lee: As we have discussed before, isReadOnly would not be a good naming since we are overriding the value - so it is not a good representitive of what it does. Though this concept is coming from the AI context, we are trying to keep it generic, and overriding isEditable with a default false value may not  be a good practice.

I understand that this may be a concern for you and would be open to exploring more options in our set meeting.

libs/designer/src/lib/ui/templates/basics/multiworkflow.tsx
Elaina-Lee: Leaving duplicate discussion to one

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -229,11 +235,17 @@ const loadTemplateFromResourcePath = async (
...
+      templateName,
+      manifest,
+      isCustomTemplate,
+      viewTemplateData
+    );
+
     if (workflowData) {
       data.workflows = {
-        [workflowId]: workflowData.workflow,
+        [templateName]: workflowData.workflow,
```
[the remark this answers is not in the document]
Elaina-Lee: workflowId was guid() before, and this was not reading the workflow correctly in case of accelerators. templateName (whether single or accelerators, this is the direct folder name) is the workflowId in case of accelerators in our store.

Without changing this, the basics tab override was not getting picked up for accelerator.

Even for single template, after custom template restructure, this should be the way - unless there is concern here.

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
Elaina-Lee: It should always match, this is just an extra precaution.

libs/designer/src/lib/ui/templates/parameters/displayParameters.tsx
tonytang-microsoft-com: As long you've considered this feedback and explicitly decided to go with isEditable default to true for a reason, it works for me :-) Let's keep it as is then.

apps/Standalone/src/templates/app/LocalTemplates.tsx
```diff
@@ -72,9 +72,30 @@ export const LocalTemplates = () => {
...
+                  'OpenAIEmbeddingModel_#workflowname#': { value: 'overriden-default-editable' },
+                  'OpenAIChatModel_#workflowname#': { value: 'overriden-default-non-editable', isEditable: false },
+                  'LogicMessage_#workflowname#': { value: 'overriden-default-non-editable', isEditable: false },
+                },
+                basicsOverride: {
+                  ['SimpleParametersOnly']: {
+                    name: { value: 'overriden-name', isEditable: false },
+                    kind: { value: 'stateful', isEditable: false },
+                  },
+                  ['Workflow1']: {
```
tonytang-microsoft-com, line 91: Quick question, how would caller know what is the name here for single workflow? Is it always 'default', or something else that is well known by the caller?

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
preetriti1: I can explain if you do that in datafetcher :)

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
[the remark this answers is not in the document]
preetriti1: Please name the workflowId as default, i am changing to that in my next set of changes.
Think about the restructure, after that we will have a workflow folder having workflow artifacts, so we will keep that folder name as "default".

preetriti1 (approved):

PR #6583, opened 2025-02-10
feat(Designer): Local caching of manifests / connectors / swaggers

rllyy97:
## Main Changes

ReactQuery now has the ability to cache results locally.
This data is compressed and cached to local browser storage.
By default we are caching swaggers, connectors, manifests, with a 24 hour timout.
Our query client has new options for `persistEnabled` and `persistKeyWhitelist` to add more query keys to whitelist for persisting.
Anything else is not persisted by default, this is intended to be expanded on later but for now these are the easy ones.

This is a minor startup perf improvement, that will work across all workflows.
If any swagger / connector / manifest is already stashed locally, we won't need to refetch it on startup.

> Also tweaked some of our standalone components, and fixed a few minor build warnings

> [!NOTE]
>**TESTS:** Added E2E test for query caching
>**LOW RISK:** This is a non-breaking feature, hidden behind a parameter on the query client


Eric-B-Wu:
Just wondering, if something were to change on the backend, would it still be going off of the cached results to populate manifests/connectors/ swaggers?

rllyy97:
> Just wondering, if something were to change on the backend, would it still be going off of the cached results to populate manifests/connectors/ swaggers?

Yes it would still go off of the cached files since we wouldn't be making any new requests, I couldn't really think of a way around that.  Although, the swaggers / manifests / connectors don't really change super often so I really only see this being an issue if there is an urgent change, like if a new version needed to get rolled back.

preetriti1:
> > Just wondering, if something were to change on the backend, would it still be going off of the cached results to populate manifests/connectors/ swaggers?
> 
> Yes it would still go off of the cached files since we wouldn't be making any new requests, I couldn't really think of a way around that. Although, the swaggers / manifests / connectors don't really change super often so I really only see this being an issue if there is an urgent change, like if a new version needed to get rolled back.

Is there a possibility of giving user the ability to hard refresh maybe clear browser cookies or something like that?.. so if in case they are running into any issue and if something is deployed in connector or backend we need to have a way to refresh it. Every service change is not accompanied by portal deployment.

hartra344:
> > > Just wondering, if something were to change on the backend, would it still be going off of the cached results to populate manifests/connectors/ swaggers?
> > 
> > 
> > Yes it would still go off of the cached files since we wouldn't be making any new requests, I couldn't really think of a way around that. Although, the swaggers / manifests / connectors don't really change super often so I really only see this being an issue if there is an urgent change, like if a new version needed to get rolled back.
> 
> Is there a possibility of giving user the ability to hard refresh maybe clear browser cookies or something like that?.. so if in case they are running into any issue and if something is deployed in connector or backend we need to have a way to refresh it. Every service change is not accompanied by portal deployment.

I don't think this will be a super large issue unless there is an urgent hotfix on the backend that changes something. Every change, even most hotfixes need to go through SDP. This only slows things down by an extra day where a user may be using a cached version since there's a 24 hour timeout. That said, maybe we could but a cache clear button somewhere that isn't in the way, maybe the "info" blade from the command bar?

hartra344 (approved):

ccastrotrejo (approved):

PR #6621, opened 2025-02-18
Feat(designer): Multi-InitializeVariable Editor Support

Eric-B-Wu:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior

We don't support Multivariable, users must create multiple variable actions.

## New Behavior

- Multivariable editor to support creating multiple variables in one action

- An modal option to combine existing variables into a single action

## Impact of Change

Marking this as a breaking change as we'll need to set the host options on portal repo, this feature will be bundle version specific (only backend versions > x.x.xx will support this) and thus inorder to not break experience, this will require a sister PR

## Test Plan

 
Unit tests and E2E tests have been added

multivariableCombine
Combining existing variables together

variableEditor
Variable Editor


libs/designer/src/lib/core/state/tokens/tokensSlice.ts
```diff
@@ -55,24 +55,9 @@ export const tokensSlice = createSlice({
...
-      }
-      if (type) {
-        state.variables[id] = variables.map((variable) => ({
-          ...variable,
-          type,
-        }));
-      }
+    updateVariableInfo: (state, action: PayloadAction<{ id: string; variables: VariableDeclaration[] }>) => {
+      const { id, variables } = action.payload;
+      state.variables[id] = variables;
```
rllyy97, line 60: Can we use `getRecordEntry()` here again

libs/designer/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
```diff
@@ -241,22 +243,27 @@ const ParameterSection = ({
...
+      if (parameter && isInitializeVariableParameter(parameter)) {
+        const variables: InitializeVariableProps[] | undefined = newState?.viewModel?.variables;
+        if (variables) {
+          dispatch(
+            updateVariableInfo({
+              id: nodeId,
+              variables: variables.map((variable) => {
+                return {
+                  name: variable.name[0].value,
+                  type: variable.type[0].value,
```
rllyy97, line 258: Do these need to be optional like they were previously?

rllyy97 (approved):
Nice work!!  Just had a few small things but otherwise looks great!

libs/designer/src/lib/core/state/tokens/tokensSlice.ts
Eric-B-Wu: So the difference before was that we were updating the variables to point to the right name/type for downstream nodes whenever the corresponding parameter changes; however, now because the payload is not what changed but the variables themselves, we don't actually access any variables in the state here, but rather just replace them altogether

Eric-B-Wu: This may be something we might want to investigate separately, as there is something that may be a concern, because the parameter is a single Editor now, when creating a new variable and changing the name, the type validation also happens; that being said, I haven't thought of a great way of doing this just yet as validation happens on the parameter level, but we'd need to bring this down to an input level

Issue #6618, opened 2025-02-17
Flow checker error - unable to save in the designer mode


m00nchaser:
![Image](<https://github.com/user-attachments/assets/556397dc-27a2-4b74-a638-8042e03da51f>)

### Describe the Bug with repro steps 

Data operation "Select" is triggering an error (invalid Json) while saving in the designer mode when the mapping has a replace function.

### What type of Logic App Is this happening in? 

Consumption (Portal)

### Which operating system are you using? 

Windows

### Did you refer to the TSG before filing this issue? [https://aka.ms/lauxtsg](<https://aka.ms/lauxtsg>) 

Yes

### Workflow JSON 

```json
{
  "type": "Select",
  "inputs": {
    "from": "@body('Parse_label_translations_(LabelTranslations)_uk')?['value']",
    "select": {
      "DESCRIPTION": "@replace (replace(replace(item()?['Description'], '\\', '\\\\'), '\"', '\\\"'), '   ', '\\n')",
      "LANGUAGEID": "@item()?['LanguageId']",
      "PRODUCTNUMBER": "@item()?['ProductNumber']"
    }
  },
  "runAfter": {
    "Parse_label_translations_(LabelTranslations)_uk": [
      "Succeeded"
    ]
  }
}
```

### Screenshots or Videos

*No response*

### Browser

Edge Version 133.0.3065.69

### Additional context

*No response*

PR #6636, opened 2025-02-19
fix(Designer): Fixes an issue where '\"'  causing json errors in expressions

Eric-B-Wu:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Fixes: https://github.com/Azure/LogicAppsUX/issues/6618
## New Behavior
User should always have the same expression, whether its opening an existing expression, saving a new expression, viewing it on the code view, etc. regardless of escape character - all of which should be handled behind the scenes
 

## Impact of Change

## Test Plan

 
E2E test and Unit tests have been added

### DIFF
--- a/libs/designer-ui/src/lib/tokenpicker/index.tsx
+++ b/libs/designer-ui/src/lib/tokenpicker/index.tsx
@@ -165,9 +165,7 @@ export function TokenPicker({
   }, [anchorKey, editor, windowDimensions.height]);
 
   const handleInitializeExpression = (s: string, n: NodeKey) => {
-    console.log(s);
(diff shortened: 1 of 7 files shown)


libs/designer-ui/src/lib/tokenpicker/index.tsx
```diff
@@ -165,7 +165,9 @@ export function TokenPicker({
   }, [anchorKey, editor, windowDimensions.height]);
 
   const handleInitializeExpression = (s: string, n: NodeKey) => {
+    console.log(s);
```
hartra344, line 168: remove

hartra344 (approved):

PR #6644, opened 2025-02-20
feat(templates): Moving templates loading from template service via storage account

preetriti1:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior
Currently we are packaging all templates as part of build which restricts us to start a designer/portal build for any template updates.

## New Behavior
We will be using template service to fetch all templates/artifacts and manifests and this in turn will use the storage account.
This is a change how we load the templates but for end user there shouldn't be any difference.

* [x] **This is a breaking change.**

We will also do manual testing after portal integration to see all templates are loaded correctly.
>**HIGH RISK:** Breaking change, since we are not packaging templates now and loading from a service. To integrate this package we will have to have a portal PR before doing any test deployments else templates will be broken.

### DIFF
--- a/apps/Standalone/src/templates/app/TemplatesConsumption.tsx
+++ b/apps/Standalone/src/templates/app/TemplatesConsumption.tsx
@@ -333,6 +333,8 @@ const getServices = (
   };
 
   const templateService = new BaseTemplateService({
+    httpClient,
+    endpoint: 'https://priti-cxf4h5cpcteue4az.b02.azurefd.net',
     openBladeAfterCreate: (_workflowName: string | undefined) => {
       window.alert('Open blade after create, consumption creation is complete');
     },
--- a/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
@@ -18,8 +18,8 @@ import {
   LogEntryLevel,
   LoggerService,
   type LogicAppsV2,
-  type Template,
   TemplateService,
+  type Template,
 } from '@microsoft/logic-apps-shared';
 import { createAsyncThunk } from '@reduxjs/toolkit';
 import type { RootState } from '../../state/templates/store';
@@ -140,12 +140,12 @@ export const initializeTemplateServices = createAsyncThunk(
 export const loadManifestsFromPaths = async (resourcePaths: string[]) => {
   try {
     const manifestPromises = resourcePaths.map(async (resourcePath) => {
-      return import(`./../../templates/templateFiles/${resourcePath}/manifest.json`);
+      return TemplateService().getResourceManifest(resourcePath);
     });
     const manifestsArray = await Promise.all(manifestPromises);
     return manifestsArray.reduce((result: Record<string, Template.Manifest>, manifestFile: any, index: number) => {
-      const manifest = manifestFile.default;
-      result[resourcePaths[index]] = manifest;
+      //const manifest = manifestFile.default;
+      result[resourcePaths[index]] = manifestFile;
       return result;
     }, {});
   } catch (error) {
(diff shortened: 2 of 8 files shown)


Elaina-Lee (approved):

libs/logic-apps-shared/src/designer-client-services/lib/template.ts
```diff
@@ -1,10 +1,15 @@
+import type { LogicAppsV2, Template } from '../../utils/src';
 import { AssertionErrorCode, AssertionException } from '../../utils/src';
 
 export interface ITemplateService {
   getExistingWorkflowNames?: () => Promise<string[]>;
   openBladeAfterCreate: (workflowName: string | undefined) => void;
   onAddBlankWorkflow: () => Promise<void>;
   getCustomResource?: (resourcePath: string, artifactType?: string) => Promise<any> | undefined;
+  getAllTemplateNames: () => Promise<string[]>;
```
Elaina-Lee, line 9: Are we sure we want this to be 'getAllTemplateNames' since this won't include the custom template names? I wonder if we should consider renaming this to account for it (i.e. in our template code right now, we use githubtemplateNames)

Or are we supporting passing custom template names as well through this function?

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
```diff
@@ -18,4 +22,30 @@ export class BaseTemplateService implements ITemplateService {
   public onAddBlankWorkflow = (): Promise<void> => this.options.onAddBlankWorkflow();
 
   public getCustomResource = (resourcePath: string): Promise<any> | undefined => this.options?.getCustomResource?.(resourcePath);
+
+  public getContentPathUrl: (resourcePath: string) => string = (resourcePath: string) =>
+    `${this.options.endpoint}/templates/${resourcePath}`;
```
Elaina-Lee, line 27: Since we have `/templates` for all endpoints, could we consider putting this as a constant?

Elaina-Lee (approved):

libs/logic-apps-shared/src/designer-client-services/lib/template.ts
preetriti1: I have removed the custom templates sections since the host would implement these functions and pass in their own templates.

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -140,12 +140,12 @@ export const initializeTemplateServices = createAsyncThunk(
...
   try {
     const manifestPromises = resourcePaths.map(async (resourcePath) => {
-      return import(`./../../templates/templateFiles/${resourcePath}/manifest.json`);
+      return TemplateService().getResourceManifest(resourcePath);
     });
     const manifestsArray = await Promise.all(manifestPromises);
     return manifestsArray.reduce((result: Record<string, Template.Manifest>, manifestFile: any, index: number) => {
-      const manifest = manifestFile.default;
-      result[resourcePaths[index]] = manifest;
+      //const manifest = manifestFile.default;
```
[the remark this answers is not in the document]
preetriti1: nah just forgot to remove this.

apps/Standalone/src/templates/app/TemplatesConsumption.tsx
```diff
@@ -333,6 +333,8 @@ const getServices = (
   };
 
   const templateService = new BaseTemplateService({
+    httpClient,
+    endpoint: 'https://priti-cxf4h5cpcteue4az.b02.azurefd.net',
```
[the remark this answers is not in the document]
preetriti1: Sorry didnt get this.... eventually we will move this to a test url and since it is public and only read only access that should  be fine. Can check with Travis later on. Right now unless we have the accounts setup cannot check these in yet.

Elaina-Lee (approved):

Elaina-Lee (approved):

PR #6645, opened 2025-02-20
fix(vscode): bug where DateTime properties have invalid type in generated unit…

andrew-eldridge:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Generated C# mock output classes for unit testing have string props instead of DateTime for props with date-time format.

## New Behavior

Pass format to type conversion method, handle date-time case

## Impact of Change

## Test Plan

Added unit test for "mapJsonTypeToCSharp", "generateCSharpClasses", "generateClassCode", "buildClassDefinition" to ensure proper conversion for different types

N/A

### DIFF
--- a/apps/vs-code-designer/src/app/utils/unitTests.ts
+++ b/apps/vs-code-designer/src/app/utils/unitTests.ts
@@ -752,7 +752,7 @@ export function buildClassDefinition(className: string, node: any): ClassDefinit
       const propName = toPascalCase(key);
 
       // Determine the child's C# type
-      let csharpType = mapJsonTypeToCSharp(subNode?.type);
+      let csharpType = mapJsonTypeToCSharp(subNode?.type, subNode?.format);
       let isObject = false;
 
       // If it's an object, we must generate a nested class.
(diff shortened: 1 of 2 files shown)


apps/vs-code-designer/src/app/utils/unitTests.ts
```diff
@@ -848,6 +848,8 @@ export function generateClassCode(classDef: ClassDefinition): string {
   for (const prop of classDef.properties) {
     if (prop.propertyType === 'string') {
       sb.push(`        ${prop.propertyName} = string.Empty;`);
+    } else if (prop.propertyType === 'DateTime') {
```
ccastrotrejo, line 851: Can you add a new test cases for this change as well?

apps/vs-code-designer/src/app/utils/unitTests.ts
```diff
@@ -976,19 +978,10 @@ export function generateCSharpClasses(namespaceName: string, rootClassName: stri
...
-    'using System.Collections.Generic;',
-    'using System.Net;',
-    '',
-    `namespace ${adjustedNamespace}`,
-    '{',
-    classCode,
-    '}',
-  ].join('\n');
-  return finalCode;
+  return [...requiredNamespaces.map((ns) => `using ${ns};`), '', `namespace ${adjustedNamespace}`, '{', classCode, '}'].join('\n');
```
ccastrotrejo, line 984: Does this change also changes the expected return value, if yes could you please update/add a test?

ccastrotrejo (changes requested):
Wrote down some comments

apps/vs-code-designer/src/app/utils/unitTests.ts
andrew-eldridge: updated test to check namespaces (return value is the same except System was added)

apps/vs-code-designer/src/app/utils/unitTests.ts
andrew-eldridge: thanks for catching this - updated tests for all functions changed

ccastrotrejo (approved):

PR #6685, opened 2025-02-26
feat(Templates): Custom Template Standalone

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [ ] Feature
* [ ] Other

## Impact of Change

### DIFF
--- a/apps/Standalone/src/App.tsx
+++ b/apps/Standalone/src/App.tsx
@@ -16,6 +16,7 @@ export const App = () => {
       <Route path="/datamapperv1" element={<DataMapperV1 />} />
       <Route path="/datamapperv2" element={<DataMapperV2 />} />
       <Route path="/templates" element={<TemplatesStandalone />} />
+      <Route path="/customtemplates" element={<CustomTemplatesStandalone />} />
       {/* Using path="*"" means "match anything", so this route
                 acts like a catch-all for URLs that we don't have explicit
                 routes for. */}
@@ -49,6 +50,14 @@ const DataMapperV2 = () => {
 };
 
 const TemplatesStandalone = () => {
+  return (
+    <Provider store={templateStore}>
+      <TemplatesWrapper />
+    </Provider>
+  );
+};
+
+const CustomTemplatesStandalone = () => {
   return (
     <Provider store={templateStore}>
       <TemplatesWrapper />
--- a/apps/Standalone/src/customTemplates/app/CustomTemplatesShell.tsx
+++ b/apps/Standalone/src/customTemplates/app/CustomTemplatesShell.tsx
@@ -1,0 +1,21 @@
+import type { ReactNode } from 'react';
+import { ReactQueryProvider } from '@microsoft/logic-apps-designer';
+import { useQuery } from '@tanstack/react-query';
+import { loadToken } from '../../environments/environment';
+import { LocalCustomTemplates } from './LocalCustomTemplates';
+
+const LoadWhenArmTokenIsLoaded = ({ children }: { children: ReactNode }) => {
+  const { isLoading } = useQuery(['armToken'], loadToken);
+  return isLoading ? null : <>{children}</>;
+};
+export const CustomTemplatesWrapper = () => {
+  return (
+    <ReactQueryProvider>
+      <LoadWhenArmTokenIsLoaded>
+        <div style={{ height: '100vh' }}>
+          <LocalCustomTemplates /> {/* local only for now until template resource is created */}
+        </div>
+      </LoadWhenArmTokenIsLoaded>
+    </ReactQueryProvider>
+  );
+};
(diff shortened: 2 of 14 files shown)


preetriti1:
Can we call it TemplateGenerator or CreateTemplateWizard

libs/logic-apps-shared/src/designer-client-services/lib/templateresource.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const InitTemplateResourceService = (templateService: ITemplateResourceService): void => {
+  service = templateService;
+};
+
+export const TemplateResourceService = (): ITemplateResourceService => {
+  if (!service) {
+    throw new AssertionException(
+      AssertionErrorCode.SERVICE_NOT_INITIALIZED,
+      'Template Resource 0Service needs to be initialized before using'
```
rllyy97, line 26: Nit: Looks like a `0` is in this string on accident

rllyy97 (approved):

apps/Standalone/src/configuretemplate/app/ConfigureTemplateShell.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const LoadWhenArmTokenIsLoaded = ({ children }: { children: ReactNode }) => {
+  const { isLoading } = useQuery(['armToken'], loadToken);
+  return isLoading ? null : <>{children}</>;
+};
+export const ConfigureTemplateWrapper = () => {
+  return (
+    <ReactQueryProvider>
+      <LoadWhenArmTokenIsLoaded>
+        <DevToolbox />
```
Elaina-Lee, line 16: Not a blocker, but I don't seem to be able to see this devToolBox on standalone at all.

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public async updateWorkflow(_id: string, _data: WorkflowData) {
+    throw new Error('Method not implemented.');
+  }
+
+  public async createArtifact(_id: string, _artifact: Template.Artifact) {
+    throw new Error('Method not implemented.');
+  }
+
+  public async isWorkflowNameAvailable(_id: string, _name: string) {
```
Elaina-Lee, line 26: What is this call for? Are we checking through the resource whether the workflow name the creator chooses is available?

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public async updateTemplate(_id: string, _manifest: Template.TemplateManifest) {
+    throw new Error('Method not implemented.');
+  }
+
+  public async updateWorkflow(_id: string, _data: WorkflowData) {
+    throw new Error('Method not implemented.');
+  }
+
+  public async createArtifact(_id: string, _artifact: Template.Artifact) {
```
Elaina-Lee, line 22: Just making sure - this will likely not be a part of the standalone since the only blade that exists in standalone is the template creation right?

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
```diff
@@ -0,0 +1,29 @@
...
+interface ITemplateResourceServiceOptions {
+  httpClient: IHttpClient;
+  baseUrl: string;
+  apiVersion: string;
+}
+
+export class BaseTemplateResourceService implements ITemplateResourceService {
+  constructor(private readonly options: ITemplateResourceServiceOptions) {}
+
+  public async updateTemplate(_id: string, _manifest: Template.TemplateManifest) {
```
Elaina-Lee, line 14: We have `updateTemplate` but not `updateWorkflowManifest` - maybe can we consider the naming convention of this one to be `updateTemplateManifest` to be consistent if we introduce `updateWorkflowManifest`? Or do we have any other plans of updating Template.WorkflowManifest?

I'm thinking to consider accelartor, it may need to be `updateWorkflowManifests(Record<id, WorkflowManifest>)`

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
preetriti1: Yes... When we give the workflow name we need to check it's availability

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
preetriti1: It is a API call for creating artifacts in the artifact blade, which is part of the new API rama is creating.... This is only in service so we can use later

libs/logic-apps-shared/src/designer-client-services/lib/base/templateresource.ts
preetriti1: Sure will change that appropriately

apps/Standalone/src/configuretemplate/app/ConfigureTemplateShell.tsx
preetriti1: It shows up for me... Maybe hard refresh it

PR #6698, opened 2025-02-27
fix(vscode): Fix nuget debug issue introduced with workspace

lambrianmsft:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

With workspaces, the tasks.json files are created with the incorrect cwd after a project is converted to nuget.

## New Behavior

tasks.json file has the correct cwd after it is converted to nuget.

## Impact of Change

### DIFF
--- a/apps/vs-code-designer/src/app/commands/initProjectForVSCode/DotnetInitVSCodeStep.ts
+++ b/apps/vs-code-designer/src/app/commands/initProjectForVSCode/DotnetInitVSCodeStep.ts
@@ -101,7 +101,7 @@ export class DotnetInitVSCodeStep extends InitVSCodeStepBase {
     }
 
     const targetFramework: string = await getTargetFramework(projFile);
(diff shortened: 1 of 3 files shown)


apps/vs-code-designer/src/app/commands/initProjectForVSCode/DotnetInitVSCodeStep.ts
```diff
@@ -101,7 +101,7 @@ export class DotnetInitVSCodeStep extends InitVSCodeStepBase {
     }
 
     const targetFramework: string = await getTargetFramework(projFile);
-    this.setDeploySubpath(context, `bin/Release/${targetFramework}/publish`);
+    await this.setDeploySubpath(context, `bin/Release/${targetFramework}/publish`);
```
ccastrotrejo, line 104: can we use path.join()?

ccastrotrejo (changes requested):

ccastrotrejo (approved):

PR #6727, opened 2025-03-04
feat(designer): Introduce "Parse document with metadata" action for consumption

bjbennet:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## New Behavior
This introduces a new action named ParseDocumentWithMetadata for parsing a document outputting text and metadata for consumption.

## Impact of Change
Adds a new action to the consumption designer.

image

image

image


ccastrotrejo (approved):

libs/logic-apps-shared/src/designer-client-services/lib/consumption/manifests/parsedocumentwithmetadata.ts
```diff
@@ -0,0 +1,132 @@
...
+        description: 'Operations to work with data in workflow.',
+        displayName: 'Data Operations',
+      },
+    } as any,
+
+    settings: {
+      secureData: {},
+      trackedProperties: {
+        scopes: [SettingScope.Action],
+      },
```
preetriti1, line 129: Are these the only setting you want?

bjbennet: This matches the settings object in the other similar actions I've added. Do you see a potential problem with this configuration?

preetriti1: Ahh, i have seen many manifests just copy paste from previous ones but miss the settings they actually want for the action. If you only need these no issues. BUt I just want everyone to review it in UI what they actually require

preetriti1 (approved):

PR #6730, opened 2025-03-05
feat(Templates): Restructure to Reflect Custom Templates Restructuring

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

- No feature updates (No change in UI)
- Restructuring templates to align with manifest separation of TemplateManifest and WorkflowManifest
- Introduction: id for both manifests, description => summary, detailsDescription => description
- Mock data updates for LocalTemplates

## Impact of Change

Restructuring: Separating Template.Manifest into Template.TemplateManifest and Template.WorkflowManifest
- Disregarding of whether a wrkflow or accelerator template, it will follow the same way of containing manifests

## Test Plan

All previous tests using Template.Manifest has been updated and have been verified to do the same with new way of redux stores updates containing separate manifests


libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -1,29 +1,48 @@
 export type SkuType = 'standard' | 'consumption';
 export type WorkflowKindType = 'stateful' | 'stateless';
-export type ConnectorRuntimeType = 'inapp' | 'shared';
+export type ConnectorRuntimeType = 'inapp' | 'shared' | 'custom';
```
preetriti1, line 3: Why do you have custom here?

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
```diff
@@ -34,35 +34,39 @@ export class BaseTemplateService implements ITemplateService {
...
+        uri: `${endpoint}/templates/${templateId}/manifest.json`,
         headers: { 'Access-Control-Allow-Origin': '*' },
       });
     }
 
-    const paths = resourcePath.split('/');
-
-    return paths.length === 2
-      ? (await import(`./../templates/${paths[0]}/${paths[1]}/manifest.json`)).default
-      : (await import(`./../templates/${resourcePath}/manifest.json`)).default;
```
preetriti1, line 50: Did we create two different methods to avoid this path split check? We can just use resourceManifest<T> based on the type being returned.
We can create multiple methods inside designer helper functions/actions but in services it is better to allow less implementation from host. I dont see why we need two different functions to download from same path and same manifest.json

libs/designer/src/lib/ui/templates/templateoverview.tsx
```diff
@@ -157,25 +157,23 @@ export const TemplateOverview = ({
 interface WorkflowItem {
   id: string;
   name: string;
-  trigger: string;
```
preetriti1, line 160: Why this is removed after a template selection? This was always being read from workflow.json when template selection was happening no?

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
Elaina-Lee: Kind allowed on templates repo is 'inapp', 'shared', and 'custom'. Do we not planning on supporting this - should this be removed from all?

libs/logic-apps-shared/src/designer-client-services/lib/base/template.ts
Elaina-Lee: Original idea was - those two return a different type (TemplateManifest, WorkflowManifest). I can do the type of OR, but that means we would have to we have to infer the type `as TemplateManifest` everytime. Also, for workflowManifest, we can clarify templateId and workflowId.

In that case, I can change it back.

libs/designer/src/lib/ui/templates/templateoverview.tsx
Elaina-Lee: No this was not being read from workflow.json. If you look at the below code, it was doing `const trigger = getPropertyValue(details, 'trigger');` which is exactly the point I brought up in the chat. WorkflowManifest doesn't have "Trigger" in our contract - how should we approach this?

libs/logic-apps-shared/src/designer-client-services/lib/template.ts
```diff
@@ -6,9 +6,9 @@ export interface ITemplateService {
   openBladeAfterCreate: (workflowName: string | undefined) => void;
   onAddBlankWorkflow: () => Promise<void>;
   getAllTemplateNames: () => Promise<string[]>;
-  getResourceManifest: (resourcePath: string) => Promise<Template.Manifest>;
-  getWorkflowDefinition: (resourcePath: string) => Promise<LogicAppsV2.WorkflowDefinition>;
-  getContentPathUrl: (templateName: string, resourcePath: string) => string;
+  getResourceManifest: (templateId: string) => Promise<Template.TemplateManifest | Template.WorkflowManifest>;
```
preetriti1, line 9: lets call it resourcePath

libs/designer/src/lib/ui/templates/templateoverview.tsx
preetriti1: I was expecting us to save this data in the custom template [ this will be inferred data from workflow.json ], so github templates would still have this data in the details of templates... and when we publish the template in production we can always infer and put it there.

libs/designer/src/lib/ui/templates/__tests__/templateCard.spec.tsx
```diff
@@ -11,63 +11,68 @@ import React from 'react';
...
+        By: '',
+        Type: '',
+        Category: '',
+      },
       artifacts: [
         {
-          type: 'workflow',
-          file: 'workflow.json',
+          type: 'description',
+          file: 'description.md',
```
preetriti1, line 35: We never have this... this portion would include the maps/schemas right? are you trying to test anything in that aspect? if not then just leave it empty

libs/designer/src/lib/ui/templates/__tests__/displayConnections.spec.tsx
```diff
@@ -13,31 +13,66 @@ import React from 'react';
...
+      },
+      details: {
+        By: '',
+        Type: '',
+        Category: '',
+      },
+      artifacts: [
+        {
+          type: 'description',
+          file: 'description.md',
```
preetriti1, line 40: Why are we not keeping this empty in test files... is something breaking? also artifacts was not a required property right so if not present we should not fail in our code

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -33,7 +34,7 @@ export const SummaryPanel = ({ workflowId }: { workflowId: string }) => {
     }),
   };
 
-  return isNullOrUndefined(manifest) ? null : (
+  return isNullOrUndefined(workflowManifest) || isNullOrUndefined(templateManifest) ? null : (
```
preetriti1, line 37: This is just a fail safe check right?

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -88,22 +89,22 @@ export const SummaryPanel = ({ workflowId }: { workflowId: string }) => {
...
-              <Text className="msla-template-overview-section-detailkey">{detailsTags[key]}:</Text>
-              <Text>{manifest.details[key]}</Text>
+              <Text className="msla-template-overview-section-detailkey">{detailsTags[key as Template.DetailsType]}:</Text>
+              <Text>{templateManifest.details[key as Template.DetailsType]}</Text>
             </div>
           );
         })}
       </div>
-      {manifest.tags?.length ? (
+      {templateManifest.tags?.length ? (
```
preetriti1, line 107: SO both for accelerators and single workflow templates we will always be showing the template tags here not the workflow specific tags... wont we have tags in both the locations?

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/quickViewPanel.tsx
```diff
@@ -72,12 +75,12 @@ export const QuickViewPanel = ({
     () => (
       <QuickViewPanelHeader
         title={manifest.title}
-        description={manifest.description}
+        summary={manifest.summary}
         sourceCodeUrl={manifest.sourceCodeUrl}
-        details={manifest?.details ?? {}}
+        details={templateManifest?.details ?? {}}
       />
```
preetriti1, line 81: Wont we need to read the trigger from workflow manifest or is it not needed in header?

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -37,10 +37,13 @@ import {
...
-  manifest: Template.Manifest;
+  manifest: Template.WorkflowManifest;
   workflowName: string | undefined;
   kind: string | undefined;
-  images?: Record<string, string>;
+  images?: {
+    light?: string;
+    dark?: string;
+  };
   connectionKeys: string[];
```
preetriti1, line 47: Didn't we have the parameterKeys also here? 
If not how were we finding the associated workflows for each parameter, was that directly looking into manifest.parmeters?

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -244,16 +248,12 @@ const loadTemplateFromResourcePath = async (
   };
 
   if (isMultiWorkflow && workflows) {
```
preetriti1, line 250: Since workflows is a required property now, lets remove the check

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
```diff
@@ -281,11 +281,19 @@ const loadTemplateFromResourcePath = async (
     }
   } else {
```
preetriti1, line 282: Can we remove this else section now? workflows is required and we dont have to explicitly check for multiworkflows since now the workflowId is also marked with default.

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -53,23 +53,30 @@ export const getUniqueConnectors = (
...
 ): Template.Connection[] => {
-  const allConnections = Object.values(connections);
+  const allConnections = Object.values(connections).map((connection) => ({ ...connection, id: connection.connectorId }));
   return getUniqueConnectorsFromConnections(allConnections, subscriptionId, location);
 };
 
+interface UniqueConnectorInfo {
+  connectorId: string;
+  operationId: string | undefined;
+}
```
preetriti1, line 63: Dont we already have an interface called OperationInfo from for this..
and why are we including operationId in uniqueconnectorInfo

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -53,23 +53,30 @@ export const getUniqueConnectors = (
   subscriptionId: string,
   location: string
 ): Template.Connection[] => {
-  const allConnections = Object.values(connections);
+  const allConnections = Object.values(connections).map((connection) => ({ ...connection, id: connection.connectorId }));
```
preetriti1, line 56: I got really confused here - please change the variable name to featuredConnectors

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
Elaina-Lee: Yes, otherwise we'd need to do null check for reading fields for the rendering for each section

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
Elaina-Lee: Are we going to check `isMultiWorkflow` for every loop for `workflowData.workflow.workflowName = workflows[workflowId].name;` then?

I thought it was a bit unpractical to have the reduce function and spread for single workflow case... but we could do that as well

libs/designer/src/lib/core/templates/utils/helper.ts
Elaina-Lee: Because OperationInfo requires that operationId is a string not undefined

libs/designer/src/lib/core/actions/bjsworkflow/templates.ts
Elaina-Lee: Associated workflows is present under parameter definition, so it can be read directly. We don't look into manifest.parameters.

libs/designer/src/lib/core/templates/utils/helper.ts
Elaina-Lee: We weren't using OperationInfo since there operationId is a required property. Removed this since we won't use operationId as per our yesterday discussion about connector icons

libs/designer/src/lib/core/templates/utils/helper.ts
Elaina-Lee: Changed to allConnectors; We use this for `connections.tsx` where we combine all connections on the summary tab (or overview page in case of accelerator) - which is not featured connectors but all connectors

libs/designer/src/lib/ui/templates/templateoverview.tsx
Elaina-Lee: Discussed offline

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/quickViewPanel.tsx
Elaina-Lee: Discussed offline; trigger not present in workflowManifest

libs/designer/src/lib/core/templates/utils/helper.ts
```diff
@@ -52,24 +52,26 @@ export const getUniqueConnectors = (
...
   location: string
-): Template.Connection[] => {
-  const allConnections = Object.values(connections);
-  return getUniqueConnectorsFromConnections(allConnections, subscriptionId, location);
+): Template.FeaturedConnector[] => {
+  const allConnectors = Object.values(connections).map((connector) => ({ ...connector, id: connector.connectorId }));
+  return getUniqueConnectorsFromConnections(allConnectors, subscriptionId, location);
 };
 
 export const getUniqueConnectorsFromConnections = (
```
preetriti1, line 60: Can you double check if this is not getting called to list the connections in teh connections grid and also showing the connections summary in templateoverview and workflow summary tab?

libs/logic-apps-shared/src/utils/src/lib/models/template.ts
```diff
@@ -1,29 +1,48 @@
 export type SkuType = 'standard' | 'consumption';
 export type WorkflowKindType = 'stateful' | 'stateless';
-export type ConnectorRuntimeType = 'inapp' | 'shared';
+export type ConnectorRuntimeType = 'inapp' | 'shared' | 'builtin';
+export type FeaturedConnectorType = ConnectorRuntimeType | 'builtin';
```
preetriti1, line 4: These both are always same right? Then why do we need a separate type?

Elaina-Lee: Removed 'builtin' from ConnectorRuntimeType

preetriti1 (approved):

PR #6731, opened 2025-03-05
feat(templates): Adding resource selection feature in basics tab

preetriti1:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Change
Added resource selection feature in the basics tab.
1. If subscription/resourceGroup/location is not filled in we will  disable navigation to other tabs since those are required parameters. Initially there will be seed value but if user changes it... we cannot make required api calls, hence blocking the navigation.
2. For standard apps, we also require input to users workflow app.

## Test Plan
This is pending in future commits. Will be adding component and e2e level tests.

When workflow app is not filled in (disabled view)
image

Enabled view
image


rllyy97 (approved):
Looks good, but we should add error surfacing for the resources too if we don't already have plans for that (later pr is fine)

preetriti1:
> Looks good, but we should add error surfacing for the resources too if we don't already have plans for that (later pr is fine)

What type of error are we looking here? apart from required validation what else we will be needing

PR #6737, opened 2025-03-06
feat(template): Support overriding managed api connections

tonytang-microsoft-com:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Current Behavior
When going to connections tab, an existing connection, if exists, will automatically be used.

## New Behavior

Add option to allow overriding managed api connections, so the specified connection provided by the caller will automatically be selected.

## Impact of Change

## Test Plan

Manually tested locally

### DIFF
--- a/libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
+++ b/libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
@@ -65,11 +65,13 @@ export const WorkflowConnections = ({ connections }: WorkflowConnectionsProps) =
     location,
     connections: { references, mapping },
     workflows,
+    viewTemplateDetails,
   } = useSelector((state: RootState) => ({
     subscriptionId: state.workflow.subscriptionId,
     location: state.workflow.location,
     connections: state.workflow.connections,
     workflows: state.template.workflows,
+    viewTemplateDetails: state.templateOptions.viewTemplateDetails,
   }));
   const columnsNames = {
     name: intl.formatMessage({ defaultMessage: 'Name', id: 'tRe2Ct', description: 'Column name for connector name' }),
@@ -84,6 +86,7 @@ export const WorkflowConnections = ({ connections }: WorkflowConnectionsProps) =
       Object.values(workflows).map((workflow) =>
         workflow.connectionKeys.map((key) => {
           const connectionItem = connections[key];
+
           return {
             id: guid(),
             workflowId: workflow.id,
@@ -183,10 +186,13 @@ export const WorkflowConnections = ({ connections }: WorkflowConnectionsProps) =
   );
 
   const onConnectionsLoaded = async (loadedConnections: Connection[], item: ConnectionItem): Promise<void> => {
+    const connectionIdToOverride = viewTemplateDetails?.connectionsOverride?.[item.connectionKey]?.connectionId;
     const itemHasConnection = item.connection?.id && item.connection?.displayName === undefined;
     const connectionToUse = itemHasConnection
       ? loadedConnections.find((connection) => connection.id === item.connection?.id)
-      : loadedConnections[0];
+      : connectionIdToOverride
+        ? loadedConnections.find((connection) => connection.id === connectionIdToOverride)
+        : loadedConnections[0];
     const hasConnection = loadedConnections.length > 0;
     updateItemInConnectionsList({
       ...item,
--- a/libs/logic-apps-shared/src/utils/src/lib/models/template.ts
+++ b/libs/logic-apps-shared/src/utils/src/lib/models/template.ts
@@ -73,6 +73,10 @@ interface ContentInfo<T> {
   isEditable?: boolean;
 }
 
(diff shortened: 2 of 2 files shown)


libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
```diff
@@ -195,7 +200,7 @@ export const WorkflowConnections = ({ connections }: WorkflowConnectionsProps) =
       connection: connectionToUse ? { id: connectionToUse.id, displayName: getConnectionDisplayName(connectionToUse) } : undefined,
     });
 
-    if (!itemHasConnection && connectionToUse) {
+    if ((connectionToOverride || !itemHasConnection) && connectionToUse) {
```
preetriti1, line 203: Do you need connectionToOverride here? because this is going to update everytime you change the tabs

libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
```diff
@@ -183,10 +186,12 @@ export const WorkflowConnections = ({ connections }: WorkflowConnectionsProps) =
...
   const onConnectionsLoaded = async (loadedConnections: Connection[], item: ConnectionItem): Promise<void> => {
+    const connectionIdToOverride = viewTemplateDetails?.connectionsOverride?.[item.connectionKey]?.connectionId;
+    const connectionToOverride = loadedConnections.find((connection) => connection.id === connectionIdToOverride);
     const itemHasConnection = item.connection?.id && item.connection?.displayName === undefined;
-    const connectionToUse = itemHasConnection
-      ? loadedConnections.find((connection) => connection.id === item.connection?.id)
-      : loadedConnections[0];
+    const connectionToUse =
+      connectionToOverride ??
+      (itemHasConnection ? loadedConnections.find((connection) => connection.id === item.connection?.id) : loadedConnections[0]);
```
preetriti1, line 194: YOu should be using the connectionToOverride in the else portion

itemhasConnection ? honor the selected connection : [overridenConnection ?? connections[0]]

tonytang-microsoft-com: I think connectionsOverride should overrides the existing connections, that's the purpose of introducing this property, right?

libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
preetriti1: No... onConnectionsLoaded will be called whenever component is getting mounted that can be even when you are changing tabs

libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
preetriti1: you are just using connectionsOverride to select a connection from the list, so when first time when you load it will auto select from connectionsOverride, but now when teh user intentionally changes from teh dropdown - switch tab and comeback you wont be honoring that selection

libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
tonytang-microsoft-com: I see. in that case, we need to have a flag to make sure we only override it for the first time. Is there a place where you would suggest to save that flag?

libs/designer/src/lib/core/state/templates/templateOptionsSlice.ts
```diff
@@ -23,6 +24,12 @@ export const templateOptionsSlice = createSlice({
     setEnableResourceSelection: (state, action: PayloadAction<boolean>) => {
       state.enableResourceSelection = action.payload;
     },
+    setConnectionsOverrideCompleted: (state, action: PayloadAction<string>) => {
+      if (!state.connectionsOverrideCompleted) {
```
preetriti1, line 28: Why is this called completed?

libs/designer/src/lib/ui/templates/connections/workflowconnections.tsx
preetriti1: Can you please what we do for basics and parameters override? We only save it onec in the store... here the component would always read the connection fore this item from store if it is not present it would look at override ... that is why i unresolved the previous comment because you need to change the order of that and remove connectionToOverride from here.

tonytang-microsoft-com: Could you please elaborate what do you mean by save it only once? Currently we are saving it only once (that is when both manifest and connections are loaded, we calculate the override and save it). I've pinged you on IM and if needed let's sync up the details.

tonytang-microsoft-com: resolved after syncing up offline.

hartra344 (approved):

preetriti1 (approved):

ccastrotrejo (approved):

PR #6753, opened 2025-03-10
fix(vscode): Updating connection keys will no longer overwrite existing connection data

### LINKED ISSUE #6717: Issue since last week with managed api connections in vscode
### VSCode Dump

In followup of issue https://github.com/Azure/logicapps/discussions/664 we currently face an issue with the log analytics managed api connection (and likely other managed api connections) in VS code.

Each time we save the workflows, the connections.json, local.settings.json and parameters.json gets edited in a way which is not desirable, because we have parameterized it ourselves.

Result is that the connections and parameters.json is deployed to Azure with data which is not present and a bearer token which is valid for 7 days (

1. Please note how the connections.json looks before save:

Image

2. Please note how the parameters.json looks before save:

Image

3. Please note how the local.settings.json looks before save:

Image

4. As can be seen, we have parameterized the api connection both for vscode (and during deployment) for setting up the api connection. 

5. Edit a workflow.json and save it

6. You will see that the 3 files are aedjusted by VS code

7. The connections.json managed api section is overridden as follows:

Image

8. The parameters.json is also adjusted to contain the runtime url and an authentication section for a temp vscode bearer token which is valid for 7 days.

Image

9. The local.settings.json contains the bottom 2 entries after save, which is the runtime url and the token itsel

Image

Result is, that this gets deployed and because the connection.json references incorrect appsettings (like WORKFLOWS_SUBSCRIPTION_ID, WORKFLOWS_LOCATIONS_NAME and the raw bearer token. And therefore the logic app workflows do not work.

What I would expect: A save should not trigger these changes to files that are being deployed.

JoaquimMalcampo:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
If a connection does not have a connection key or if their connection key is expired, it will call an api to fetch the connection key to be saved in local.settings.json but also other data related to the connection reference which will be saved in connections.json. 

## New Behavior
Fixes #6717 
Will no longer overwrite existing connection data when calling api for connection keys. This will preserve any custom parameterization user has. 

## Test Plan
Ran unit tests


apps/vs-code-designer/src/app/utils/codeless/connection.ts
ccastrotrejo, line 1: Can you add a unit test in which we expect/verify that indeed these logic is not overwriting the existing connection data?

ccastrotrejo, line 1: NIT: Do we need to update the `getConnectionReference` function as we now don't expect to return anything?

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/utils/codeless/connection.ts
JoaquimMalcampo: We don't need to update getConnectionReference because it is still used when adding a new connection to the logic app. see: https://github.com/Azure/LogicAppsUX/blob/67fae0e88d3014ab154d3ebc0b8124c15c1b532e/apps/vs-code-designer/src/app/utils/codeless/connection.ts#L251

Only the cases when the connection exists and needs a connectionKey it needs the side effect of the API call and not the return value.

PR #6764, opened 2025-03-11
feat(designer): Introduce chunk text with metadata action

abhargavams:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Impact of Change

### DIFF
--- a/libs/logic-apps-shared/src/designer-client-services/lib/base/operationmanifest.ts
+++ b/libs/logic-apps-shared/src/designer-client-services/lib/base/operationmanifest.ts
@@ -125,6 +125,7 @@ const edifactbatchencode = 'edifactbatchencode';
 const edifactdecode = 'edifactdecode';
 const parsedocument = 'parsedocument';
 const parsedocumentwithmetadata = 'parsedocumentwithmetadata';
+const chunktextwithmetadata = 'chunktextwithmetadata';
 
 export const apiManagementConnectorId = '/connectionProviders/apiManagementOperation';
 export const azureFunctionConnectorId = '/connectionProviders/azureFunctionOperation';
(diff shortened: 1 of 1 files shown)


libs/logic-apps-shared/src/designer-client-services/lib/base/operationmanifest.ts
```diff
@@ -125,6 +125,7 @@ const edifactbatchencode = 'edifactbatchencode';
 const edifactdecode = 'edifactdecode';
 const parsedocument = 'parsedocument';
 const parsedocumentwithmetadata = 'parsedocumentwithmetadata';
+const chunktextwithmetadata = 'chunktextwithmetadata';
```
preetriti1, line 128: Is this present both in standard and consumption? If you make changes here it is assumed it is prsent in both places.
and i have not seen any entry for this in consumption data

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/base/operationmanifest.ts
abhargavams: Can you please tell where to do the changes if we have to go only for standard.

preetriti1: You have to create similar object in standard/operationmanifest.ts service implementation like how we have for consumption inside consumption/operationmanifest.ts for supportedTypes and manifestObjects.

Currently consumption is a super set of all types supported in standard, hence we didnt have any new object. But please follow that to create else this will also show up in consumption and might fail workflow runtime.

bjbennet: I tested the changes in the local desinger and Im only seeing the "chunk with metadata" action in standard workflows and not in consumption workflows

Standard
 image

Consumption
image

PR #6766, opened 2025-03-11
feat(designer): Introduce agent connections

ccastrotrejo:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

> [!NOTE]
> I added skip validation for saving workflow when enable agentic loops is checked
> URL and key-based authentication is the only auth method for now

## Current Behavior

Agent action didn't required a connection creation 

## New Behavior

Agent now requires a connection creation and also saves the connection in the new `AgentConnections` in the connections.json

## Impact of Change

Not a breaking change

## Test Plan

 

**Test will be added later**

#### Screen recording for agent connection

https://github.com/user-attachments/assets/d63a80d1-f5bd-4305-8e00-ab6da5a9ea10

#### Agent connections after saved workflow
CleanShot 2025-03-12 at 10 50 37@2x


ccastrotrejo (commented):
Self reviewed

PR #6779, opened 2025-03-12
fix(vscode): Fix parameterize connection checks to check all connections

JoaquimMalcampo:
## Type of Change

* [ x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior
Parameterize connections command relies on a check which checks connections if they are parameterized. Currently, that check only checks connections until they see one connection that is parameterized but then skips the rest. This change will make sure that check will return true if all connections are parameterized.

### DIFF
--- a/apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
@@ -72,7 +72,7 @@ export function parameterizeConnection(
  */
 export function hasParameterizedConnection(connectionsData: ConnectionsData): boolean {
   if (!connectionsData || Object.keys(connectionsData).length === 0) {
-    return false;
+    return true;
   }
   for (const connectionType in connectionsData) {
     if (connectionType !== 'serviceProviderConnections') {
@@ -105,7 +105,7 @@ export function hasParameterizedConnection(connectionsData: ConnectionsData): bo
  */
 export function areAllConnectionsParameterized(connectionsData: ConnectionsData): boolean {
   if (!connectionsData || Object.keys(connectionsData).length === 0) {
-    return false;
+    return true;
   }
   for (const connectionType in connectionsData) {
     if (connectionType !== 'serviceProviderConnections') {
--- a/apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
@@ -66,54 +66,6 @@ export function parameterizeConnection(
 }
 
 /**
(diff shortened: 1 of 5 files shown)


apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
```diff
@@ -66,11 +66,11 @@ export function parameterizeConnection(
 }
 
 /**
- * Checks if the connections data is parameterized.
+ * Checks if there exists a parameterized connection data entry.
  * @param {ConnectionsData} connectionsData - The connections data object.
  * @returns A boolean indicating whether the connections data is parameterized or not.
  */
-export function isConnectionsParameterized(connectionsData: ConnectionsData): boolean {
+export function hasParameterizedConnection(connectionsData: ConnectionsData): boolean {
```
ccastrotrejo, line 73: NIT: is this new `hasParameterizedConnection` used somewhere?

apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
```diff
@@ -98,6 +98,39 @@ export function isConnectionsParameterized(connectionsData: ConnectionsData): bo
   return false;
 }
 
+/**
+ * Checks if all connections are parameterized.
+ * @param {ConnectionsData} connectionsData - The connections data object.
+ * @returns A boolean indicating whether all connections are parameterized or not.
+ */
+export function areAllConnectionsParameterized(connectionsData: ConnectionsData): boolean {
```
ccastrotrejo, line 106: Can you add unit tests for this?

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
JoaquimMalcampo: No, I was thinking that there might be logic in the future which might want to check if some connections are parameterized especially when the user would have custom parameterization and has the parameterizeConnectionsDuringImport setting to false.

apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
JoaquimMalcampo: added

apps/vs-code-designer/src/app/utils/codeless/parameterizer.ts
ccastrotrejo: If we are not using this, lets remove it for now.

ccastrotrejo (approved):

PR #6785, opened 2025-03-13
feat(Templates): New `Review` components for templates

rllyy97:
## Main Changes

- Moved `Review` components to their own files
  - Old `ReviewCreatePanel`
  - New `ReviewAddPanel` primarily for ai-foundry
  - New `ReviewExisiting` primarily for ai-foundry
- Modified dev toolbox component
- Moved some components to fluent v9 and cleaned up some css, fixed some dark mode issues
 
> [!IMPORTANT]
> **RISK**: 
> Low risk:  Refactoring and new components
> 
> **TESTS**: 
> Added unit test

## Screenshots

image
image

### New existing workflow component
image

## Mocks

image


libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/reviewCreateTab.tsx
```diff
@@ -75,8 +88,127 @@ export const ReviewCreatePanel = () => {
...
+  const subscriptionDisplayName = useMemo(
+    () => subscriptions?.find((sub) => sub.id === `/subscriptions/${subscriptionId}`)?.displayName ?? '-',
+    [subscriptions, subscriptionId]
+  );
+
+  const styles = useStyles();
+
+  // Compact view
+  if (isCompactView) {
+    return (
```
preetriti1, line 154: Can we use a different component here so we do not add strings for action and workflow mixed in here.
It would be great if this component can take in resourceOverrides, so if we need to use this here in azure portal we can call this workflowName or template name rather action name or action.

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/reviewCreateTab.tsx
```diff
@@ -75,8 +88,127 @@ export const ReviewCreatePanel = () => {
...
+    LOCATION: intl.formatMessage({
+      defaultMessage: 'Location',
+      id: 'OO37qq',
+      description: 'Label for location',
+    }),
+    RESOURCE_GROUP: intl.formatMessage({
+      defaultMessage: 'Resource group',
+      id: 'YAGa7I',
+      description: 'Label for resource group',
+    }),
```
preetriti1, line 116: We already have these resources used in resourcePicker, so lets reuse those resource strings.

preetriti1 (changes requested):

libs/designer/src/lib/ui/panel/templatePanel/createWorkflowPanel/tabs/reviewCreateTab.tsx
rllyy97: What do you mean by reuse the resource strings?  In resource picker they're inside a useMemo that uses an instance of the intl hook, that's not really something we can export afaik.  And it was my understanding that we can't pull out the objects from the formatMessage calls since intl won't pick those up when generating it's translation files.

rllyy97: We could probably try making a hook to just hold a big object of all of our strings if we want to go that route.

libs/designer/src/lib/ui/templates/review/ReviewCreatePanel.tsx
```diff
@@ -0,0 +1,151 @@
...
+                    </Text>
+                  </div>
+                </div>
+              );
+            })}
+          </div>
+        </>
+      )}
+
+      {isConsumption && !Object.keys(connections).length && !Object.keys(parameterDefinitions).length ? (
```
preetriti1, line 107: Do you mind adding the section to for subscription/resourcegroup/location here as well in this format when enableresourceselection is true in the options store please?

libs/designer/src/lib/core/index.ts
```diff
@@ -87,3 +87,6 @@ export { useCanUndo, useCanRedo } from './state/undoRedo/undoRedoSelectors';
 export { resetDesignerView } from './state/designerView/designerViewSlice';
 export * from './queries/runs';
 export { reloadTemplates } from './actions/bjsworkflow/templates';
```
preetriti1, line 89: COuld you export the type of rootState of template store

libs/designer/src/lib/ui/templates/review/ReviewCreatePanel.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    </Text>
+                  </div>
+                </div>
+              );
+            })}
+          </div>
+        </>
+      )}
+
+      {enableResourceSelection && (
```
preetriti1, line 121: If this is hte exact same section, can we put this in a file so it can be exported from the three files

preetriti1 (approved):
If possible move the common resource section to a component if it can be reused, else I will move in my PR.

PR #6828, opened 2025-03-18
feat(Templates): Review tab improvements + Template provider `useExternalRedux` flag

rllyy97:
## Main Changes

Templates review tabs no longer show auth + parameters if there are no values to display.
Templates provider has new `useExternalRedux` flag to prevent redux provider creation.


libs/designer/src/lib/ui/templates/review/ReviewExisting.tsx
```diff
@@ -28,7 +29,7 @@ export const ReviewExisting = ({ resourceOverrides }: ReviewExistingProps) => {
       <div className="msla-templates-review-block">
         <Text>{resourceOverrides?.workflowName ?? resourceStrings.WORKFLOW_NAME}</Text>
         <Text weight="semibold" className={styles.actionName}>
-          {existingWorkflowName}
+          {existingWorkflowName ?? templateName}
```
preetriti1, line 32: Can you revert this change,? this is the workflow name and lets pass in the value from caller, my current pr is doing that.. we are not storing those data here because it might cause other cascading changes which is required for a template

preetriti1 (approved):

PR #6831, opened 2025-03-19
feat(templates): Add action for fetching template connections from selected workflows

preetriti1:
## Change
This PR contains helper methods and action methods to fetch used connections from selected workflows.

## Test Plan
Just tested manually through a temporary test page component, the UI is still not ready so havent written e2e tests yet.
Might add CITs when we have enough e2e functionalities ready.


hartra344 (approved):

libs/designer/src/lib/ui/templates/basics/resourcepicker.tsx
```diff
@@ -138,9 +139,9 @@ export const ResourcePicker = ({ viewMode = 'default', onSelectApp }: ResourcePi
           resources={(allLogicApps ?? []).map((app) => ({
             id: app.id,
             name: app.name,
-            displayName: app.name,
+            displayName: equals(app.plan, 'consumption') ? `${app.name} (Consumption)` : `${app.name} (Standard)`,
```
rllyy97, line 142: Do we need to localize these strings?

rllyy97 (approved):

libs/designer/src/lib/ui/templates/basics/resourcepicker.tsx
preetriti1: I thought these are non localized strings because these are skus... I can update in next PR if that is not the case

PR #6843, opened 2025-03-20
fix(Templates): Made `useWorkflowTemplate` hook + usage null safe

rllyy97:
## Main Changes

Made the `useWorkflowTemplate` hook + usage null-safe.
Was throwing errors over in AI-Foundry due to a race condition with an async-thunk call.

### DIFF
--- a/libs/designer/src/lib/ui/templates/basics/resource.tsx
+++ b/libs/designer/src/lib/ui/templates/basics/resource.tsx
@@ -8,6 +8,7 @@ import { validateWorkflowName } from '../../../core/actions/bjsworkflow/template
 import { useIntl } from 'react-intl';
 import { useTemplatesStrings } from '../templatesStrings';
 import { TemplateDisplay } from '../review/ReviewAddPanel';
+import { useMemo } from 'react';
 
 interface ResourceSectionProps {
   workflowId: string;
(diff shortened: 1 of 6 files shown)


libs/designer/src/lib/ui/templates/basics/resource.tsx
```diff
@@ -59,10 +60,9 @@ export const ResourceSection = (props: ResourceSectionProps) => {
 const WorkflowName = ({ workflowId, label, placeholder }: { workflowId: string; label: string; placeholder?: string }) => {
   const dispatch = useDispatch<AppDispatch>();
   const styles = useStyles();
-  const {
-    workflowName,
-    errors: { workflow: workflowError },
-  } = useWorkflowTemplate(workflowId);
+  const workflowTemplate = useWorkflowTemplate(workflowId);
+  const workflowName = useMemo(() => workflowTemplate?.workflowName, [workflowTemplate]);
+  const workflowError = useMemo(() => workflowTemplate?.errors?.workflow, [workflowTemplate]);
```
preetriti1, line 65: I am assuming it is only this component because all other components render only when template data is present. Else it would be a bug and we will want to catch that.

rllyy97: Yes this was the component running into the error over in AI Foundry

Elaina-Lee (approved):

PR #6844, opened 2025-03-20
feat(Templates): Infer workflow trigger from definition for accelerator

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

Infer workflow type from workflow.json for accelerator since it is not present on workflow manifest. Removing Solution type on accelerator workflow panel.

## Impact of Change

image

### DIFF
--- a/libs/designer/src/lib/ui/templates/templatesStrings.ts
+++ b/libs/designer/src/lib/ui/templates/templatesStrings.ts
@@ -38,3 +38,30 @@ export const useTemplatesStrings = () => {
     },
   };
 };
+
+export const useConnectorStatusStrings = () => {
+  const intl = useIntl();
+
+  return {
+    connected: intl.formatMessage({
+      defaultMessage: 'Connected',
+      id: 'oOGTSo',
+      description: 'Connected text',
+    }),
+    notConnected: intl.formatMessage({
+      defaultMessage: 'Not connected',
+      id: 'WxJJcQ',
+      description: 'Not Connected text',
+    }),
+    authenticated: intl.formatMessage({
+      defaultMessage: 'Authenticated',
+      id: 'iwKxSD',
+      description: 'Connection authenticated text',
+    }),
+    notAuthenticated: intl.formatMessage({
+      defaultMessage: 'Not authenticated',
+      id: 'QdRn5z',
+      description: 'Connection not authenticated text',
+    }),
+  };
+};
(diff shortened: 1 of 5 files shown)


rllyy97 (approved):

libs/designer/src/lib/ui/templates/templatesStrings.ts
```diff
@@ -38,3 +39,55 @@ export const useTemplatesStrings = () => {
...
+      defaultMessage: 'Event',
+      id: '2iq12E',
+      description: 'Trigger belongs to Event category',
+    }),
+  };
+
+  return Object.values(triggers)
+    .map((trigger) => detailTriggerStrings[trigger.type as keyof typeof detailTriggerStrings] ?? detailTriggerStrings.Event)
+    .join(', ');
+};
```
preetriti1, line 93: Apart from resource strings i already created a method to getTriggerType in my last PR, can you use that and then check with Divya on what we need to call all types.

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
```diff
@@ -99,14 +100,22 @@ export const SummaryPanel = ({ workflowId }: { workflowId: string }) => {
...
+        {!isMultiWorkflow && (
+          <div className="msla-template-overview-section-detail">
+            <Text className="msla-template-overview-section-detailkey">{detailsTags.Type}:</Text>
+            <Text>{templateManifest.details.Type}</Text>
+          </div>
+        )}
+
+        <div className="msla-template-overview-section-detail">
+          <Text className="msla-template-overview-section-detailkey">{detailsTags.Trigger}:</Text>
+          <Text>{isMultiWorkflow ? getTriggerFromDefinition(workflowDefinition?.triggers ?? {}) : templateManifest.details.Trigger}</Text>
```
preetriti1, line 112: SO we are stamping this while saving the template manifest right for single workflow?

preetriti1 (approved):

libs/designer/src/lib/ui/panel/templatePanel/quickViewPanel/tabs/summaryTab.tsx
Elaina-Lee: It's part of the manifest so no need to have explicit conditions for stamping for single workflow; yes I checked it renders correct data from templateManifest

PR #6848, opened 2025-03-21
feat(Templates): Configure Template Tabs with Initial Flow

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior
This introduces:
- tabs structure for configure template wizard experience

This changes:
- some of the template panel UI components naming from panel to generic since they are being reused in non-panel places

## Test Plan
UI Elements are all placeholders of values represented. This is a PR for general structuring of configure template wizard. Tests will be added as individual tabs / panels get worked on.

https://github.com/user-attachments/assets/1cf13eb5-cc1d-4d4c-8020-ec1961545174


libs/designer/src/lib/common/constants.ts
```diff
@@ -482,13 +482,20 @@ export default {
     SWAGGER_ENDPOINT: 'SWAGGER_ENDPOINT',
     TESTING: 'TESTING',
   },
-  TEMPLATE_PANEL_TAB_NAMES: {
+  TEMPLATE_TAB_NAMES: {
     OVERVIEW: 'OVERVIEW',
```
preetriti1, line 486: Shouldn't we group them separately?

libs/designer/src/lib/core/state/templates/panelSlice.ts
```diff
@@ -33,6 +37,16 @@ export const panelSlice = createSlice({
       state.isOpen = true;
       state.currentPanelView = TemplatePanelView.CreateWorkflow;
     },
+    openConfigureWorkflowPanelView: (state) => {
```
preetriti1, line 40: Instead of having separate actions for each panel, can we just send the panel view in action payload so from the caller? or from a method to simplify calling from ux components? the stores here just have many small methods.

libs/designer/src/lib/core/state/templates/tabSlice.ts
```diff
@@ -0,0 +1,27 @@
+import type { PayloadAction } from '@reduxjs/toolkit';
+import { createSlice } from '@reduxjs/toolkit';
+import { resetTemplatesState } from '../global';
+
+export interface TabState {
+  selectedTabId: string | undefined;
+}
+
+const initialState: TabState = {
+  selectedTabId: undefined,
```
preetriti1, line 10: I am assuming we will store the status of each tab too here like incomplete, completed or in error.

libs/designer/src/lib/ui/panel/configureTemplatePanel/configureWorkflowsPanel/configureWorkflowsPanel.tsx
```diff
@@ -0,0 +1,79 @@
...
+    [selectedTabProps?.footerContent]
+  );
+
+  return (
+    <Panel
+      styles={{ main: { padding: '0 20px', zIndex: 1000 }, content: { paddingLeft: '0px' } }}
+      isLightDismiss={false}
+      type={PanelType.custom}
+      customWidth={'50%'}
+      isOpen={isOpen && currentPanelView === TemplatePanelView.ConfigureWorkflows}
```
preetriti1, line 69: Is it possible to return null before doing any initialization here if configureworkflows is not the panel view? 
that would help in reducing all re-renders or re-calculation that happen due to store changes

preetriti1 (changes requested):
Can you move all UI components like panel, tabs and even small components inside core/ui/configuretemplate
that would keep it well organized and help in creating separate export package for templates too.

libs/designer/src/lib/core/state/templates/tabSlice.ts
Elaina-Lee: I was actually thinking it would go here: `libs\designer\src\lib\ui\panel\configureTemplatePanel\configureWorkflowsPanel\usePanelTabs.tsx`

We are pushign usePanelTabs to handle all the tabs logic for other components, so I wanted to be consistent with that.

preetriti1 (approved):

PR #6849, opened 2025-03-21
fix(Templates): Remove built-in operations from filtering dropdown

Elaina-Lee:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

Removing the built-in operations from the featured connectors (connections) dropdown as we do not support it.

## Impact of Change


libs/designer/src/lib/ui/templates/filters/templatesearchfilters.tsx
```diff
@@ -207,7 +207,7 @@ const Filters = ({ detailFilters }: { detailFilters: TemplateDetailFilterType })
       templateManifest.skus.includes(isConsumption ? 'consumption' : 'standard')
     );
     const allConnectors = Object.values(skuTemplates).flatMap((template) => template.featuredConnectors ?? []);
-    const uniqueConnectorsFromConnections = getUniqueConnectorsFromConnections(allConnectors, subscriptionId, location);
+    const uniqueConnectorsFromConnections = getUniqueConnectorsFromConnections(allConnectors, subscriptionId, location, true);
```
preetriti1, line 210: Always put the argument name in comment for constant values like /* remove... */true

preetriti1 (approved):

PR #6875, opened 2025-03-27
feat(Designer): Node `description` changes + trigger dialog

rllyy97:
## Main Changes

- Node `comment` property renamed to `description` in all ux instances
- Node description input is always visible on triggers
  - Also removed the option to add/remove from context menu for triggers
- Added a new dialog for editing trigger description on save
  - This is to be used in the client, and passed an `onSubmit` prop
  - Can also dismiss and won't show again for that workflow (saved in local storage)
  - The dialog only shows up when the workflow is a request + response with any action in between
  - The trigger description defaults to the middle action's description
 
> [!IMPORTANT]
> **RISK**: 
> Low risk, adding new dialog and adjusting unused existing dialog
> 
> **TESTS**: 
> Updated related tests

---

image
image


hartra344 (approved):

preetriti1:
Do we give any warning on dismiss, saying to edit again you should do this this...

rllyy97:
> Do we give any warning on dismiss, saying to edit again you should do this this...

Nothing happens when you dismiss the dialog other than marking the workflow in local storage so that it doesn't appear again

PR #6893, opened 2025-03-28
feat(Templates): Converted `TemplateDisplay` description to markdown

rllyy97:
## Main Changes

Converted the `TemplateDisplay` description to markdown to match the data coming in from templates.

### Example

image

### DIFF
--- a/libs/designer/src/lib/ui/panel/templatePanel/panel.less
+++ b/libs/designer/src/lib/ui/panel/templatePanel/panel.less
@@ -291,4 +291,8 @@
 		grid-template-columns: 1fr 1fr;
 		gap: 6px 32px;
 	}
+
+	&.description {
+		margin: 4px 0;
+	}
 }
--- a/libs/designer/src/lib/ui/templates/templateDisplay.tsx
+++ b/libs/designer/src/lib/ui/templates/templateDisplay.tsx
@@ -28,7 +28,11 @@ export const TemplateDisplay = ({ titleLabel, showDescription, cssOverrides }: T
       <Text weight="semibold" className={styles.actionName}>
         {templateTitle}
       </Text>
-      {showDescription && templateDescription && <Markdown className={'description'}>{templateDescription}</Markdown>}
+      {showDescription && templateDescription && (
+        <Markdown className="msla-template-markdown description" linkTarget="_blank">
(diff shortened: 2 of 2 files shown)


Elaina-Lee:
I'm wondering, why did we introduce templateDisplay to the designer/src repo instead of designer-ui/src repo? We have `libs\designer-ui\src\lib\templates` that has similarly styled generic components.

Elaina-Lee (approved):

libs/designer/src/lib/ui/templates/templateDisplay.tsx
```diff
@@ -27,7 +28,7 @@ export const TemplateDisplay = ({ titleLabel, showDescription, cssOverrides }: T
       <Text weight="semibold" className={styles.actionName}>
         {templateTitle}
       </Text>
-      {showDescription && <Text>{templateDescription}</Text>}
+      {showDescription && templateDescription && <Markdown className={'description'}>{templateDescription}</Markdown>}
```
Elaina-Lee, line 31: For other template components, we put `linkTarget="_blank"` so that it will open in another page. Also, for template markdown components, we use `msla-template-markdown` (reference: libs\designer\src\lib\ui\panel\templatePanel\quickViewPanel\tabs\summaryTab.tsx) - this gets rid of weird margin gap and ensures the consistent link color (markdown default link color is slightly different from the color we use). Unless intended, can we follow the same format?

Elaina-Lee (changes requested):

libs/designer/src/lib/ui/templates/templateDisplay.tsx
rllyy97: Makes sense!  I'll update it

libs/designer/src/lib/ui/panel/templatePanel/panel.less
```diff
@@ -291,4 +291,8 @@
 		grid-template-columns: 1fr 1fr;
 		gap: 6px 32px;
 	}
+
+	&.description {
```
Elaina-Lee, line 295: Could we remove this since it is no longer being referenced?

Elaina-Lee (approved):

PR #6912, opened 2025-03-30
feat(templates): Adding toaster component in configure template to show template information

preetriti1:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Change
This includes a toaster component  which shows template information all the time in the wizard component.
It shows auto dismissable toaster when workflows are saved and when template is published, but the info component is rendered again when tabs are switched. Following the mocks.

## Test Plan
Manually tested this in standalone wizard component

templatetoaster


libs/designer/src/lib/ui/configuretemplate/toasters.tsx
```diff
@@ -0,0 +1,90 @@
...
+    if (show) {
+      updateToast({
+        content: (
+          <Toast>
+            <ToastTitle>{title}</ToastTitle>
+            <ToastBody style={{ paddingTop: 12, marginLeft: '-18px', fontSize: 'small' }}>{content}</ToastBody>
+          </Toast>
+        ),
+        ...toastDetails,
+        intent: 'success',
```
rllyy97, line 63: Do we want to pull this out to a prop or is always success fine?

rllyy97 (approved):

libs/designer/src/lib/ui/configuretemplate/toasters.tsx
preetriti1: We can pull it to a prop if needed in future. Right now this is always success

PR #6914, opened 2025-03-31
feat(designer): Agent Instruction Editor

Eric-B-Wu:
Creating a new editor to support Agent Instructions

AGENT6


takyyon:
Can we add some spacing between each section please? System and User sections. Also, maybe reduce the description size a bit since the title is smaller.

Eric-B-Wu:
> Can we add some spacing between each section please? System and User sections. Also, maybe reduce the description size a bit since the title is smaller.

I had purposely ket them together because to have users understand its a single editor, if that's not needed I can go ahead and clean up the spacing

takyyon:
> > Can we add some spacing between each section please? System and User sections. Also, maybe reduce the description size a bit since the title is smaller.
> 
> I had purposely ket them together because to have users understand its a single editor, if that's not needed I can go ahead and clean up the spacing

Well it is sort of a single editor from our side (on how we treat it on the BE) but for the user System and User messages should be two different sections

Eric-B-Wu:
> > > Can we add some spacing between each section please? System and User sections. Also, maybe reduce the description size a bit since the title is smaller.
> > 
> > 
> > I had purposely ket them together because to have users understand its a single editor, if that's not needed I can go ahead and clean up the spacing
> 
> Well it is sort of a single editor from our side (on how we treat it on the BE) but for the user System and User messages should be two different sections

I'm going to add the the margin in between the editors, but I'll keep the left padding on the editors so users can understand both are still under the 'Instruction for Agent' and the description text applies for both sub-editors

takyyon (approved):

PR #6929, opened 2025-03-31
feat(Designer): Exported 2 extra state methods for templates

rllyy97:
## Main Changes

Exported `validateWorkflowsBasicInfo` and `useExistingWorkflowNames` for use in templates flow outside designer.

### DIFF
--- a/libs/designer/src/lib/core/index.ts
+++ b/libs/designer/src/lib/core/index.ts
@@ -94,7 +94,13 @@ export { storeStateToUndoRedoHistory, onUndoClick, onRedoClick } from './actions
 export { useCanUndo, useCanRedo } from './state/undoRedo/undoRedoSelectors';
 export { resetDesignerView } from './state/designerView/designerViewSlice';
 export * from './queries/runs';
-export { reloadTemplates, resetStateOnResourceChange, type WorkflowTemplateData } from './actions/bjsworkflow/templates';
+export * from './queries/template';
+export {
+  reloadTemplates,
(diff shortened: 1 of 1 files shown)


ccastrotrejo (approved):

libs/designer/src/lib/core/index.ts
```diff
@@ -94,7 +94,13 @@ export { storeStateToUndoRedoHistory, onUndoClick, onRedoClick } from './actions
 export { useCanUndo, useCanRedo } from './state/undoRedo/undoRedoSelectors';
 export { resetDesignerView } from './state/designerView/designerViewSlice';
 export * from './queries/runs';
-export { reloadTemplates, resetStateOnResourceChange, type WorkflowTemplateData } from './actions/bjsworkflow/templates';
+export * from './queries/template';
+export {
+  reloadTemplates,
+  resetStateOnResourceChange,
```
preetriti1, line 100: where is useExistingWorkflowName? btw we do not need that one for consumption.

preetriti1 (approved):

libs/designer/src/lib/core/index.ts
rllyy97: It's in `./queries/template`

rllyy97: Ah ok I'm seeing the consumption logic now, well now we have it exported for if we use it with standard in the future.

hartra344 (approved):

PR #6972, opened 2025-04-07
feat(Templates): Custom Templates Parameters Tab

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior

- Working version of display and edit parameters with correct UI
- Update state, enable save on dirty state only, display associated workflows on accelearator case

## To be added

- Validation of parameters and displaying errors and blocking save on errors

## Test Plan
- UI elements are still to be polished (i.e. allowed values to be added, associated workflows section needs to be reviewed)
- Validation tests will be added along with validation logic & UI in the next PR

https://github.com/user-attachments/assets/248d3535-b749-4d78-bb9a-8b6927d91251


ccastrotrejo (approved):

libs/designer/src/lib/ui/configuretemplate/parameters/customizeParameter.tsx
```diff
@@ -0,0 +1,101 @@
...
+        label: resourceStrings.ParameterName,
+        value: parameterDefinition.name || '',
+        type: 'text',
+        required: true,
+      },
+      {
+        label: resourceStrings.Type,
+        value: parameterDefinition.type || '',
+        type: 'text',
+        required: true,
```
preetriti1, line 34: Do we need required field for text fields? There is no need to show the red asterix there no?

libs/designer/src/lib/ui/configuretemplate/parameters/customizeParameter.tsx
```diff
@@ -0,0 +1,101 @@
...
+            ...parameterDefinition,
+            default: value,
+          });
+        },
+      },
+      {
+        label: resourceStrings.Description,
+        value: parameterDefinition.description || '',
+        type: 'textarea',
+        required: true,
```
preetriti1, line 64: Why do we need description as required?

libs/designer/src/lib/ui/configuretemplate/parameters/parameterslist.tsx
```diff
@@ -0,0 +1,149 @@
...
+                <TableCellLayout>{item.name}</TableCellLayout>
+              </TableCell>
+              <TableCell>
+                <TableCellLayout>{item.type}</TableCellLayout>
+              </TableCell>
+              <TableCell>
+                <TableCellLayout>{item.default}</TableCellLayout>
+              </TableCell>
+              <TableCell>
+                <TableCellLayout>{item.allowedValues}</TableCellLayout>
```
preetriti1, line 125: Can we remove allowed values from the PR?

libs/designer/src/lib/ui/configuretemplate/resources.ts
```diff
@@ -149,6 +154,56 @@ export const useResourceStrings = () => {
       id: 'ekM77J',
       description: 'Label for workflow Name',
     }),
+    Placeholder: intl.formatMessage({
+      defaultMessage: '--',
+      id: '5lRHeK',
+      description: 'Accessibility label indicating that the value is not set',
```
preetriti1, line 160: I was thinking why are we localising --, can't we hardcore that?

Elaina-Lee: But this is a legitimate value that shows up when there are no values, shouldn't we keep it for accessbility to indicate that this is a placeholder value until the actual value is placed?

libs/designer/src/lib/ui/configuretemplate/parameters/customizeParameter.tsx
Elaina-Lee: I was following the figma design but removing them

PR #6978, opened 2025-04-08
feat(designer): Add parameters to agent action

ccastrotrejo:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Impact of Change

CleanShot 2025-04-08 at 14 46 13@2x


libs/logic-apps-shared/src/designer-client-services/lib/standard/manifest/agentloop.ts
```diff
@@ -54,6 +54,43 @@ export default {
           'x-ms-editor': 'agentinstruction',
           type: 'array',
         },
+        agentModelType: {
```
takyyon, line 57: Oh, all of these need to go under agentModelSettings: {}

takyyon (changes requested):

takyyon (approved):

PR #7024, opened 2025-04-14
fix(designer): Add waiting for input and update output channel to be optional

takyyon:
## Type of Change

* [ ] Bug fix
* [ ] Feature
* [ ] Other

## Impact of Change

image
image


rllyy97 (approved):

libs/designer/src/lib/ui/panel/agentChat/agentChat.tsx
```diff
@@ -189,13 +190,15 @@ export const AgentChat = ({
       dispatch(setFocusNode(agentLastOperation));
       dispatch(changePanelNode(agentLastOperation));
     },
-    [Object.keys(agentLastOperations).length, dispatch]
+    [agentLastOperations, dispatch]
```
ccastrotrejo, line 193: Don't change this, I still need to figure out a way to handle that the agentLastOperations don't update on every render. This makes the chat to render infinite times. I will work on this today.

ccastrotrejo: Thanks

ccastrotrejo (approved):

PR #7037, opened 2025-04-16
feat(designer): Update pinned node to alternated selected node

ccastrotrejo:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## Current Behavior

- Current behavior allows to have an extra "pinned node" in the side panel. This only allowed to have pinned nodes, this means that the extra node information will always be there.

## New Behavior

-  New behavior allows now to have another "alternate selected node", which means we can have an extra node in the panel without explicitly pinning it. New behavior supports old behavior as well.
- The new implementation allow us to have a "persistence" mode for the alternated selected node, which is either "pinned" or "selected"

## Impact of Change

Not a breaking change

## Test Plan

 

- Existed e2e tests didn't break and I will add some e2e tests after bug bash

https://github.com/user-attachments/assets/3737d4dd-211f-471f-9afd-cffda792fe5a


libs/designer/src/lib/core/state/panel/panelSlice.ts
```diff
@@ -269,13 +286,21 @@ export const panelSlice = createSlice({
...
+        state.operationContent.alternateSelectedNode?.nodeId &&
+        !actionIds.includes(state.operationContent.alternateSelectedNode.nodeId ?? '')
+      ) {
+        state.operationContent.alternateSelectedNode.nodeId = undefined;
       }
       if (state.operationContent.selectedNodeId && !actionIds.includes(state.operationContent.selectedNodeId ?? '')) {
         state.operationContent.selectedNodeId = undefined;
       }
-      if (state.operationContent.pinnedNodeId == null && state.operationContent.selectedNodeId == null) {
+      console.log(
```
rllyy97, line 298: Remove console

rllyy97 (approved):

PR #7061, opened 2025-04-21
feat(Templates): Custom Templates Validation & Error Handling

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## New Behavior
Added validation for templates ad the review + publish stage, which applies to all present data. Serialization to be sending to the back-end is not included in this PR. This validation is for front-end to ensure having all necessary data.

## Test Plan
Unit tests are added for validaiton functions

https://github.com/user-attachments/assets/b8ec441b-6934-450f-af3c-b49d065a5dae


libs/logic-apps-shared/src/utils/src/lib/models/template.ts
ccastrotrejo, line 1: NIT: maybe not needed?

libs/designer/src/lib/ui/configuretemplate/tabs/useWizardTabs.tsx
```diff
@@ -21,33 +21,46 @@ export const useConfigureTemplateWizardTabs = ({
...
+    ({ errors }) =>
+      !!errors?.workflow ||
+      !!errors?.kind ||
+      (errors?.manifest && Object.values(errors?.manifest ?? {}).some((value) => value !== undefined))
+  );
 
   return [
     workflowsTab(resources, dispatch, onSaveWorkflows, {
-      tabStatusIcon: 'in-progress',
+      tabStatusIcon: hasAnyWorkflowErrors ? 'error' : runValidation ? 'success' : 'in-progress',
```
rllyy97, line 44: Nit:
Since this icon priority logic is the same across all of the tabs (ie. check for error, check for validation, check for in-progress, default to undefined), instead of passing the icon as a string, we should just pass in flags for things like `isError`, `isSuccess`, etc, then we don't have to have these long conditional chains / might be better for readability

rllyy97: Can be a later change

Elaina-Lee: We originally had a flag `isError`. The reason why we changed it:
1. The flag is only used for tab status icon
2. Then we would need 3 different flags/props just for one icon: `isError`, `isSuccess`, `isInProgress` 
3. in `templatesContent.tsx`, we have the function for `TabStatusIcon` which now uses switch case to show icons but will need to be nested if statements instead
4. For above 3 - if we use switch statements for flags, we need to identify logic for prioritizing flags within the designer-ui component (for example, what if isSuccess and isError and isProgress are all true?); this logic may be different for rendering for custom templates and templates
5. For above 4 - if we decide to go this path, then to ensure we render the correct icon (not to worry about one flag getting prioritized over the other ones), we may have to end up with different logic for different flags, which I am not sure if is necessarily more readable
       isError: hasAnyWorkflowErrors
       isSuccess: !parametersHasError && runValidation && enableWizard,
       isInProgress: !enableWizard && !runValidation......

libs/designer/src/lib/ui/configuretemplate/workflows/customizeWorkflows.tsx
```diff
@@ -180,6 +196,47 @@ const CustomizeWorkflowSection = ({
...
+          updateWorkflowDataField(workflowId, {
+            ...workflow,
+            manifest: {
+              ...workflow.manifest,
+              images: {
+                ...workflow.manifest?.images,
+                dark: value,
+              },
+            } as Template.WorkflowManifest,
+          });
```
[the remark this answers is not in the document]
Elaina-Lee: Since this is partial data, there is a chance where there exists no value under workflow.manifest, resulting in type errors like below:

image

Another path of reflecting this comment would be like below, I would be unsure on whether this look much simpler:
image

libs/designer/src/lib/ui/configuretemplate/tabs/useWizardTabs.tsx
rllyy97: 3:  We're already doing the if statements, they're just here in each object instead of once inside the component
4+5:  I feel like the prioritization is already standardized no?  If there are errors, show the error icon, if it's complete show the success icon, otherwise show the in-progress icon.  We don't need to check everything for each flag, that's the point of passing the flag results themselves

rllyy97 (approved):

PR #7079, opened 2025-04-23
fix(Templates): Fixed issue with templates connections during creation

rllyy97:
## Main Changes

Fixed small issue with connection parameter object when creating LA from foundry template flow.


ccastrotrejo (approved):

preetriti1:
Why do we need to pass connectionId separately? Was workflow PUT failing with this or in runtime it was not working?

PR #7126, opened 2025-04-30
fix(Templates): Fixed dark mode styling bugs

rllyy97:
## Main Changes

- Adjusted templates flow to some Fluent v9 components for better dark mode styling
- Modified any color constants to use Fluent color variables

---
image
image
image
image


libs/designer/src/lib/ui/templates/cards/templateCard.tsx
```diff
@@ -106,7 +106,7 @@ export const TemplateCard = ({ templateName, isLightweight, blankWorkflowProps,
           </Text>
         </div>
       </div>
-
+      <Divider />
```
Elaina-Lee, line 109: Are we replacing border bottom with divider?

Elaina-Lee (approved):

libs/designer/src/lib/ui/templates/cards/templateCard.tsx
rllyy97: Yes, divider handles the theming automatically and looks identical

PR #7129, opened 2025-04-30
feat(Templates): Workflow Selection with User Input Resource Id & Duplicate Validation and handling

Elaina-Lee:
## Type of Change

* [ ] Bug fix
* [X] Feature
* [ ] Other

## New Behavior
- Consistent behavior of triggering function for customize workflows tab navigation by tab or next button
- Distinguishes workflowId and workflow.id: Until saved, workflowId is resourceId until the save is triggered - when save happens, it uses workflow.id (user-input) to swap out workflowId; Once saved, workflowId == workflow.id.
- Supports initial duplicate workflow.id for user-input or on workflow selection without overwriting data in the store with the new structure
- Gives user proper warning about the duplicate id (for user-input) and disables save until user fixes it
- When the workflowId is resourceId (before saved), enables workflow.id field to be textfield, allowing user-input
- When workflowId is not resourceId (after saved), disables workflow.id field, make it text (read-only), not allowing user-input
- always use workflow.id instead of workflowId to minimize possible errors
- workflowName will be dipslayed as a tab if any of the workflows do not match the original resource name
- duplicate ids will cause the id text field to error out as typing
- will gaurantee the unique name of the workflow on selection by adding -1, -2, etc.

## Impact of Change

Updated (with additional functionalities through discussion):

https://github.com/user-attachments/assets/85ea5450-1da9-410d-b1c1-ad86241c57b1


libs/designer/src/lib/ui/configuretemplate/workflows/selectWorkflows.tsx
```diff
@@ -50,10 +50,12 @@ export const SelectWorkflows = ({
         const normalizedWorkflowId = id.toLowerCase();
         onWorkflowsSelected([normalizedWorkflowId]);
       } else {
-        onWorkflowsSelected([]);
+        const workflowsOfSameSubscription =
+          Object.values(selectedWorkflowsList).map((workflow) => workflow.manifest?.metadata?.workflowSourceId as string) ?? [];
```
preetriti1, line 54: Not able to understand this part, why are we saving the same workflows list when logicapp is reselected... should you be checking if they belong to the same parent here before getting the current list

libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
```diff
@@ -622,16 +622,20 @@ export const getWorkflowsWithDefinitions = async (
...
+  const allWorkflowKeys = Object.keys(workflows);
   const allWorkflows = Object.values(workflows);
+
   const promises = allWorkflows.map((workflow) =>
     getWorkflowDefinitionForStandard(workflow.manifest?.metadata?.workflowSourceId as string)
   );
   const allWorkflowsData = (await Promise.all(promises)).reduce((result: Record<string, Partial<WorkflowTemplateData>>, data, index) => {
     const { kind, definition: workflowDefinition } = data;
     if (workflowDefinition) {
+      const workflowId = allWorkflowKeys[index];
```
preetriti1, line 634: does the id still contains the name of the actual resource and id field inside manifest is the newly changed one?

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
```diff
@@ -29,23 +29,39 @@ export const useConfigureWorkflowPanelTabs = ({
...
   }));
 
-  const hasError = false; // Placeholder for actual error state
-
   const [selectedWorkflowsList, setSelectedWorkflowsList] =
     useFunctionalState<Record<string, Partial<WorkflowTemplateData>>>(workflowsInTemplate);
 
+  const currentSelectedWorkflowsList = selectedWorkflowsList();
+  const duplicateIds = useMemo(() => {
+    const seen = new Set<string>();
```
preetriti1, line 37: WHy do we have duplicateids

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
```diff
@@ -29,23 +29,39 @@ export const useConfigureWorkflowPanelTabs = ({
...
+    }
+    return Array.from(duplicateIds);
+  }, [currentSelectedWorkflowsList]);
+
   const onWorkflowsSelected = (normalizedWorkflowIds: string[]) => {
     setSelectedWorkflowsList((prevSelectedWorkflows) => {
       const newSelectedWorkflows: Record<string, Partial<WorkflowTemplateData>> = {};
       for (const normalizedWorkflowId of normalizedWorkflowIds) {
-        const prevSelectedWorkflow = Object.values(prevSelectedWorkflows).find((workflow) =>
+        const [preSelectedWorkflowId, prevSelectedWorkflow] = Object.entries(prevSelectedWorkflows).find(([, workflow]) =>
```
preetriti1, line 55: small typo prevSelectedWorkflowId

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
```diff
@@ -29,23 +29,39 @@ export const useConfigureWorkflowPanelTabs = ({
...
           equals(workflow.manifest?.metadata?.workflowSourceId, normalizedWorkflowId)
-        );
-        const id = prevSelectedWorkflow?.id ?? getResourceNameFromId(normalizedWorkflowId);
-        newSelectedWorkflows[id] = prevSelectedWorkflow
+        ) ?? [undefined, undefined];
+
+        const workflowId = preSelectedWorkflowId ?? normalizedWorkflowId;
+
+        newSelectedWorkflows[workflowId] = prevSelectedWorkflow
```
preetriti1, line 61: The workflowId should not have the complete arm resource id no.... why do we need to store the complete id, since this name should be unique within the template

Elaina-Lee: Because otherwise it will not be unique. Imagine about the scenario someone chooses the name of the same id as a present workflow; then they select the workflow with the same name. We are using the initial id as the key up to the point of save. In this case, with the same key, the data will be merged. So therefore, for the unsaved workflows, we are using resource id to enforce the unique-ness, then on save, we are removing this resource id and replacing it with the user input id as the key. This logic is present in the save under usePanelTabs.tsx

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
```diff
@@ -64,30 +80,40 @@ export const useConfigureWorkflowPanelTabs = ({
...
   };
 
   const onSaveCompleted = useCallback(() => onSave?.(Object.keys(selectedWorkflowsList()).length > 1), [onSave, selectedWorkflowsList]);
 
   const onSaveChanges = () => {
+    // 1. Update the workflowId with user-input id (For newly selected workflow)
+    setSelectedWorkflowsList((prevSelectedWorkflows) => {
+      const newSelectedWorkflows: Record<string, Partial<WorkflowTemplateData>> = prevSelectedWorkflows;
+      for (const [workflowId, workflowData] of Object.entries(prevSelectedWorkflows)) {
+        if (workflowData.id && workflowId !== workflowData.id) {
```
preetriti1, line 108: We already have the resource id saved in workflowSourceId, so any such comparisons can be performed with that no, can we just check if the key !== id, where key can be just the resourceName from resourceId?

libs/designer/src/lib/ui/configuretemplate/workflows/customizeWorkflows.tsx
```diff
@@ -1,30 +1,57 @@
...
   selectedWorkflowsList: Record<string, Partial<WorkflowTemplateData>>;
   updateWorkflowDataField: (workflowId: string, workflowData: Partial<WorkflowTemplateData>) => void;
+  duplicateIds: string[];
 }) => {
+  const intl = useIntl();
   const workflowEntries = Object.entries(selectedWorkflowsList);
 
   return (
     <div className="msla-templates-tab msla-panel-no-description-tab">
+      {duplicateIds.length ? (
```
preetriti1, line 34: Instead of doing this - can we check if the name exists when they are typing it itself,,, so we can show them this name already exists.

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
Elaina-Lee: Key cannot just be the resourceName from resourceId because user could have used the same resourceName as the input for other workflows that they had added. In that case, two workflow data will be merged due to using the temporary same key.

libs/designer/src/lib/ui/configuretemplate/workflows/customizeWorkflows.tsx
Elaina-Lee: We could - we could show the error message if the id is in the duplicateIds. We need to check not only if the name in the resource exists, but we need to check if it is included in the other input id fields (not-existing in the resource yet). The reason why I chose this is because we need to use this duplicateIds to block 'save' on the usePanelTabs.

libs/designer/src/lib/ui/configuretemplate/panels/configureWorkflowsPanel/usePanelTabs.tsx
preetriti1: With your current changes we should not have duplicateids because we are always putting unique name now

preetriti1 (approved):

PR #7130, opened 2025-04-30
fix(designer): disable designer 'Save' after saving, disable 'Create unit test' when designer state dirty

andrew-eldridge:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

- Designer 'Save' button remains enabled after saving workflow
- Designer 'Create unit test' button is always enabled

## New Behavior

- Add missing update to isDesignerDirty state after saving
- Disable 'Create unit test' when designer state is dirty


apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
```diff
@@ -208,6 +208,11 @@ export default class OpenDesignerForLocalProject extends OpenDesignerBase {
           this.panelMetadata.azureDetails?.workflowManagementBaseUrl
         );
         await this.validateWorkflow(this.panelMetadata.workflowContent);
+
+        this.sendMsgToWebview({
```
ccastrotrejo, line 212: Can we put this function inside the `this.saveWorkflow` from above? Maybe right by the end of the function when we do `writeFileSync` which is where we overwrite the workflow.json file?

What do you think? Just in case if something goes wrong while overwriting the file, the user does realize that indeed the workflow file hasn't been saved

ccastrotrejo (changes requested):

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
andrew-eldridge: Yeah I agree it makes more sense there. I'm not sure why validateWorkflow is called after saveWorkflow here, is that correct?

ccastrotrejo: Yep, the reason of this is
1.- Validate doesn't really work for now. There is a BE task that needs to be completed to make it work it
2.- For vscode extension we allow users to save workflows even if the workflows are invalid, this for the sake of letting user develop more free and save work in progess

ccastrotrejo (approved):

PR #7233, opened 2025-05-14
feat(templates): Fit and Finish of custom templates with exporting methods for portal

preetriti1:
## Type of Change

* [ ] Bug fix
* [x] Feature
* [ ] Other

## Changes
1. Exporting download template as zip in package.
2. Updating tab footer buttons as navigation and action types.
3. Adding descriptions in all tabs as per the figma
4. fit and finish of some ui


libs/designer/src/lib/ui/configuretemplate/templateprofile/manifestform.tsx
```diff
@@ -30,6 +33,17 @@ export const TemplateManifestForm = () => {
...
     <div className="msla-templates-wizard-tab-content" style={{ width: '70%', marginTop: '8px' }}>
+      <DescriptionWithLink
+        text={intl.formatMessage({
+          defaultMessage:
+            'Add details to help template users evaluate this template. The profile includes the information shown to users and settings that control how the template is filtered and displayed.',
+          id: 'tkkN++',
+          description: 'Description for template profile tab',
+        })}
+        linkText={resources.LearnMore}
+        linkUrl="https://learn.microsoft.com/en-us/azure/logic-apps/logic-apps-templates-manifest"
```
ccastrotrejo, line 44: NIT: Is this the right link? It redirects to a 404, I guess it hasn't been published, but wanted to double check

preetriti1: Yes it is not... Divya will get us the link later.. these are just placeholders

PR #7243, opened 2025-05-15
fix(vscode): Fixed error when generating deployment scripts

lambrianmsft:
## Type of Change

* [x] Bug fix
* [ ] Feature
* [ ] Other

## Current Behavior

Logic App must be selected in order to generate deployment scripts. 

## New Behavior

Generate deployment scripts will work even if logic app is not selected. If there are multiple logic apps there will be prompting for one.

## Impact of Change

### DIFF
--- a/apps/vs-code-designer/src/app/commands/generateDeploymentScripts/generateDeploymentScripts.ts
+++ b/apps/vs-code-designer/src/app/commands/generateDeploymentScripts/generateDeploymentScripts.ts
@@ -52,7 +52,7 @@ export async function generateDeploymentScripts(context: IActionContext, node?: 
     ext.outputChannel.appendLog(localize('initScriptGen', 'Initiating script generation...'));
 
     addLocalFuncTelemetry(context);
-    if (Object.keys(node).length > 0 && (await isLogicAppProject(node.fsPath))) {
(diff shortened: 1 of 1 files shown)


DanielleCogs (approved):

apps/vs-code-designer/src/app/commands/generateDeploymentScripts/generateDeploymentScripts.ts
```diff
@@ -52,7 +52,7 @@ export async function generateDeploymentScripts(context: IActionContext, node?:
     ext.outputChannel.appendLog(localize('initScriptGen', 'Initiating script generation...'));
 
     addLocalFuncTelemetry(context);
-    if (node) {
+    if (Object.keys(node).length > 0 && (await isLogicAppProject(node.fsPath))) {
```
ccastrotrejo, line 55: Can we please check that node is not undefined or null. I can see that the `node` in the params is optional, this means that is can be undefined (thats why we are checking it here). Doing `Object.keys(undefined)` will throw an error

ccastrotrejo (changes requested):

ccastrotrejo (approved):

PR #7310, opened 2025-05-23
fix(vscode): Apim connector parameterization check fix

JoaquimMalcampo:
## Type of Change

* [x ] Bug fix
* [ ] Feature
* [ ] Other

Fixes identification of parameterization of APIM connectors to avoid parameterizing an already parameterized logic app.

## Test Plan
Added an additional check for APIM connector in test case

### DIFF
--- a/apps/vs-code-designer/src/app/utils/codeless/__test__/parameterizer.test.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/__test__/parameterizer.test.ts
@@ -40,6 +40,15 @@ describe('areAllConnectionsParameterized', () => {
           displayName: 'func01',
         },
       },
+      apiManagementConnections: {
+        apiManagement1: {
+          apiId:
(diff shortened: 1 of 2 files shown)


apps/vs-code-designer/src/app/utils/codeless/__test__/parameterizer.test.ts
```diff
@@ -40,6 +40,15 @@ describe('areAllConnectionsParameterized', () => {
           displayName: 'func01',
         },
       },
+      apiManagementConnections: {
```
lambrianmsft, line 43: Add tests where the apiManagementConnection has appsettings and parameters

lambrianmsft: Is there different behavior if the apiId is parameterized?

PR #7320, opened 2025-05-27
test(designer): Adding redux state tests

hartra344:


pnpm-lock.yaml
ccastrotrejo, line 1: Is this change needed? I don't see any changes on the package.json

ccastrotrejo (changes requested):

pnpm-lock.yaml
hartra344: I think it's only prettier or something that got a hold of it, it only added a newline at the end. non issue.

Issue #7408, opened 2025-06-02
[Custom templates] Template gallery: Tabs not working as expected


bonicaayala:
### Description
When selecting the tabs `All`, `My templates`, and `Micorosoft authored` the tabs don't display which tab is active causing confusion. 

### Screenshot

![Image](https://github.com/user-attachments/assets/c64a01ae-8c65-4bef-b724-56d3690211bd)

### Video

https://github.com/user-attachments/assets/117b8f09-ee77-4b07-a273-1c9fa16a7589

PR #7413, opened 2025-06-02
fix(Templates): Tab navigation broken UI + Template card css overlapping issue

Elaina-Lee:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Fix
- Fixed tab navigation not working properly (https://github.com/Azure/LogicAppsUX/issues/7408)
- Fixed template card css overlapping on scroll issue

## Impact of Change

https://github.com/user-attachments/assets/28949928-70d0-4860-9c73-451daeb4b97b

### DIFF
--- a/libs/designer/src/lib/ui/templates/filters/templatesearchfilters.tsx
+++ b/libs/designer/src/lib/ui/templates/filters/templatesearchfilters.tsx
@@ -25,6 +25,7 @@ interface GalleryTab {
 }
 
 export interface TemplateSearchAndFilterProps {
+  tabFilterKey?: string;
   tabDetails?: GalleryTab[];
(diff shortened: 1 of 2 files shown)


libs/designer/src/lib/ui/templates/filters/templatesearchfilters.tsx
```diff
@@ -33,6 +33,7 @@ export interface TemplateSearchAndFilterProps {
 }
 
 const templateDefaultTabKey = 'all';
+const tabFilterKey = 'publishedBy';
```
preetriti1: Or let's pass tabfirlterkey from caller.... So agents still uses type and we can use published by

preetriti1 (approved):

PR #7427, opened 2025-06-03
fix(Template): Template parameter definition not updating on resource

Elaina-Lee:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Fix
Parameter definition was only being updated in the state and not in the resource API call

## Impact of Change

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
@@ -231,8 +231,15 @@ export const updateWorkflowParameter = createAsyncThunk(
     const {
       template: { manifest, parameterDefinitions },
     } = getState() as RootState;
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
```diff
@@ -231,8 +231,15 @@ export const updateWorkflowParameter = createAsyncThunk(
     const {
       template: { manifest, parameterDefinitions },
     } = getState() as RootState;
-    const parameter = parameterDefinitions[parameterId];
-    const allParameters = Object.values(parameterDefinitions);
+    const modifiedParameterDefinition = {
```
preetriti1: small typo modifiedParameterDefinitions

preetriti1 (approved):

PR #7428, opened 2025-06-03
fix(Templates): Featured connectors not showing up on dropdown

Elaina-Lee:
## Type of Change

* [X] Bug fix
* [ ] Feature
* [ ] Other

## Fix
Featured connector on the dropdown was showing forever loading sign

## Impact of Change

before fix
image

after fix
image

### DIFF
--- a/libs/designer/src/lib/core/configuretemplate/utils/queries.ts
+++ b/libs/designer/src/lib/core/configuretemplate/utils/queries.ts
@@ -5,6 +5,7 @@ import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { getConnector } from '../../queries/operation';
 import type { NodeOperation } from '../../state/operation/operationMetadataSlice';
 import { getReactQueryClient } from '../../ReactQueryProvider';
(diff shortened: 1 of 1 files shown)


libs/designer/src/lib/core/configuretemplate/utils/queries.ts
```diff
@@ -86,7 +87,7 @@ export const getTemplate = async (templateId: string): Promise<ArmResource<any>>
 
 export const useAllConnectors = (operationInfos: Record<string, NodeOperation>) => {
   const allConnectorIds = Object.values(operationInfos).reduce((result: string[], operationInfo) => {
-    const normalizedConnectorId = operationInfo.connectorId?.toLowerCase();
+    const normalizedConnectorId = operationInfo.connectorId?.toLowerCase()?.replace(`_${workflowIdentifier}`, '');
```
preetriti1, line 90: connectorIds should never have the workflowIdentifier, the bug is somewhere else

Elaina-Lee:
Discussed with Priti, she's looking into the root cause.

PR #7452, opened 2025-06-06
fix(templates): Fixing publishing of templates and other bug fixes

preetriti1:
This is cherry picking fixes made for custom templates to allow publishing single and multi workflow templates.


Elaina-Lee (changes requested):
One more of my fix

Elaina-Lee (approved):

PR #7487, opened 2025-06-09
fix(designer): Parse Dynamic Invoke Based on if NextPage exists

Eric-B-Wu:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

When impelementing the dynamic data call for fetching the next page there was a regression on our dynamic calls that didn't return an array. This addresses the issue to make sure we return the data as-is and has custom logic to parse if pagniation exists.

## Impact of Change

- **Users**: Pagniation works without breaking other dynamic data calls
- **Developers**: Logic change in _executeAzureDynamicApi
- **System**: none

## Test Plan

- x] Unit tests added/updated
- [ ] E2E tests added/updated
- [x] Tested in: Standalone

## Contributors
@preetriti1 for identifying the issue and @DanielleCogs for reverting the original issue in the hotfix branch for this bugfix


takyyon (approved):

libs/logic-apps-shared/src/designer-client-services/lib/base/connector.ts
```diff
@@ -186,13 +186,17 @@ export abstract class BaseConnectorService implements IConnectorService {
...
 
-      // Collect all the values from the initial and subsequent pages
-      const allValues = await this._getAllPagedValues(initialResponse, request.headers);
+      const nextLink = initialResponse.nextLink || initialResponse['@odata.nextLink'];
+      const isArrayPagination = Array.isArray(initialResponse.value) && nextLink;
 
-      return { ...initialResponse, value: allValues };
+      if (isArrayPagination) {
+        const allValues = await this._getAllPagedValues(initialResponse, request.headers);
+        return { ...initialResponse, value: allValues, __usedNextPage: true };
```
preetriti1, line 196: I am assuming with the unwrap it will make sure the value returned is always an array or the expected type

Eric-B-Wu: yeah we only wrap it if it does the pagniation fetch (otherwise return as normal), and only unwrap on the paginated results

PR #7499, opened 2025-06-10
fix(designer): Initialize Variable Object Updating Serialization, deserialization, and validation

Eric-B-Wu:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
1. Modified deserialization logic to address issue when using {} within a token (had a conflict with our token-matching regex)
2. Created a change to support unwrapped tokens (not wrapped in quotes) in Object variables; this enables us to use variable type-interpolation
3. Fixed an issue that we'd always string-interpolate values of objects in the value of a object variable. Now its determined by a combination of factors: a) if the token itself is wrapped in quotes b) the token being wrapped by quotes c) the token type
4. Added more rigorous validation on variable object validation
## Impact of Change

- **Users**: Fixes 2 known bugs on variables (one issue on serialization, one issue on deserialization)
- **Developers**: Logic changes object variables
- **System**: N/A

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@Durga-S1 CSS working with customer on one of the issue


libs/logic-apps-shared/src/parsers/lib/common/helpers/expression.ts
```diff
@@ -59,6 +59,13 @@ export function isTemplateExpression(value: string): boolean {
   return value.charAt(0) === '@' || value.indexOf('@{') > 0;
 }
 
+export function isSingleTokenExpression(value: string): boolean {
```
anthony0t: I'm curious why do we need to special handle this? What was root cause of the issue?

Even if we add a check on ending with '}', it does not guarantee it's a valid expression.

libs/designer-ui/src/lib/editor/initializevariable/util.ts
```diff
@@ -275,102 +310,127 @@ export const validateVariables = (variables: InitializeVariableProps[]): Initial
...
+      default:
+        break;
     }
   });
 
   return errors;
 };
 
+const validateObjectType = (value: string, index: number, errors: InitializeVariableErrors[], intl: any) => {
+  if (isSingleTokenExpression(value)) {
```
anthony0t: I'd suggest to do:
if (hasSingleTokenSegment(valueSegments)) {
    check the type of the valueSegment.token.type is object if it exist.
}

libs/designer-ui/src/lib/editor/initializevariable/util.ts
```diff
@@ -166,6 +168,33 @@ export const parseSchemaAsVariableEditorSegments = (
   }
 };
 
+const wrapUnquotedTokens = (value: string, nodeMap?: Map<string, ValueSegment>): string => {
+  // Find property values that contain only tokens (one or more @{...} tokens with no other text)
```
anthony0t, line 172: I tried to search for the actual serialization logic but didn't find it. Anyway, can we make sure the actual serialization logic matches the logic used here?

Another question, why don't we directly invoke serialization of this value and validate it's a valid JSON? This could avoid duplicate of logic. (something to consider if appropriate and not blocking the review)

Eric-B-Wu: Its because on other parameters the editor passes the valueSegment value back and during serialization the parameterValueToString handles the JSON. However, we can't wait until parameter serialization in the case of variables because we need an actual JSON so that we can parse and pass back a valid variable editor object

libs/designer-ui/src/lib/editor/initializevariable/util.ts
Eric-B-Wu: I'm not too big of a fan on preventing users from saving based on token type especially since we don't have any other validations on it currently; that being said I'll implement this for now and if there are complaints we can address in a follow up item

PR #7576, opened 2025-06-18
fix(templates): Removing publish buttons from non summary tabs

preetriti1:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
This removes the buttons to publish the templates from workflows, profile and parameters tab and now it can only be published from summary tab.
When workflows are being added we also move the template to development state now to prevent validation errors.
This was needed because it was creating incorrect data while saving data in other tabs and resulting in errors and not letting users move forward.

## Impact of Change

- **Users**: Users would be able to add new workflows in template which is already published and save changes correctly
- **Developers**: No new changes
- **System**: No new changes

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@divyaswarnkar @bonica-ayala @Elaina-Lee 

## Screenshots/Videos
image

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
@@ -55,6 +55,7 @@ import {
   getParametersForWorkflow,
   getSupportedSkus,
   getTemplateConnectionsFromConnectionsData,
+  sanitizeConnectorId,
   sanitizeConnectorIds,
   suffixConnectionsWithIdentifier,
   suffixParametersWithIdentifier,
@@ -67,7 +68,12 @@ import {
   updateTemplateParameterDefinition,
 } from '../../state/templates/templateSlice';
 import type { WorkflowTemplateData } from './templates';
-import { initializeNodeOperationInputsData, type NodeDependencies, type NodeInputs } from '../../state/operation/operationMetadataSlice';
+import {
+  initializeNodeOperationInputsData,
+  type NodeOperation,
+  type NodeDependencies,
+  type NodeInputs,
+} from '../../state/operation/operationMetadataSlice';
 import type { WorkflowParameter } from '../../../common/models/workflow';
 import { getAllInputParameters } from '../../utils/parameters/helper';
 import { shouldAddDynamicData } from '../../templates/utils/parametershelper';
@@ -221,11 +227,7 @@ export const loadResourceDetailsFromWorkflowSource = createAsyncThunk(
 export const updateWorkflowParameter = createAsyncThunk(
   'updateWorkflowParameter',
   async (
-    {
-      parameterId,
-      definition,
-      changedStatus,
-    }: { parameterId: string; definition: Template.ParameterDefinition; changedStatus: Template.TemplateEnvironment | undefined },
+    { parameterId, definition }: { parameterId: string; definition: Template.ParameterDefinition },
     { getState, dispatch }
   ): Promise<void> => {
     const service = TemplateResourceService();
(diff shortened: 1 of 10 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
```diff
@@ -346,26 +340,24 @@ export const initializeAndSaveWorkflowsData = createAsyncThunk(
...
+      equals(oldState, 'Development') ? undefined : 'Development',
       /* updateTemplateManifest */ true,
       /* clearWorkflows */ true
     );
 
     dispatch(updateAllWorkflowsData({ workflows: finalWorkflowsData, manifest: updatedTemplateManifest }));
     dispatch(updateConnectionAndParameterDefinitions({ connections: finalConnections, parameterDefinitions: finalParameterDefinitions }));
     dispatch(initializeNodeOperationInputsData(operationsData));
-    if (oldState !== publishState) {
-      dispatch(updateEnvironment(publishState));
```
Elaina-Lee, line 367: Shouldn't we have logic to updateEnvironment to "Development" if oldState is not 'Development'? We are updating the state to "Development" in the resource if the oldState is not "Development"

For example, 
`if (equals(oldState, 'Development')) {
   dispatch(updateEnvironment('Development'));
}`

libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
```diff
@@ -395,17 +385,13 @@ export const saveWorkflowsData = createAsyncThunk(
...
       oldState as Template.TemplateEnvironment,
-      publishState,
+      /* newState */ undefined,
       /* updateTemplateManifest */ false,
       /* clearWorkflows */ false
     );
 
     dispatch(updateAllWorkflowsData({ workflows }));
 
-    if (oldState !== publishState) {
```
Elaina-Lee, line 405: Here makes sense - we are using `saveWorkflowsData` only if we are updating workflow manifest of existing workflows right? So the state will never change.

libs/designer/src/lib/core/configuretemplate/ConfigureTemplateWizard.tsx
```diff
@@ -65,7 +65,7 @@ export const ConfigureTemplateWizard = ({
     }
   };
 
-  const onSaveTemplate = (prevStatus: Template.TemplateEnvironment, newStatus: Template.TemplateEnvironment) => {
+  const onSaveTemplate = (prevStatus: Template.TemplateEnvironment, newStatus?: Template.TemplateEnvironment) => {
```
Elaina-Lee, line 68: If newStatus is not present here, then the logic within showing the Toaster data does not make sense.
Let's say prevStatus = Production, newStatus = undefined (because still Production?),
then by the logic, it will be isNewStatusPublished = false, so it will show the toaster title of "Your template has been unpublished".

Elaina-Lee (changes requested):

libs/designer/src/lib/core/configuretemplate/ConfigureTemplateWizard.tsx
preetriti1: Sure will update the toaster... This is not complete yet since I am waiting from Divya on the scenarios of failure on what to do

libs/designer/src/lib/core/actions/bjsworkflow/configuretemplate.ts
preetriti1: Yes will add

libs/designer/src/lib/core/configuretemplate/ConfigureTemplateWizard.tsx
```diff
@@ -96,7 +96,7 @@ export const ConfigureTemplateWizard = ({
             id: 'ILcDyX',
             description: 'Content for the toaster for after publishing template.',
           })
-        : equals(prevStatus, 'Development')
+        : !newStatus || equals(prevStatus, 'Development')
```
Elaina-Lee, line 99: So by this logic, if we have a published template (prevStatus = 'Production') that is not changing the status (so, newStatus = undefined), this content will be: `Your template in action is in development mode.` which is untrue.

In this PR, I am not seeing how the toaster function being referenced is changing; do we want to revisit this toaster part once more things are decided?

preetriti1: Actually if we are not changing the state we need a different content.. I am not sure what to put here - I would like to not have a content in that case. I can fix it in next PR

PR #7647, opened 2025-06-25
fix(vscode): Update servicesHelper configuration and connector service inheritance

ccastrotrejo:
## Commit Type
- [x] fix - Bug fix

## Risk Level

## What & Why
Fixed VS Code extension servicesHelper configuration to use dynamic `appSettings.ProjectDirectoryPath` instead of hardcoded path for `workflowAppLocation`. Additionally updated connector service inheritance from StandardConnectorService to ensure proper service integration. This ensures the workflow app location is correctly configured based on the actual project directory and services are properly configured.

## Impact of Change
- **Users**: Users can now resolve artifacts based on their logic app project path instead of relying on hardcoded paths or schema client fallbacks
- **Developers**: Updated service inheritance patterns for VS Code extension connector services
- **System**: Improved configuration reliability for VS Code extension environment

## Test Plan
- [ ] E2E tests added/updated
- [x] Tested in: Manual testing in VS Code extension environment with different project directories and connector service configurations

## Contributors
@daviburg @preetriti1

## Screenshots/Videos

https://github.com/user-attachments/assets/8741c607-d497-4d97-aa55-b5ba5bb2032f


daviburg (changes requested):
We should be careful to follow the intended contract and only provide the workflow app folder when requested per the operation manifest.
Please enable just for the operations with
```json
           "dynamicContent": {
               "payloadConfiguration": [
                   "WorkflowAppLocation"
               ]
           },
```

apps/vs-code-react/src/app/designer/servicesHelper.ts
```diff
@@ -175,23 +175,26 @@ export const getDesignerServices = (
...
         delete resolvedConnectionInfo.displayName;
-
-        return {
-          connection: resolvedConnectionInfo,
-        };
+        defaultConfiguration.connection = resolvedConnectionInfo;
       }
 
-      return undefined;
+      return defaultConfiguration;
```
preetriti1, line 200: We should be always adding workflowAppLocation when it is Local and the dynamicContent property has that configurtion set

preetriti1 (changes requested):

libs/logic-apps-shared/src/designer-client-services/lib/vscode/connector.ts
```diff
@@ -0,0 +1,132 @@
...
+    super({
+      ...options,
+      baseUrl: options.apiHubServiceDetails?.baseUrl ?? options.baseUrl,
+      apiVersion: options.apiHubServiceDetails?.apiVersion ?? options.apiVersion,
+    });
+    const { apiVersion, baseUrl, getConfiguration } = options;
+    validateRequiredServiceArguments({ apiVersion, baseUrl, getConfiguration });
+  }
+
+  async getListDynamicValues(
```
preetriti1: YOu also need to do it for gettreeitems which is for folder and file picker...

libs/logic-apps-shared/src/designer-client-services/lib/vscode/connector.ts
```diff
@@ -0,0 +1,132 @@
...
+      });
+    } else {
+      response = await httpClient.post({
+        uri,
+        queryParameters: { 'api-version': apiVersion },
+        content: { parameters: invokeParameters, configuration },
+      });
+    }
+
+    return this._getResponseFromDynamicApi(response, uri);
```
preetriti1: If this is exactly similar to the base class that is StandardConnectorService then move this to a protected method and just call it from both vscodeservice and standard service, that way you only have code until getting the configuration

libs/logic-apps-shared/src/designer-client-services/lib/vscode/connector.ts
ccastrotrejo: It looks like the `getTreeDynamicValues` is not supported for Standard. Thats why I didn't implement it. Is that the function you meant with `gettreeitems`?
https://github.com/Azure/LogicAppsUX/blob/5f08d02445a1ac686285fd28654396549ea65fda/libs/logic-apps-shared/src/designer-client-services/lib/standard/connector.ts#L146
image

libs/logic-apps-shared/src/designer-client-services/lib/vscode/connector.ts
ccastrotrejo: Will do that thanks!

libs/logic-apps-shared/src/designer-client-services/index.ts
```diff
@@ -35,3 +35,4 @@ export * from './lib/userpreference';
 export * from './lib/resource';
 export * from './lib/role';
 export * from './lib/providers/FavoriteContextProvider';
+export * from './lib/vscode';
```
preetriti1: Why do we have vscode services in this package? can we keep the service implementation of vscode in vscode package only? this way we wont be adding additional code which will not be used in other portals

Issue #7518, opened 2025-06-12
Dictionary editor not populating friendly name of dynamic inputs


DanielleCogs:
### Severity

P2 - High (Major functionality broken)

### Describe the Bug with repro steps

Token metadata should be refreshed and token friendly name should be shown, this is not happening

1. Make a logic app and add microsoft forms connectors
2. Add new ADO connector like the below screenshot, adding inputs from forms

### What type of Logic App Is this happening in?

Consumption (Portal)

### Which operating system are you using?

Windows

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json

```

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/a44cc382-26fd-4d36-a46d-a57f071d9d0b)

### Environment

NA

PR #7655, opened 2025-06-26
Feat(designer): Dynamic Output not updating titles of view model editors

Eric-B-Wu:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

- [] Medium - Moderate changes, some user impact

## What & Why

Fixed token titles not updating in editor components after dynamic data loads. The existing code only updated tokens in the main parameter values, but some editors store additional tokens in their editorViewModel property that were getting stale titles. Now both locations get updated when new dynamic outputs are fetched.
Fixes https://github.com/Azure/LogicAppsUX/issues/7518

Added optimizations to avoid creating new objects when no changes are needed, reducing unnecessary re-renders.

## Impact of Change

- **Users**: Token titles in editors such as dictionary editor will now display the correct dynamic field names instead of showing stale/incorrect titles
- **Developers**: existing updateExistingInputTokenTitles function now handles both parameter.value and parameter.editorViewModel automatically
- **System**:  Reduced object allocations and Redux re-renders through reference equality optimizations when no token updates are needed

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors

@preetriti1  @DanielleCogs for doing the initial investigation

## Screenshots/Videos

Before:
Screenshot 2025-06-26 at 3 55 31 PM

After:
Screenshot 2025-06-26 at 3 54 11 PM


libs/designer/src/lib/core/state/operation/operationMetadataSlice.ts
```diff
@@ -656,29 +656,104 @@ export const operationMetadataSlice = createSlice({
...
+const updateTokenTitlesInViewModel = (viewModel: any, tokenTitles: Record<string, string>): any => {
+  if (!viewModel || typeof viewModel !== 'object') {
+    return viewModel;
+  }
+
+  // Handle ValueSegment arrays - base case for our editors
+  if (Array.isArray(viewModel) && viewModel.every((item) => isValueSegment(item))) {
+    let hasChanges = false;
+    const updatedSegments = viewModel.map((segment) => {
+      if (isTokenValueSegment(segment) && segment.token?.key) {
```
preetriti1, line 669: Why would token not contain key?

Eric-B-Wu: this is moreso for the check on segment.token where even though we check if its a tokenSegment in isTokenValueSegment, this function doesn't define the type, so segment.token can be undefined still in the view of Typescript

preetriti1 (approved):

PR #7743, opened 2025-07-09
fix(mcp): Change Connector browse view to only fetch/use managedApis

Eric-B-Wu:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

Changing connector browse view to only fetch/use managed apis, this includes removing the filtering logic so that we only have actions and shared connectors.

## Impact of Change

- **Users**: Users should only see managed apis in the mcp browse view
- **Developers**: Adding a couple of queries so that it only fetches managed apis
- **System**: Will make the fetching quicker as we won't need to process builtin/custom apis

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors

@Elaina-Lee, @preetriti1 tagging other mcp devs


libs/designer/src/lib/core/queries/browse.ts
```diff
@@ -26,6 +26,35 @@ const favoriteActionQueryPageSize = 10;
...
+  const data = useMemo(() => {
+    const azure = azureData?.pages.flatMap((page) => page.data) ?? [];
+    return azure.filter((operation) => operation !== undefined);
+  }, [azureData]);
+
+  const hasNextPage = azureHasNextPage;
+  const isLoading = azureFetching;
+
+  return useMemo(() => ({ data, isLoading, hasNextPage }), [data, isLoading, hasNextPage]);
+};
```
preetriti1: Do we really need this?

libs/designer/src/lib/core/queries/browse.ts
```diff
@@ -26,6 +26,35 @@ const favoriteActionQueryPageSize = 10;
...
+
+export const useAllManagedApiIdsWithActions = () => {
+  const allOperations = useAllManagedOperations();
+
+  return useMemo(() => {
+    const actions = allOperations.data.filter((operation) => !operation.properties?.trigger);
+
+    const apiIdsWithActions = [...new Set(actions.map((action) => cleanConnectorId(action?.properties.api.id)))];
```
preetriti1: Hey I didnt ask to do this..... I think there is somewhere in connector data which says capability... if it is non empty and doesnt have action then we should filter it out.. Here we are just making unnecessary operations api call which we should not do.

For now lets skip this part if the capability is not present in first class properly, we will just show empty operations when connector is selected

libs/designer/src/lib/ui/mcp/panel/OperationSelectionPanel.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export const OperationSelectionPanel = ({ onDismiss, onBack }: ConnectorDetailPanelProps) => {
+  const intl = useIntl();
+  const styles = useConnectorSelectionStyles();
+
+  const { selectedConnectorId } = useSelector((state: RootState) => state.mcpPanel);
+  const { data: allConnectors } = useAllManagedConnectors();
+
+  const selectedConnector = useMemo(() => allConnectors?.find((c) => c.id === selectedConnectorId), [allConnectors, selectedConnectorId]);
```
preetriti1: I think we should just do a getConnector,this will anyways be needed in other parts during connection creation, so doing a get connector would just cache that in teh query cache.

libs/designer/src/lib/core/queries/browse.ts
Eric-B-Wu: I think we did this in designer due to not having capability not being present in all connectors and that we make the call to get all the operations for search; but if we assume the second part then I'll remove the fetching of managed operations and purely filter on capabilities

libs/designer/src/lib/ui/mcp/panel/OperationSelectionPanel.tsx
Eric-B-Wu: yeah this was all temporary, the operations selection panel I haven't started yet; this is mostly just placeholder work. I can move to this logic when our connection service is implemented

PR #7821, opened 2025-07-16
feat(mcp): Creating Paramters Tab for MCP Operations 

Eric-B-Wu:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Implements the Edit Operations panel allowing users to modify operation parameters and descriptions
- Add/remove optional parameters with conditional visibility
- Parameter value editing with local state management, then serialization (Save) with redux integration
- E2E experience now implemented
- Upgraded SearchableDropdown from Fluent UI v8 to v9 to fix conflict with drawer overlay

## Impact of Change

- **Users**: Can now edit operation parameters and descriptions through a dedicated panel
- **Developers**: New `EditOperation` component with ref-based API for getting changes
- **System**: Minimal performance impact - uses local state until save to reduce Redux dispatches

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors

@preetriti1, @Elaina-Lee 
## Screenshots/Videos

https://github.com/user-attachments/assets/f9c2946d-29c7-49c3-b84d-46d2d8d7bfd5

After merge/change with new UI

https://github.com/user-attachments/assets/677cf7f4-78f0-4047-83ac-6384f8b03a67


libs/designer-ui/src/lib/token/inputToken.tsx
```diff
@@ -302,7 +306,9 @@ export const InputToken: React.FC<InputTokenProps> = ({ value, brandColor, icon,
     >
       <div className={styles.tokenIcon} style={iconStyle} aria-hidden="true" />
       {isSecure && <LockClosedRegular className={styles.secureIcon} aria-hidden="true" />}
-      <div className={styles.tokenTitle}>{title}</div>
+      <div className={styles.tokenTitle} style={{ maxWidth: '200px' }} title={title}>
+        {title}
+      </div>
```
preetriti1: Where are we using this?

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanelInner.tsx
```diff
@@ -17,62 +26,173 @@ export const EditOperationPanelInner = () => {
...
+    handleDismiss();
+  }, [isDirty, INTL_TEXT.unsavedChangesMessage, resetToOriginalState, handleDismiss]);
+
+  const handleDismissWithReset = useCallback(() => {
+    resetToOriginalState();
+    handleDismiss();
+  }, [resetToOriginalState, handleDismiss]);
+
+  const handleSave = useCallback(() => {
+    if (!selectedOperationId || !parameters?.parameterGroups) {
```
preetriti1, line 83: Save should not be enabled when there is nothing

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanelInner.tsx
```diff
@@ -17,62 +26,173 @@ export const EditOperationPanelInner = () => {
...
+      return;
+    }
+
+    // Get all changes from the EditOperation component
+    const latestConditionalVisibilityChanges = editOperationRef.current?.getConditionalVisibilityChanges() || {};
+    const parameterValueChanges = editOperationRef.current?.getParameterValueChanges() || {};
+    const descriptionChange = editOperationRef.current?.getDescriptionChange();
+
+    Object.entries(latestConditionalVisibilityChanges).forEach(([parameterId, isVisible]) => {
+      let parameterGroupId: string | undefined;
```
preetriti1, line 98: Use simple for loop its better on perf than these iterators

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanelInner.tsx
```diff
@@ -17,62 +26,173 @@ export const EditOperationPanelInner = () => {
...
+
+      Object.entries(parameters.parameterGroups).forEach(([groupId, group]) => {
+        const param = group.parameters.find((p) => p.id === parameterId);
+        if (param) {
+          parameterGroupId = groupId;
+        }
+      });
```
preetriti1, line 105: We must already have a method to get the groupid for a parameter, if not this hsould be added in the parameters helper

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanelInner.tsx
```diff
@@ -17,62 +26,173 @@ export const EditOperationPanelInner = () => {
...
+    resetToOriginalState();
+    handleDismiss();
+  }, [isDirty, INTL_TEXT.unsavedChangesMessage, resetToOriginalState, handleDismiss]);
+
+  const handleDismissWithReset = useCallback(() => {
+    resetToOriginalState();
+    handleDismiss();
+  }, [resetToOriginalState, handleDismiss]);
+
+  const handleSave = useCallback(() => {
```
preetriti1, line 82: Unfortunately we cannot just update everything only on save button click, we wont be able to support dynamic content with this model...
the parameter changes need to be present in the store to use the dynamic pipeline which gets the dependent parameters from store to use the commited value... this would mean we have to redo dynamic pipeline which makes this WI very complex.

Just for save changes we will have to keep the original state of parameters in components and if user doesnt save we override the parameter values in store using that original state. if they do save changes we can update conditional visibility but values need to be updated as user is updating them.

libs/designer/src/lib/ui/mcp/parameters/EditOperation.tsx
```diff
@@ -0,0 +1,563 @@
...
+  useEffect(() => {
+    if (!parameters?.parameterGroups) {
+      setLocalConditionalVisibility({});
+      return;
+    }
+
+    const initialVisibility: Record<string, boolean> = {};
+    Object.values(parameters.parameterGroups).forEach((group) => {
+      group.parameters.forEach((param) => {
+        if (!param.required && param.conditionalVisibility) {
```
preetriti1, line 72: Are we able to show important parameters correctly in the initial load?

preetriti1 (commented):
Elaina - Once this is checked in, let fix the PR comments

libs/designer-ui/src/lib/token/inputToken.tsx
Eric-B-Wu: oh this was just sidetracked when I noticed this in the External chat that tokens were too long in dictionary editor was changing the shape of the editor; but I was too deep in my parameters tab, but not yet complete to switch branches, so it was sorta snuck in

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanelInner.tsx
Eric-B-Wu: I'm scanning every parameter group until you find a matching parameter, think we'll still need a double for loop; I guess we can break out early something like:
```
for (const [parameterId, isVisible] of Object.entries(latestConditionalVisibilityChanges)) {
  let parameterGroupId: string | undefined;

  for (const [groupId, group] of Object.entries(parameters.parameterGroups)) {
    if (group.parameters.find((p) => p.id === parameterId)) {
      parameterGroupId = groupId;
      break;
    }
  }

  if (parameterGroupId) {
    dispatch(
      updateParameterConditionalVisibility({
        nodeId: selectedOperationId,
        groupId: parameterGroupId,
        parameterId,
        value: isVisible,
      })
    );
  }
}
```

libs/designer/src/lib/ui/mcp/parameters/EditOperation.tsx
Eric-B-Wu: think important is determined beforehand when adding conditionalVisbility

PR #7831, opened 2025-07-17
fix(designer): Disable Channels until teams integration is ready

takyyon:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
Request/Response channel is being deprecated in favor of A2A behavior. We need to disable the channels tab until Teams integration is ready.

## Impact of Change

- **Users**: Users won't be able to setup channels since Request/Response is no longer supported (BE still supports it but we don't want users to depend on it)
- **Developers**: None
- **System**: A2A will be the new way of building conversational which was previously done with channels

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@rohithah  @divyaswarnkar 

## Screenshots/Videos
image

### DIFF
--- a/libs/designer/src/lib/common/hooks/experimentation.ts
+++ b/libs/designer/src/lib/common/hooks/experimentation.ts
@@ -18,8 +18,8 @@ export function useShouldEnableDynamicConnections(): boolean | null {
   return enabled;
 }
 
-export function useChannelsTabForAgentLoop(): boolean | null {
-  const [enabled, setEnabled] = useState<boolean | null>(null);
+export function useChannelsTabForAgentLoop(): boolean {
+  const [enabled, setEnabled] = useState<boolean>(false);
   useEffect(() => {
     const check = async () => {
       try {


libs/designer/src/lib/common/hooks/experimentation.ts
```diff
@@ -17,3 +17,20 @@ export function useShouldEnableDynamicConnections(): boolean | null {
...
 }
+
+export function useChannelsTabForAgentLoop(): boolean {
+  const [enabled, setEnabled] = useState<boolean>(false);
+  useEffect(() => {
+    const check = async () => {
+      try {
+        const result = await enableChannelsInAgentLoop();
+        setEnabled(result);
+      } catch (_e) {
```
ccastrotrejo, line 28: Nit: if you are not using the error you can only to the catch {}

takyyon: I will actually refactor to merge all the exp calls in a helper function (since this isn't mantain-able longer term with many exps being added)

PR #7839, opened 2025-07-17
fix(designer): Send original response if dynamic data pagination fails

Eric-B-Wu:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

Fixes an issue where on fetching nextPage on dynamic data that if the nextPage fails that we no longer send the original response. This is an issue because non-ARM connectors will fail due to not having the correct auth token (this will need to be revisited). For now this PR will handle it so that we at least send the original payload.

## Impact of Change

- **Users**: Fixes regression for dynamic data pagination failures.
- **Developers**: Ensures original payload is sent despite paging errors.
- **System**: None

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors

@preetriti1, @DanielleCogs


libs/logic-apps-shared/src/designer-client-services/lib/helpers.ts
```diff
@@ -75,6 +77,13 @@ export function getTemplateManifestFromResourceManifest(resourceManifest: any):
 
 export function unwrapPaginatedResponse(response: any): any {
   if (response && response.__usedNextPage) {
+    if (response.__paginationIncomplete) {
+      LoggerService().log({
+        message: `Pagination request unsuccessful, some values may be missing. ${response.__paginationError || ''}`,
+        level: LogEntryLevel.Error,
+        area: 'unwrapPaginatedResponse',
+      });
+    }
```
preetriti1: Can you move this to caller so we can log the connectorId and operationId

PR #7872, opened 2025-07-23
fix(Mcp): Mcp bug bash fix for card unselection behavior

Elaina-Lee:
## Commit Type

- [X] fix - Bug fix

## Risk Level

- [X] Low - Minor changes, limited scope

## What & Why
It was not unselecting the operation on card click if it has been selected. This was due to using name in store using getResourceId, but still using id on selection passing around. Since this is an arm resource, instead of manually parsing id, we are sticking to name, which is the same value. As well, one minor content revision suggested by Bonica

## Impact of Change

- **Users**: 
- **Developers**: 
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@bonica-ayala


libs/designer/src/lib/ui/mcp/operations/OperationSelectionGrid.tsx
```diff
@@ -44,7 +44,7 @@ const OperationCell = ({ operation, isSelected, showConnectorName, onCardClick,
             </Body1>
             <Checkbox
               checked={isSelected}
-              onChange={(_, data) => onCheckboxChange(id, data.checked === true)}
+              onChange={(_, data) => onCheckboxChange(name, data.checked === true)}
```
preetriti1, line 47: Can you please make sure on operation deselection we are updating the stores correctly and the operation is deleted which is unselected.

Elaina-Lee: This PR is not just fixing the behavior, is fixing the root cause.

https://github.com/user-attachments/assets/b349e734-7fa6-427f-8b34-621787fd243c

preetriti1: Did you make sure when you click on save and close this panel correct operations show up in the list in the UI?

Elaina-Lee: Thanks for pointing it out - addressed it

preetriti1 (changes requested):
Instead of clearing like this... Please issue a deletenodes command and that should go to both operations slice and connection slice.. sorry I can't approve with this... No need to update existing dispatch, but get the deleted ones and dispatch delete call

libs/designer/src/lib/core/state/connection/connectionSlice.ts
```diff
@@ -109,6 +109,12 @@ export const connectionSlice = createSlice({
       const { nodeId } = action.payload;
       delete state.connectionsMapping[nodeId];
     },
+    removeNodesConnectionData: (state, action: PayloadAction<{ nodeIds: NodeId[] }>) => {
+      const { nodeIds } = action.payload;
+      for (const nodeId of nodeIds) {
+        delete state.connectionsMapping[nodeId];
+      }
+    },
   },
```
preetriti1: Remove this, and add a reducer in this slice which would wait for deinitializeOperations.fulfilled and then you can do this, we dont have to add new reducer action.

libs/designer/src/lib/core/actions/bjsworkflow/mcp.ts
```diff
@@ -163,3 +166,12 @@ export const initializeConnectionMappings = createAsyncThunk(
     }
   }
 );
+
+export const deinitializeOperations = createAsyncThunk(
+  'deinitializeOperations',
+  async ({ operationIds }: { operationIds: string[] }, { dispatch }) => {
+    dispatch(deinitializeOperationInfos({ ids: operationIds }));
+    dispatch(deinitializeNodes(operationIds));
+    dispatch(removeNodesConnectionData({ nodeIds: operationIds }));
```
preetriti1: instead just return operationIds here, so that can be used in connections slice.

PR #7889, opened 2025-07-25
fix(Mcp): Operation data selection change inconsistent behavior

Elaina-Lee:
## Commit Type

- [X] fix - Bug fix

## Risk Level

- [X] Low - Minor changes, limited scope

## What & Why
Fixed - Only updates the newly selected operation in store, this way the original added data is not overrwitten (reset) to the originally fetched data.
Fixed - if there's no change of selected operations, the submit function won't have close the panel, and users will be confused
Fixed - only if there's selection change, the submit will trigger => if there's deselection change, the submit will also happen

## Impact of Change

- **Users**: Prevents unexpected data loss when switching operations, no more confusion about panel not closing
- **Developers**: fixx internal to selection logic
- **System**: None

## Test Plan

- [ ] E2E tests added/updated
- [X] Manual testing completed
- [ ] Tested in: 

## Screenshots/Videos

https://github.com/user-attachments/assets/5c32aa92-9636-4218-8a17-dfc3d84b3ae9

https://github.com/user-attachments/assets/a6fb70d5-77d7-49da-ac59-65cbf225a0df


libs/designer/src/lib/ui/mcp/panel/connector/usePanelTabs.tsx
```diff
@@ -49,28 +49,34 @@ export const useMcpConnectorPanelTabs = (): McpPanelTabProps[] => {
...
+    return selectedOperations.filter((operationId) => !Object.keys(operationInfos).includes(operationId));
+  }, [operationInfos, selectedOperations]);
+
   const deselectedOperationIds = useMemo(() => {
     return Object.keys(operationInfos).filter((operationId) => !selectedOperations.includes(operationId));
   }, [operationInfos, selectedOperations]);
 
   const handleSubmit = useCallback(() => {
-    if (selectedConnectorId && selectedOperations.length > 0) {
-      const selectedOperationsData = selectedOperations.map((operationId) => ({
```
preetriti1: Shouldn't the selectedOperations be always non empty for user to save the selection. I think we should disable the button if none are selected.

preetriti1 (changes requested):

libs/designer/src/lib/ui/mcp/panel/connector/usePanelTabs.tsx
Elaina-Lee: They can deselct operations, in which case they should be able to proceed right? Also maybe they have changed the connection, which would have nothing to do with selected operations. I don't think we should disable buttons if we consider those.

preetriti1: ya so those are newlyselectedoperations, but selected operations should still be non empty no? otherwise are we enabling the save button if someone deselects all operations?

preetriti1 (approved):
Make sure you merge with latest main because I removed clearAllSelections and clonePanel dispatch from here

PR #7919, opened 2025-07-30
feat(Mcp): Updating the parameter field to support model and user input

Elaina-Lee:
## Commit Type

- [X] feature - New functionality
- [ ] fix - Bug fix

## Risk Level

- [X] Medium - Moderate changes, some user impact

## Impact of Change

- **Users**: 
- **Developers**: 
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:


libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanel.tsx
```diff
@@ -20,12 +22,14 @@ export const EditOperationPanel = () => {
...
 
-  const { selectedOperationId, operationMetadata, isOpen, panelMode } = useSelector((state: RootState) => ({
+  const { selectedOperationId, operationMetadata, isOpen, panelMode, inputParameters } = useSelector((state: RootState) => ({
     selectedOperationId: state.mcpSelection.selectedOperationId,
     operationMetadata: state.operations.operationMetadata,
     isOpen: state.mcpPanel?.isOpen ?? false,
     panelMode: state.mcpPanel?.currentPanelView ?? null,
+    inputParameters: state.operations.inputParameters,
   }));
+  const parameters = selectedOperationId ? inputParameters[selectedOperationId] : null;
```
preetriti1: You have to use useMemo, otherwise you wont get latest changes

libs/designer/src/lib/ui/mcp/parameters/EditOperation.tsx
```diff
@@ -319,8 +325,11 @@ export const EditOperation = ({ description, handleDescriptionInputChange, onPar
                         parameter={param}
                         isConditional={isConditional}
                         onParameterVisibilityUpdate={onParameterVisibilityUpdate}
-                        handleParameterValueChange={handleParameterValueChange}
+                        parameterInputType={userInputParamIds[param.id as string] ? 'user' : 'model'}
```
preetriti1: Why do we need to typecast it as string, isnt it string?

libs/designer/src/lib/ui/mcp/parameters/EditOperation.tsx
```diff
@@ -319,8 +325,11 @@ export const EditOperation = ({ description, handleDescriptionInputChange, onPar
...
+                        onParameterInputTypeChange={handleParamterInputTypeChange}
                         handleRemoveConditionalParameter={handleRemoveConditionalParameter}
+                        parameterError={parameterErrors[param.id as string]}
+                        removeParameterError={removeParameterError}
```
preetriti1, line 449: So basically we are updating the parameter error if user fills in the value right?

libs/designer/src/lib/ui/mcp/parameters/parameterfield.tsx
```diff
@@ -0,0 +1,233 @@
...
+        properties: propertiesToUpdate,
+        isTrigger: false,
+        operationInfo,
+        connectionReference: reference,
+        nodeInputs,
+        dependencies,
+        updateTokenMetadata: false,
+      };
+
+      dispatch(updateParameterAndDependencies(updateParameterAndDependencyPayload));
```
preetriti1: No need to store this in variable we can directly inline.

libs/designer/src/lib/ui/mcp/parameters/parameterfield.tsx
```diff
@@ -0,0 +1,233 @@
...
+        parameters: [
+          {
+            groupId,
+            parameterId,
+            propertiesToUpdate: { value: newValueSegments },
+          },
+        ],
+        isUserAction: true,
+      };
+      dispatch(updateNodeParameters(updateParametersPayload));
```
preetriti1: We don tneed to do this, this is already happening in the above dispatch function

libs/designer/src/lib/ui/mcp/parameters/parameterfield.tsx
```diff
@@ -0,0 +1,233 @@
...
+    // Reset parameter value when model is selected
+    if (data.value === 'model') {
+      onHandleRemoveError();
+      onParameterValueChange({
+        value: [
+          {
+            id: parameter?.value?.[0]?.id,
+            type: 'literal',
+            value: '',
+          },
```
preetriti1: use the function from './utils/parameters/segment.ts' createLiteralValueSegment('')

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanel.tsx
```diff
@@ -38,15 +42,61 @@ export const EditOperationPanel = () => {
...
+      description: 'Error message for parameter is empty',
+    }),
   };
 
+  const userInputParamIds = getUserInputParamIds();
+
+  const runValidationOnUserInput = useCallback(() => {
+    if (!selectedOperationId || !parameters?.parameterGroups) {
+      return;
+    }
```
preetriti1: I dont want us to expose these parameters and its state in the main panel, can't we do this in EditOperation, where on Save you can invoke saying runValidation and showErrors. 
save is not going to close the panel so we should be good

preetriti1 (commented):
I think few changes are required here.

libs/designer/src/lib/ui/mcp/panel/operation/EditOperationPanel.tsx
preetriti1: This is the only comment that needs to be addressed later

Elaina-Lee: We cannot... The save function goes to the footer content which has to be rendered here in the EditOperationPanel. I had to move it out of the EditOperation for this reason.

Elaina-Lee: EditOperation is only the body of the panel, and save is on the footer.

PR #8295, opened 2025-09-25
Add get func children

ccastrotrejo:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## Impact of Change

- **Users**: 
- **Developers**: 
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:

### DIFF
--- a/apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
+++ b/apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
@@ -265,6 +265,7 @@ async function startFuncTask(
             const response = await sendRequestWithTimeout(context, statusRequest, statusRequestTimeout, undefined);
             if (response.parsedBody.state.toLowerCase() === 'running') {
               funcTaskReadyEmitter.fire(workspaceFolder);
+              taskInfo.childProcessId = [await pickChildProcess(taskInfo), await pickFuncHostChildProcess(taskInfo)];
               return taskInfo;
             }
           } catch (error) {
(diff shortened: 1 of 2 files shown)


apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -361,3 +362,31 @@ export async function getWindowsChildren(pid: number): Promise<OSAgnosticProcess
     return { command: c.name, pid: c.pid };
   });
 }
+
+async function pickFuncHostChildProcess(taskInfo: IRunningFuncTask): Promise<string | undefined> {
```
andrew-eldridge, line 366: This file now has "findChildProcess", "pickChildProcess", and "pickFuncHostChildProcess". Is it possible to better differentiate these names or add a high-level summary on the functions explaining the difference?

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -361,3 +362,31 @@ export async function getWindowsChildren(pid: number): Promise<OSAgnosticProcess
...
+  }
+
+  const children: OSAgnosticProcess[] =
+    process.platform === Platform.windows ? await getWindowsChildren(funcPid) : await getUnixChildren(funcPid);
+  const childRegex = /(func|dotnet)(\.exe|)?$/i;
+  let child: OSAgnosticProcess | undefined = children.reverse().find((c) => childRegex.test(c.command || ''));
+
+  // If child is null or undefined, look one level deeper in child processes
+  if (!child) {
+    for (const possibleParent of children) {
```
andrew-eldridge, line 379: It looks like this checks children of children on the output of pickChildProcess, which checks children itself. Is the idea that we need to check 3 levels down?

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -265,6 +265,7 @@ async function startFuncTask(
             const response = await sendRequestWithTimeout(context, statusRequest, statusRequestTimeout, undefined);
             if (response.parsedBody.state.toLowerCase() === 'running') {
               funcTaskReadyEmitter.fire(workspaceFolder);
+              taskInfo.childProcessId = [await pickChildProcess(taskInfo), await pickFuncHostChildProcess(taskInfo)];
```
andrew-eldridge, line 268: Does this mean there are two separate processes that need killed from pickChildProcess and pickFuncHostChildProcess?

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
ccastrotrejo: Yep, I will try to make the differences in the code more obvious. I can explain it here as well
- findChildProcess: Finds any child process
- pickChildProcess: Picks the child process from the fun process. We should reuse the `findChildProcess` inside of here.
-  pickFuncHostChildProcess: Picks the child process of the powershell process, which is the func process, and the children of the func process - This only happens in Windows

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
ccastrotrejo: Yes, this only happens in Windows. Basically in windows when we execute the func host start. The following process tree gets created. Therefore we need to kill all of them. I think this happens because of dotnet 8. Tagging @lambrianmsft who knows more about it

- Powershell process
   - Func process
     -  Func process

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
ccastrotrejo: Yes

Issue #8274, opened 2025-09-22
VS Code at start up silently wipes out the Lib/Custom folder and that breaks deployment to Azure.


SergeiMou:
### Severity

P2 - High (Major functionality broken)

### Describe the Bug with repro steps

For some of my logic apps, upon loading the workspace into VSCode, for some unknown reason this piece of code from the [My functions].csproj file gets executed which wipes out the LogicApps\Lib\Custom folder, containing my functions?

<img width="713" height="167" alt="Image" src="https://github.com/user-attachments/assets/6d31f058-fde8-40fa-9bbf-7db47efb0062" />

The consequence is that when you deploy the app to Azure, the deployment procedure, seeing that the Lib folder is empty, deletes all previously deployed functions in Azure, so the workflows get broken. It took quite a bit of time to understand what and how it happens, but I still don't understand why and how to fix it. The workaround is clear - compile before deploy, but how can I prevent this piece of code from getting executed and wiping out the Lib folder?  As with the other issue that I posted today, #8272 namely, it only occurs in one dev environment, but not in another. So, the question is again - how can I turn it into an actionable item. Is there "verbose" logging option that I can enable and attach the log to the issue? I am going to open a support ticket, perhaps MS support knows the magic word to make VSCode behave as it should.

### What type of Logic App Is this happening in?

Standard (VSCode)

### Which operating system are you using?

Windows

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json

```

### Environment

edge, Extension 5.165.8

Issue #8273, opened 2025-09-22
Debugger is throwing an error


SergeiMou:
### Severity

P2 - High (Major functionality broken)

### Describe the Bug with repro steps

For one of the logic apps, when I hit F5, debugger is throwing the error below

<img width="632" height="267" alt="Image" src="https://github.com/user-attachments/assets/72390191-c487-4b4b-a581-76afcdfa6b74" />

I don't see anything in this logic app that would stand it apart from the rest. How can I turn it into an actionable item? Is there the `verbose' logging option so that I can attach the logs and the issue can be properly investigated?

### What type of Logic App Is this happening in?

Consumption (Portal)

### Which operating system are you using?

Windows

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json

```

### Environment

Edge, Extension 5.165.8

SergeiMou:
Hi @wsilveiranz 
I was able to reproduce the issue. It happens when the following two conditions are met: 
1) Your solution contains both Workflows and Functions
2) There are no calls to the functions in your workflows

Workflow like this will give you the error:
<img width="460" height="317" alt="Image" src="https://github.com/user-attachments/assets/fb01f179-072e-4bdf-8866-37a6ff47d347" />
Error is gone: 
<img width="592" height="512" alt="Image" src="https://github.com/user-attachments/assets/6bba7411-17b9-4387-b9b8-2faf356850d8" />

Real life scenario: you create a new workspace, VSCode gives you the sample workflow and the default celsius to fahrenheit function. You start building your logic, delete the call to the default function, etc... As you progress, you at some point want to make a local run, you hit F5 and get the error. Once you add a call to any function in your Functions project, the error goes away. 
So, this is kind of not a breaking error - if you don't call a function there is not much to debug, but this is huge a distractor. Once you see the error you start trying to fix it thinking that your project is broken and debugger is not working at all.

Working on the reproducing the error I've noticed another issue. Even if debugger starts with no errors 
I can't set a breakpoint in workflow json files. I am getting the message "No symbols have been loaded for this document" when attempting to set a breakpoint and in the "Debug Console" windows I see many warning like this: 
Skipped loading symbols. Module is optimized and the debugger option 'Just My Code' is enabled.

[...]

PR #8307, opened 2025-10-01
fix(vscode): Build custom code on deploy, handle error when fails to attach to custom code worker

andrew-eldridge:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
Fixes edge case where custom code project is never built before running "deploy" command. Checks whether build is needed prior to deploying. Also fixes issue where an error will occur when attempting to attach to custom code worker process when there is no local function action in workflow. Addresses #8274 and #8273.

## Impact of Change

- **Users**: Fixes bugs (1) where custom code is not built (empty {la}\lib\custom\*) when deploying, and (2) where error occurs on debugging custom code project with no local function actions
- **Developers**: N/A
- **System**: N/A

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@andrew-eldridge

### DIFF
--- a/apps/vs-code-designer/src/app/commands/deploy/deploy.ts
+++ b/apps/vs-code-designer/src/app/commands/deploy/deploy.ts
@@ -55,7 +55,8 @@ import type { Uri, MessageItem, WorkspaceFolder } from 'vscode';
 import { deployHybridLogicApp, zipDeployHybridLogicApp } from './hybridLogicApp';
 import { createContainerClient } from '../../utils/azureClients';
 import { uploadAppSettings } from '../appSettings/uploadAppSettings';
-import { resolveConnectionsReferences } from '@microsoft/logic-apps-shared';
+import { isNullOrUndefined, resolveConnectionsReferences } from '@microsoft/logic-apps-shared';
+import { buildCustomCodeFunctionsProject } from '../buildCustomCodeFunctionsProject';
 
 export async function deployProductionSlot(
   context: IActionContext,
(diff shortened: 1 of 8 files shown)


ccastrotrejo (approved):

apps/vs-code-designer/src/app/commands/deploy/deploy.ts
```diff
@@ -87,6 +88,18 @@ async function deploy(
...
+  try {
+    if (!isNullOrUndefined(workspaceFolder)) {
+      const logicAppNode = workspaceFolder.uri;
+      // Only build custom code projects on open designer if custom code binaries don't already exist in the logic app folder
+      if (!(await fse.pathExists(path.join(logicAppNode.fsPath, 'lib', 'custom')))) {
+        await buildCustomCodeFunctionsProject(actionContext, logicAppNode);
+      }
+    }
+  } catch {
+    // Ignore error if thrown, forego building custom code project
```
ccastrotrejo: NIT: in the catch of the error, do you think it would be useful to log some telemetry in case we get a github issue, so we can find sort of figure out that the custom code project indeed didn't got build and the reason behind it?

andrew-eldridge: It appears this is actually already done in the buildCustomCodeFunctionProject file (it handles errors and writes to telemetry/debug outputs). So I removed the catch block here since it's redundant

PR #8370, opened 2025-10-13
fix(designerv2): Pass in the isDraft flag while running the workflow

takyyon:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
Running workflow in draft mode should pass in the correct flag so it saves the correct version

## Impact of Change

- **Users**: While running a draft workflow, it will only save the draft workflow
- **Developers**: None
- **System**: None

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:

### DIFF
--- a/apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
@@ -46,8 +46,8 @@ import * as portfinder from 'portfinder';
 import * as vscode from 'vscode';
 import { Uri, window, workspace, type MessageItem } from 'vscode';
 import { findChildProcess } from '../../commands/pickFuncProcess';
-import pstree from 'ps-tree';
 import find_process from 'find-process';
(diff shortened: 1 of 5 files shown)


libs/designer-v2/src/lib/ui/FloatingRunButton/index.tsx
```diff
@@ -25,7 +25,7 @@ const RunWithPayloadIcon = bundleIcon(FlashSettingsFilled, FlashSettingsRegular)
 export interface FloatingRunButtonProps {
   siteResourceId?: string;
   workflowName?: string;
-  saveDraftWorkflow: (workflowDefinition: Workflow, customCodeData: any, onSuccess: () => void) => Promise<any>;
+  saveDraftWorkflow: (workflowDefinition: Workflow, customCodeData: any, onSuccess: () => void, isDraftSave?: boolean) => Promise<any>;
```
Eric-B-Wu, line 28: do we need to make any changes here for standalone?

takyyon: Standalone already has the changes

PR #8373, opened 2025-10-14
feat(vscode): Enable Managed Service Identity (MSI) authentication by default for Azure Logic Apps connections

samikay101:
[Work Item: Add MSI authentication support and enable by default for Logic Apps connections](https://msazure.visualstudio.com/One/_workitems/edit/35560952)
## Commit Type

## Risk Level
- [x] low - Minor changes, low user/system impact

## What & Why

This PR adds **Managed Service Identity (MSI)** authentication support for Azure Logic Apps connections and **enables it by default** for all users. 

Previously, the extension only supported key-based authentication, which required storing connection keys in application settings. With MSI, Azure services can authenticate using Azure Active Directory identities without managing secrets.

**Key Changes:**
- Added `isMSIEnabled()` helper function to check if MSI is enabled
- Modified connection creation to use MSI authentication when enabled
- Connection keys are now only stored in settings when MSI is disabled
- Changed default from `ext.useMSI = false` to `ext.useMSI = true`

## Impact of Change

- **Users**: 
  - **Breaking Change**: New installations will use MSI by default instead of key-based authentication
  - Existing users with `ext.useMSI = false` will continue using key-based auth
  - MSI provides better security by eliminating stored connection keys
  - Requires proper Azure role assignments for MSI to work

- **Developers**: 
  - New `isMSIEnabled()` function available for checking MSI status
  - Connection reference model now supports both `ManagedServiceIdentity` and `Raw` authentication types
  - Settings are only populated with connection keys when MSI is disabled

- **System**: 
  - Reduced secret storage when MSI is used
  - Telemetry updates to track MSI configuration
  - Connection key generation now conditional based on MSI status

## Test Plan

- [ ] E2E tests added/updated
- [X] Manual testing completed
- [ ] Tested in:

### DIFF
--- a/apps/vs-code-designer/src/app/utils/codeless/connection.ts
+++ b/apps/vs-code-designer/src/app/utils/codeless/connection.ts
@@ -189,7 +189,7 @@ async function getConnectionReference(
     connectionProperties,
   } = reference;
 
-  const useMsi = ext.useMSI; // Check if MSI is enabled from extension settings
+  const useMsi = ext.useMSI;
 
   return axios
     .post(
--- a/apps/vs-code-designer/src/app/commands/deploy/deploy.ts
+++ b/apps/vs-code-designer/src/app/commands/deploy/deploy.ts
@@ -26,7 +26,12 @@ import { localize } from '../../../localize';
 import { LogicAppResourceTree } from '../../tree/LogicAppResourceTree';
 import { SlotTreeItem } from '../../tree/slotsTree/SlotTreeItem';
 import { SubscriptionTreeItem } from '../../tree/subscriptionTree/subscriptionTreeItem';
-import { createAclInConnectionIfNeeded, getConnectionsJson } from '../../utils/codeless/connection';
+import {
+  createAclInConnectionIfNeeded,
+  getConnectionsJson,
+  updateAuthenticationInConnections,
+  updateAuthenticationParameters,
+} from '../../utils/codeless/connection';
 import { getParametersJson } from '../../utils/codeless/parameter';
 import { isPathEqual, writeFormattedJson } from '../../utils/fs';
 import { addLocalFuncTelemetry } from '../../utils/funcCoreTools/funcVersion';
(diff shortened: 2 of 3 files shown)


apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -516,3 +533,11 @@ async function createAccessPolicyInConnection(
       throw new Error(`Error in creating accessPolicy - ${name} for connection - ${connectionId}. ${error}`);
     });
 }
+
+/**
+ * Determines whether Managed Service Identity (MSI) authentication is enabled.
+ * @returns {boolean} True if MSI is enabled in the extension configuration, false otherwise.
+ */
+export function isMSIEnabled(): boolean {
```
andrew-eldridge: nit: the function is redundant, usages can just be replaced with ext.useMSI

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -188,25 +189,35 @@ async function getConnectionReference(
...
     .post(
       `${formatSetting(workflowBaseManagementUri)}/${formatSetting(connectionId)}/listConnectionKeys?api-version=2018-07-01-preview`,
       { validityTimeSpan: '7' },
       { headers: { authorization: accessToken } }
     )
     .then(({ data: response }) => {
-      const appSettingKey = `${referenceKey}-connectionKey`;
-      settingsToAdd[appSettingKey] = response.connectionKey;
+      // Only add connection key to settings if NOT using MSI
+      if (!useMsi) {
```
andrew-eldridge, line 202: It seems like we're always use MSI now based on ext.useMSI always being true, is there any case where we'd still be using connection keys here? If "useMSI" should be an extension setting that the user can modify it needs to be added to the configuration in package.json and retrieved with getGlobalSetting (see "autoRuntimeDependenciesValidationAndInstallation")

samikay101: Thats a great question. In a future PR we're going to add a UI configuration that will allow the user to opt in/out of MSI upon workspace creation, if that is the case, we will still need to hit this code

apps/vs-code-designer/src/app/utils/codeless/connection.ts
samikay101: fixed, thank you

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -281,6 +297,7 @@ export async function getConnectionsAndSettingsToUpdate(
       context.telemetry.properties.connectionKeyGenerated = `${referenceKey}-connectionKey generated and is valid for 7 days`;
       areKeysGenerated = true;
     } else if (
+      !isMSIEnabled &&
```
ccastrotrejo: You are not calling the function in here

apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -259,9 +270,14 @@ export async function getConnectionsAndSettingsToUpdate(
         parameterizeConnectionsSetting
       );
 
-      context.telemetry.properties.connectionKeyGenerated = `${referenceKey}-connectionKey generated and is valid for 7 days`;
-      areKeysGenerated = true;
-    } else if (isApiHubConnectionId(reference.connection.id) && !localSettings.Values[`${referenceKey}-connectionKey`]) {
+      context.telemetry.properties.connectionKeyGenerated = isMSIEnabled()
```
ccastrotrejo, line 273: lets store the value in a variable like you did above. This will. help to not call the function 4 times in this function

apps/vs-code-designer/src/app/utils/codeless/connection.ts
samikay101: fixed, just going to use ext.useMSI directly

apps/vs-code-designer/src/app/utils/codeless/connection.ts
samikay101: fixed, just going to use ext.useMSI directly

PR #8381, opened 2025-10-15
feat: Add handoffs support for Consumption workflows

Bhavd13:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why
Enabled the handoff panel for Consumption workflows to support Agent-to-Agent (A2A) conversational scenarios. Previously, the handoff functionality was only available for Standard Logic Apps workflows with `workflowKind: 'agent'`. This change extends support to Consumption workflows that use Agent Request triggers.

The implementation adds detection logic to identify Consumption A2A workflows by checking for:
1. Request-type triggers with the `agent` kind

## Impact of Change

- Users : Consumption workflow users creating conversational/A2A scenarios can now access and configure handoffs through the UI panel, improving the authoring experience for agent-based workflows

- Developers : Updated `useIsA2AWorkflow` selector to support both Standard and Consumption workflow detection patterns. No breaking changes to existing APIs

- System : Minimal performance impact - added operation type checking logic runs only when determining panel visibility

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Consumption workflows with Agent Request triggers and Agent operations
- [x] Verified handoff panel appears for A2A Consumption workflows
- [x] Verified existing Standard A2A workflow functionality remains unchanged
- [x] Tested panel visibility logic with various workflow configurations

## Contributors
@Bhavd13, @rllyy97

## Screenshots/Videos
image

### DIFF
--- a/libs/designer-ui/src/lib/card/subgraphCard/subgraphCard.less
+++ b/libs/designer-ui/src/lib/card/subgraphCard/subgraphCard.less
@@ -29,6 +29,9 @@
     flex-direction: row;
     display: flex;
     gap: 10px;
+    align-items: center;
+    width: 100%;
+    min-width: 0;
   }
 
   .msla-subgraph-title {
@@ -47,13 +50,22 @@
     background-color: var(--brand-color, black);
   }
 
+  .msla-subgraph-title-icon {
+    width: 15px;
+    height: 15px;
+    flex-shrink: 0;
+  }
+
   .msla-subgraph-title-text {
     display: flex;
     font: 14px/20px normal @semibold-font-family;
-    .msla-subgraph-title-icon {
-      margin-right: 4px;
-      align-self: baseline;
-      margin-top: 3px;
+    flex: 1;
+    min-width: 0;
+
+    span {
+      word-wrap: break-word;
+      overflow-wrap: break-word;
+      word-break: break-word;
     }
   }
   .msla-panel-card-error-wrapper {


libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
```diff
@@ -68,7 +68,33 @@ export const useIsAgenticWorkflowOnly = () => {
...
+
+    const triggerNodeId = Object.keys(nodesMetadata).find((nodeId) => nodesMetadata[nodeId]?.isTrigger === true);
+
+    let isConsumptionA2A = false;
+    if (triggerNodeId) {
+      const triggerOperation = operations[triggerNodeId];
+
+      if (triggerOperation) {
+        const isRequestType = equals(triggerOperation.type, 'Request', true);
+        const hasAgentOperations = Object.values(operations).some((operation) => equals(operation.type, 'Agent', true));
```
ishank12: Is this required?
For A2A trigger we need to check triggerOperation.kind = Agent right ?

libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
```diff
@@ -68,7 +68,33 @@ export const useIsAgenticWorkflowOnly = () => {
...
+        const hasA2ANamePattern =
+          triggerNodeId.toLowerCase().includes('chat') ||
```
ishank12: this name pattern can be changed by user right ?
we dont need this check right ?

libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
Bhavd13: I initially tried checking `triggerOperation.kind === 'Agent'`, but during testing, the `kind` property was coming back as `undefined` in the Redux state, even though it exists in the workflow definition JSON.

libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
Bhavd13: The trigger display name can be changed by users in the UI, but the trigger node ID cannot be. This check is placed in order to distinguish http trigger from a2a because both have `triggerOperation.type = Request` So if I add agent action to http trigger it will also show handoffs panel which is not the intended behaviour. Open to suggestions if there is any other way to distinguish.

__mocks__/workflows/HandoffConversationalConsumption.json
```diff
@@ -0,0 +1,182 @@
...
+          "undeletable": true
+        }
+      }
+    },
+    "actions": {
+      "TriageAgent": {
+        "type": "Agent",
+        "inputs": {
+          "parameters": {
+            "deploymentId": "Gpt-4o",
```
ishank12: modelId

__mocks__/workflows/HandoffConversationalConsumption.json
```diff
@@ -0,0 +1,182 @@
...
+              {
+                "role": "System",
+                "content": "You are a triage agent for ACME Inc who is responsible for deciding which agent should handle a given user message. Invoke the right hand-off tool based on the decision."
+              }
+            ],
+            "agentModelSettings": {
+              "agentChatCompletionSettings": {
+                "maxTokens": 3000,
+                "frequencyPenalty": 0.1,
+                "presencePenalty": 0.1,
```
ishank12: are all these valid for consumption

__mocks__/workflows/HandoffConversationalConsumption.json
```diff
@@ -0,0 +1,182 @@
...
+                "temperature": 0.1,
+                "topP": 0.1
+              },
+              "agentHistoryReductionSettings": {
+                "agentHistoryReductionType": "messageCountReduction",
+                "messageCountLimit": 50
+              }
+            }
+          },
+          "modelConfigurations": {
```
ishank12: modelConfigurations required ?

libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
```diff
@@ -68,7 +68,33 @@ export const useIsAgenticWorkflowOnly = () => {
 };
 
 export const useIsA2AWorkflow = () => {
-  return useSelector((state: RootState) => equals(state.workflow.workflowKind, 'agent', false));
+  return useSelector((state: RootState) => {
+    const isStandardA2A = equals(state.workflow.workflowKind, 'agent', false);
```
rllyy97: Let's return early after this line if `isStandardA2A` is true, just to avoid all the extra logic running

libs/designer/src/lib/core/state/designerView/designerViewSelectors.ts
rllyy97: Trigger ID can be changed by the user, we should not check over the trigger / node ID.
We should check over the kind being `Agent`, if you're not seeing it in redux something else must not be working.
Should be `state.workflow.operations[triggerNodeId].kind`

rllyy97 (approved):

PR #8412, opened 2025-10-20
feat(designer): Support serialization/deserialization of MCP connection

anthony0t:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Adds support for a new connection type (MCP connection) in connections.json, enabling serialization and deserialization in the designer. No customer impact is expected at this stage, but this enables future MCP Client tool integration. Changes are isolated to minimize regression risk.

## Impact of Change

- **Users**: Ussers will be able to add MCP connections
- **Developers**: During deserialization, the McpConnections if present, will be deserialized and saved as connection in store. During serialization, those connections will be serialized back as well.
- **System**: None

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:

### DIFF
--- a/libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
+++ b/libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
@@ -365,6 +365,13 @@ export const getConnectionMappingForNode = (
         return mapping;
       }
     }
+    if (equals(operation.type, 'McpClientTool') && equals(operation.kind, 'Managed')) {
+      const connectionReferenceKey = (operation as any).inputs.connectionReference.connectionName;
+      if (connectionReferenceKey !== undefined) {
+        const mapping = Promise.resolve({ [nodeId]: connectionReferenceKey });
+        return mapping;
+      }
+    }
     return Promise.resolve(undefined);
   } catch (error) {
     const errorMessage = `Failed to get connection mapping for node: ${error}`;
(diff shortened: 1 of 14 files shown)


libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -365,6 +365,13 @@ export const getConnectionMappingForNode = (
         return mapping;
       }
     }
+    if (equals(operation.type, 'McpClientTool') && equals(operation.kind, 'Managed')) {
```
takyyon, line 368: does this have to be case-sensitive?

libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -365,6 +365,13 @@ export const getConnectionMappingForNode = (
...
+      const connectionReferenceKey = (operation as any).inputs.connectionReference.connectionName;
```
takyyon, line 369: inputs and input.conntionReference always exists?

anthony0t: Yes it should always exist.

libs/designer/src/lib/core/actions/bjsworkflow/connections.ts
anthony0t: In theory, yes they are case insensitive. I will take a look at it to see if we tried to normalize it to lower case and we can skip the case insensitive compare here in that case.

anthony0t: I checked and indeed, we need to do case insensitive compare here. Not sure if it's a bug on other places where it does not do case insensitive compare. I'll do some investigation separately, and close this for now.

PR #8417, opened 2025-10-21
feat(designer): core business logic for serializaton/deserialization/add/delete of MCP actions

anthony0t:
core business logic for serializaton/deserialization/add/delete of MCP actions

## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

This change added the serialization/deserialization/add/delete logic for MCP actions.

## Impact of Change

- **Users**: 
User can add/delete mcp actions
- **Developers**: 
Added two new method to add/delete mcp actions. Also added support for serialization/deserialization of mcp actions.
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:


libs/designer-v2/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
```diff
@@ -359,6 +359,14 @@ const isIfAction = (action: LogicAppsV2.ActionDefinition): action is LogicAppsV2
   return equals(action?.type, 'if');
 };
 
+const isAgentCondition = (action: LogicAppsV2.AgentCondition | LogicAppsV2.McpClient): action is LogicAppsV2.AgentCondition => {
+  return !action?.type || !equals(action?.type, 'mcpclienttool');
+}
+
+const isMcpClient = (action: LogicAppsV2.AgentCondition | LogicAppsV2.McpClient): action is LogicAppsV2.McpClient => {
```
takyyon, line 366: Left this comment in the other PR too. Do we need two checks here? Can we not just use `if(isMcpClient()) else {}`

libs/logic-apps-shared/src/designer-client-services/lib/standard/operationmanifest.ts
```diff
@@ -20,6 +22,19 @@ export class StandardOperationManifestService extends BaseOperationManifestServi
...
     }, {});
+
+    this.allBuiltInConnectors[mcpclientconnector.id.toLowerCase()] = mcpclientconnector;
+  }
+
+  override isSupported(operationType?: string, operationKind?: string): boolean {
+    const normalizedOperationType = operationType?.toLowerCase() ?? '';
+    const normalizedOperationKind = operationKind?.toLowerCase() ?? '';
+
+    if (normalizedOperationType === "mcpclienttool" && normalizedOperationKind !== "managed") {
```
takyyon: nit: For case in-sensitive check, I would suggest using "equals()" across the PR

takyyon (approved):

libs/designer-v2/src/lib/core/parsers/BJSWorkflow/BJSDeserializer.ts
anthony0t: We can call other method inside here, but we would need to methods because they are type guards

PR #8450, opened 2025-10-24
fix(designerv2): Disable run for specific trigger types in Draft mode

takyyon:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
Run is only supported for specific trigger types in draft mode since BE isn't always listening on the trigger.

## Impact of Change

- **Users**: Users can use Run With Payload for all triggers and Run for specific ones
- **Developers**: None
- **System**: None

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:

### DIFF
--- a/apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
+++ b/apps/vs-code-designer/src/app/commands/workflows/exportLogicApp.ts
@@ -25,7 +25,7 @@ import { writeFileSync } from 'fs';
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as AdmZip from 'adm-zip';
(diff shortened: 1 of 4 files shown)


libs/designer-v2/src/lib/ui/FloatingRunButton/index.tsx
```diff
@@ -54,6 +55,34 @@ export const FloatingRunButton = ({
 
   const operationState = useSelector((state: RootState) => state.operations);
 
+  // Trigger types that are allowed to run in DraftMode
+  const allowedTriggerTypes = useMemo(
```
ccastrotrejo: NIT: if this is a constant value can you put it outside the component and avoid having an extra useMemo()

PR #8458, opened 2025-10-26
refactor(vscode): Centralize internationalization messages in vs-code-react

ccastrotrejo:
## Commit Type

## Risk Level

## What & Why
This PR implements a centralized internationalization (i18n) system for the vs-code-react application to eliminate duplicate message definitions and improve maintainability. The changes introduce a custom hook pattern (`useIntlMessages`) that works with centralized message definitions, replacing the pattern of defining i18n messages inline within components.

## Impact of Change
- **Users**: No user-facing changes - all existing text and translations remain the same
- **Developers**: New pattern for defining and using internationalized messages through centralized message definitions and `useIntlMessages` hook. Reduces code duplication and improves type safety
- **System**: Improved maintainability and reduced bundle size by eliminating duplicate message definitions across components

## Test Plan
- [x] Tested in: Local development environment, verified all components render correctly with new i18n pattern

## Key Changes
- Created centralized message repository (`apps/vs-code-react/src/intl/messages.ts`) with organized message definitions
- Implemented custom `useIntlMessages` hook for formatting messages with type safety
- Migrated 24+ components from inline `intl.formatMessage()` calls to the centralized pattern
- Updated `IntlProvider` configuration to support the new pattern
- Fixed minor whitespace and message ID issues

## Contributors
@ccastrotrejo

## Screenshots/Videos
N/A - No UI changes, internal refactoring only


apps/vs-code-react/src/app/createWorkspace/createWorkspace.tsx
```diff
@@ -181,7 +183,7 @@ export const CreateWorkspace: React.FC = () => {
     });
   };
 
-  const intlText = {
+  const intlMessages = {
```
andrew-eldridge, line 121: nit: it looks like there's inconsistency between files with using intlText vs intlMessages

apps/vs-code-react/src/app/export/status/status.tsx
```diff
@@ -47,10 +40,10 @@ interface FinalStatusGadgetProps {
 }
 
 const FinalStatusGadget: React.FC<FinalStatusGadgetProps> = ({ finalStatus }) => {
-  const intl = useIntl();
   const workflowState = useSelector((state: RootState) => state.workflow);
   const { targetDirectory } = workflowState.exportData;
   const styles = useExportStyles();
+  const intl = useIntl();
```
andrew-eldridge, line 46: do we still need useIntl here instead of useIntlMessages?

apps/vs-code-react/src/app/createWorkspace/createWorkspace.tsx
ccastrotrejo: ohh let me check that and make a single pattern on these. Thanks!

apps/vs-code-react/src/app/export/status/status.tsx
ccastrotrejo: We still use it for the message that is below, which sends a parameter to the localized message. Let me research on how we can apply that to the list of messages. Otherwise for those types of message we are going to keep this implementation

PR #8460, opened 2025-10-27
fix(vscode): Limit .net framework to windows OS 

ccastrotrejo:
## Commit Type

## Risk Level

## What & Why
This PR refactors the `Platform` type definition from being local to the VS Code designer app into a shared library model (`@microsoft/vscode-extension-logic-apps`). This enables the Platform type to be used across multiple packages in the monorepo, particularly allowing the React-based webview to access platform information for conditional UI rendering (e.g., showing .NET Framework option only on Windows).

## Impact of Change
- **Users**: No user-facing changes. The functionality remains the same.
- **Developers**: The `Platform` type is now imported from `@microsoft/vscode-extension-logic-apps` instead of local constants. All import statements across 17+ files have been updated to reference the new shared location.
- **System**: No performance or architectural impact. This is a pure refactor that improves code organization and reusability.

## Test Plan
- [x] Tested in: Local development environment
- Verified that all imports resolve correctly
- Confirmed that the VS Code extension builds successfully
- Tested that platform-dependent UI logic works as expected (Windows shows .NET Framework option)

## Contributors
@ccastrotrejo

## Screenshots/Videos

CleanShot 2025-10-27 at 14 18 49@2x


apps/vs-code-react/src/app/createWorkspace/steps/dotNetFrameworkStep.tsx
```diff
@@ -210,9 +210,11 @@ export const DotNetFrameworkStep: React.FC = () => {
               placeholder={intlText.NET_VERSION_PLACEHOLDER}
               className={styles.inputControl}
             >
-              <Option value="net472" text=".NET Framework">
-                .NET Framework
-              </Option>
+              {platform === Platform.windows ? (
+                <Option value="net472" text=".NET Framework">
```
andrew-eldridge, line 214: Does this mean the only option is .net 8 for non-windows? If so does it make sense to remove this step/option entirely

ccastrotrejo: Yes, thats right thats the only option. 
Yep, we can remove it later. I think for now I will keep it and just limit it an in another PR we can remove it

PR #8575, opened 2025-11-13
fix(vscode): Harden connection key expiration logic to prevent startup crashes on invalid or non-JWT keys

samikay101:
[Work Item: VS Code fails to load when connection key is invalid or non-JWT in local.settings.json](https://msazure.visualstudio.com/One/_workitems/edit/35901332)

## Commit Type

* [x] fix - Bug fix
* [ ] feature - New functionality
* [ ] refactor - Code restructuring without behavior change
* [ ] perf - Performance improvement
* [ ] docs - Documentation update
* [ ] test - Test-related changes
* [ ] chore - Maintenance/tooling

## Risk Level

* [x] Low - Minor changes, limited scope
* [ ] Medium - Moderate changes, some user impact
* [ ] High - Major changes, significant user/system impact

## What & Why

This change hardens the `isKeyExpired` logic so that malformed, empty, or parameterized connection keys no longer cause a designer startup crash.

Previously, the method assumed that all connection keys in `local.settings.json` were valid JWTs. In cases where the key was empty or incorrectly parameterized (e.g., `@parameters('...')`), `jwtTokenHelper.extractJwtTokenPayload` threw an exception, breaking the entire connection-loading pipeline.

**Fix:**

* Wrapped the JWT parsing logic in `try/catch`.
* If parsing fails, telemetry is logged.
* Invalid keys are treated as “expired,” triggering automatic regeneration instead of crashing.

This preserves existing behavior for valid keys while recovering gracefully from bad or unexpected values.

## Impact of Change

* **Users**: Prevents designer startup failures when connection keys are invalid. Improves reliability when switching connections or reopening projects.
* **Developers**: Telemetry now surfaces invalid connection key formats, improving diagnosability of connection issues.
* **System**: No behavioral regressions; invalid keys are safely regenerated.

## Test Plan

* [x] Manual testing completed:

  * Verified startup with valid JWT key
  * Verified startup with empty key
  * Verified startup with parameterized key (`@parameters(...)`)
  * Verified expired key regeneration path still works
* [ ] Unit tests added/updated
* [ ] E2E tests added/updated
* [ ] Tested in: local VS Code environment


apps/vs-code-designer/src/app/utils/codeless/connection.ts
```diff
@@ -155,13 +155,46 @@ async function addConnectionDataInJson(
...
+    const payload: Record<string, any> = jwtTokenHelper.extractJwtTokenPayload(connectionKey);
+    const secondsSinceEpoch = date / 1000;
+    const buffer = bufferInHours * 3600;
+    const expiry = payload[JwtTokenConstants.expiry];
 
-  return expiry - buffer <= secondsSinceEpoch;
+    return expiry - buffer <= secondsSinceEpoch;
+  } catch (error) {
+    // safe fallback for invalid/malformed keys
+    const message = error instanceof Error ? error.message : String(error);
```
andrew-eldridge, line 187: It's good to have the error check here, but does this also address the root cause of why parameters references were showing up in local settings in the first place? This should never happen

samikay101: Here’s a clearer explanation of what I found:

I wasn't able to repro the exact scenario bug hit where a **parameter reference** ended up inside `local.settings.json`, mostly because she was using an older released VSIX and a lot has been checked into `main` since then. That bug may already have been fixed upstream.

However, I followed her scenario step-by-step:

1. I added a connection and confirmed the connection key was generated correctly in `connections.json`.
2. Then I manually made the key invalid (either empty or a non-JWT string).
3. I closed the workspace and reopened it.

At this point, the VS Code extension is supposed to detect that the connection key is invalid and regenerate a fresh key on startup.
But that didn’t happen.

Instead, when I tried adding another connection of the same type:

* I was unable to save the workflow
* The extension threw an error saying the connection key could not be parsed
* The designer effectively got into a broken state

So while I couldn’t repro Nidhi’s **exact** bug involving `@parameters(...)` inside local settings, I was able to repro the **same underlying failure mode**:

> The extension does **not** properly validate or recover from an invalid connection key on startup.

To fix this, I added logic that:

* Checks whether the connection key is valid on **workspace open**, and
* Checks again when **saving** the workflow

If the key is invalid in either case, we now treat it as expired and regenerate a valid one instead of crashing.

So the root issue seems to be that the extension was missing validation for invalid connection keys in multiple code paths. The parameterization bug may be gone now, but the invalid-key scenario was still causing the same crash.

PR #8576, opened 2025-11-14
fix(vscode): Fix dynamic connector failures by passing OBO token into agent iframe and correcting dynamic connection serialization

samikay101:
[Work Item](https://msazure.visualstudio.com/One/_workitems/edit/35915402)

## Commit Type

* [x] fix - Bug fix
* [ ] feature - New functionality
* [ ] refactor - Code restructuring without behavior change
* [ ] perf - Performance improvement
* [ ] docs - Documentation update
* [ ] test - Test-related changes
* [ ] chore - Maintenance/tooling

## Risk Level

* [ ] Low - Minor changes, limited scope
* [x] Medium - Moderate changes, some user impact
* [ ] High - Major changes, significant user/system impact

## What & Why

Dynamic connections were effectively broken when using the agent because we were **not passing the OBO user token into the chat/agent iframe**. As a result, the agent could not act on behalf of the signed-in user and all dynamic connector scenarios failed.

This change fixes that by:

* Adding `oboUserToken` to the `AgentQueryParams` model and wiring it into the iframe URL used by `Chat.tsx`.
* Ensuring we persist a default MSI audience (`Workflows.DynamicConnection.DefaultAuthenticationAudience`) into local settings when the workflow authentication method is set, so dynamic connections have a sensible default audience.
* Adjusting `getConnectionReference` to:

  * Detect **dynamic** connections via `connectionProperties.runtimeSource === 'Dynamic'`.
  * Flatten `runtimeSource` and `dynamicConnectionProxyUrl` to the top level of the connection reference for dynamic connections.
  * Keep `connectionProperties` for non-dynamic connections (with `connectionRuntimeUrl` removed to avoid duplicate runtime URLs).
* Plumb `connectionData` from the VS Code extension (`openOverview`) into the React workflow state so the UI has access to connection metadata when working with dynamic connectors.

Together, these changes unblock agent-based dynamic connector usage and align the connection shape and auth data with what the runtime expects.

## Impact of Change

* **Users**:

  * Dynamic connectors invoked through the agent now work as expected because the agent receives the OBO token.
  * No UX changes in the designer, but previously failing dynamic connector scenarios should now succeed.

* **Developers**:

  * New constants in `constants.ts`:

    * `workflowsDynamicConnectionDefaultAuthAudienceKey`
    * `defaultMsiAudience`
  * `AgentQueryParams` now includes an optional `oboUserToken` field.
  * Workflow React state (`WorkflowSlice`) includes a new `connectionData` property that is initialized from the extension’s `openOverview` command.
  * `getConnectionReference` now treats dynamic vs non-dynamic connections differently:

    * Dynamic: flattens `runtimeSource` and `dynamicConnectionProxyUrl` to the top level.
    * Non-dynamic: retains `connectionProperties` (minus `connectionRuntimeUrl`).

* **System**:

  * Behavior change for **dynamic** connections in `connections.json` (flattened properties instead of nested under `connectionProperties`), which better matches how the runtime and designer expect these values.
  * Default MSI audience is persisted to local settings, providing a consistent value for token acquisition.
  * No expected performance impact; logic is limited to connection shaping and URL construction.

## Test Plan

* [ ] Unit tests added/updated
* [ ] E2E tests added/updated
* [x] Manual testing completed
* [ ] Tested in: 

**Manual testing:**

* Local VS Code workspace with a Logic Apps Standard project:

  * Configure MSI / dynamic connection.
  * Open overview and launch the chat/agent.
  * Verify the chat iframe URL now includes `oboUserToken` when available.
  * Execute operations via the agent that require the dynamic connector and confirm they succeed where they previously failed.
* Inspect generated `connections.json`:

  * Confirm dynamic connections include `runtimeSource` and `dynamicConnectionProxyUrl` at the top level.
  * Confirm non-dynamic connections still retain `connectionProperties` (without duplicate `connectionRuntimeUrl`).

### DIFF
--- a/apps/vs-code-react/src/app/overview/services/workflowService.ts
+++ b/apps/vs-code-react/src/app/overview/services/workflowService.ts
@@ -7,6 +7,7 @@ import {
   type IHttpClient,
   type ConnectionsData,
   equals,
+  resolveConnectionsReferences,
 } from '@microsoft/logic-apps-shared';
 
 export const fetchAgentUrl = (
@@ -112,24 +113,44 @@ const fetchOBOData = async (
 
     for (const key of Object.keys(apiConnections)) {
       const connection = apiConnections[key];
+
       // Check for runtimeSource at the root level (new structure after your changes)
       const runtimeSourceAtRoot = connection.runtimeSource;
       // Also check in connectionProperties for backward compatibility
       const runtimeSourceInProperties = connection.connectionProperties?.runtimeSource;
 
       // Check both locations for Dynamic runtime source
-      if (equals(runtimeSourceAtRoot ?? '', 'Dynamic', true) || equals(runtimeSourceInProperties ?? '', 'Dynamic', true)) {
+      const isDynamic = equals(runtimeSourceAtRoot ?? '', 'Dynamic', true) || equals(runtimeSourceInProperties ?? '', 'Dynamic', true);
+
+      if (isDynamic) {
         connectionId = connection.connection.id;
-        connectionId = resolveConnectionIdPlaceholders(connectionId, azureSubscriptionId, resourceGroup);
         break;
       }
     }
 
     if (!connectionId) {
+      // No dynamic connection found – OBO is optional, so we can return null
       return null;
     }
 
-    const oboKeysUri = `${armBaseUrl}${connectionId}/listDynamicConnectionKeys?api-version=2015-08-01-preview`;
+    const appsettings: Record<string, string> = {
+      WORKFLOWS_SUBSCRIPTION_ID: azureSubscriptionId,
+      WORKFLOWS_RESOURCE_GROUP_NAME: resourceGroup,
+    };
+
+    // Wrap the connectionId in a JSON payload so resolveConnectionsReferences
+    // can safely JSON.parse after doing text replacement.
+    const connectionIdPayload = JSON.stringify({ connectionId });
+
+    const resolvedPayload = resolveConnectionsReferences(
+      connectionIdPayload,
+      /* parameters */ undefined,
+      /* appsettings */ appsettings
+    ) as { connectionId?: string };
+
+    const resolvedConnectionId = resolvedPayload?.connectionId ?? connectionId;
+
+    const oboKeysUri = `${armBaseUrl}${resolvedConnectionId}/listDynamicConnectionKeys?api-version=2015-08-01-preview`;
 
     const oboResponse = await httpClient.post<any, any>({
       uri: oboKeysUri,
@@ -150,25 +171,4 @@ const fetchOBOData = async (
     });
     return null;
   }
-};
-
-// Helper function to resolve app settings placeholders in connection IDs
-const resolveConnectionIdPlaceholders = (connectionId: string, subscriptionId: string, resourceGroup: string): string => {
-  let resolvedConnectionId = connectionId;
-
-  // Replace subscription ID placeholder - handle both encoded and non-encoded formats
-  // Non-encoded format: @{appsetting('WORKFLOWS_SUBSCRIPTION_ID')}
-  resolvedConnectionId = resolvedConnectionId.replace(/@\{appsetting\('WORKFLOWS_SUBSCRIPTION_ID'\)\}/g, subscriptionId);
-
-  // URL-encoded format (just in case): @%7Bappsetting('WORKFLOWS_SUBSCRIPTION_ID')%7D
-  resolvedConnectionId = resolvedConnectionId.replace(/@%7Bappsetting\('WORKFLOWS_SUBSCRIPTION_ID'\)%7D/g, subscriptionId);
-
-  // Replace resource group placeholder - handle both encoded and non-encoded formats
-  // Non-encoded format: @{appsetting('WORKFLOWS_RESOURCE_GROUP_NAME')}
-  resolvedConnectionId = resolvedConnectionId.replace(/@\{appsetting\('WORKFLOWS_RESOURCE_GROUP_NAME'\)\}/g, resourceGroup);
-
-  // URL-encoded format (just in case): @%7Bappsetting('WORKFLOWS_RESOURCE_GROUP_NAME')%7D
-  resolvedConnectionId = resolvedConnectionId.replace(/@%7Bappsetting\('WORKFLOWS_RESOURCE_GROUP_NAME'\)%7D/g, resourceGroup);
-
-  return resolvedConnectionId;
 };


apps/vs-code-designer/src/app/commands/workflows/openOverview.ts
```diff
@@ -283,3 +287,17 @@ function getWorkflowStateType(workflowName: string, kind: string, settings: Reco
         ? localize('logicapps.statelessDebug', 'Stateless (debug mode)')
         : localize('logicapps.stateless', 'Stateless');
 }
+
+async function getConnectionData(projectPath: string): Promise<Record<string, any>> {
```
ccastrotrejo: NIT: There is a getConnectionsJson() function that does the same thing, it only returns it as a string. You only will need to do the `JSON.parse(connectionContent)`

apps/vs-code-react/src/app/overview/services/workflowService.ts
```diff
@@ -20,29 +31,67 @@ export const fetchAgentUrl = (
...
-        queryParams = { apiKey: a2aKey };
+      if (a2aKey || oboToken) {
+        queryParams = {
+          ...(a2aKey && { apiKey: a2aKey }),
+          ...(oboToken && { oboUserToken: oboToken }),
+        };
       }
 
-      return {
+      const result = {
```
andrew-eldridge: nit- unneeded var, just return

apps/vs-code-react/src/app/overview/services/workflowService.ts
```diff
@@ -59,3 +108,84 @@ const fetchA2AAuthKey = async (workflowName: string, baseUrl: string, httpClient
...
+// Helper function to fetch OBO (On-Behalf-Of) data
+const fetchOBOData = async (
+  siteResourceId: string,
+  baseUrl: string,
+  httpClient: IHttpClient,
+  clientId: string,
+  tenantId: string,
+  azureSubscriptionId: string,
+  resourceGroup: string,
+  siteName: string,
```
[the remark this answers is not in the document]
andrew-eldridge: please remove the unused params here

apps/vs-code-react/src/app/overview/services/workflowService.ts
```diff
@@ -59,3 +108,84 @@ const fetchA2AAuthKey = async (workflowName: string, baseUrl: string, httpClient
...
+) => {
+  try {
+    const armBaseUrl = 'https://management.azure.com';
+
+    // Find dynamic connection
+    const apiConnections = connectionsData?.managedApiConnections ?? {};
+    let connectionId = '';
+
+    for (const key of Object.keys(apiConnections)) {
+      const connection = apiConnections[key];
```
andrew-eldridge, line 115: This looks like it gets the connectionId for a random connection with runtimeSource=='Dynamic', does it not matter which connection is used?

apps/vs-code-react/src/app/overview/services/workflowService.ts
```diff
@@ -59,3 +108,84 @@ const fetchA2AAuthKey = async (workflowName: string, baseUrl: string, httpClient
...
+      level: LogEntryLevel.Error,
+      message: `Failed to get OBO data: ${error}`,
+      area: 'fetchAgentUrl',
+    });
+    return null;
+  }
+};
+
+// Helper function to resolve app settings placeholders in connection IDs
+const resolveConnectionIdPlaceholders = (connectionId: string, subscriptionId: string, resourceGroup: string): string => {
```
andrew-eldridge: Do we already have a util for resolving parameters like this? Seems like it shouldn't be handled in workflowService. @ccastrotrejo

libs/designer-ui/src/lib/overview/chat.tsx
```diff
@@ -66,11 +66,13 @@ export const ChatButton: React.FC<ChatButtonProps> = ({ loading, data, isWorkflo
     if (loading) {
       return <Spinner size="medium" label={IntlText.LOADING} />;
     }
+    const chatUrl = `${data?.chatUrl}?apiKey=${data?.queryParams?.apiKey}${data?.queryParams?.oboUserToken ? `&oboUserToken=${data?.queryParams?.oboUserToken}` : ''}${buttonCommonProps.isDarkMode ? '&mode=dark' : ''}`;
```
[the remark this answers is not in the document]
andrew-eldridge: I think copilot is right- string should be escaped here

apps/vs-code-react/src/app/overview/app.tsx
```diff
@@ -129,7 +131,16 @@ export const OverviewApp = () => {
     ['agentUrl', isWorkflowRuntimeRunning, baseUrl],
     async () => {
       invariant(!!httpClient, 'Agent URL should not be retrieved unless httpClient is available');
-      return fetchAgentUrl(workflowProperties.name, baseUrl, httpClient, clientId, tenantId);
+      return fetchAgentUrl(
```
andrew-eldridge, line 134: Important: fetchAgentUrl needs to be updated in the workflowService used by designer as well, not just overview

apps/vs-code-designer/src/app/commands/workflows/openOverview.ts
```diff
@@ -283,3 +287,17 @@ function getWorkflowStateType(workflowName: string, kind: string, settings: Reco
...
+    const connectionFilePath = join(projectPath, 'connections.json');
+    if (existsSync(connectionFilePath)) {
+      const connectionContent = readFileSync(connectionFilePath, 'utf8');
+      return JSON.parse(connectionContent);
+    }
+  } catch (error) {
+    // Log error but don't throw to avoid breaking the overview
+    console.warn('Failed to read connection.json:', error);
+  }
+  return {};
```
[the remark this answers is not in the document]
andrew-eldridge: Please handle this scenario and fallback to passing undefined connections if needed

andrew-eldridge (changes requested):

apps/vs-code-designer/src/app/commands/workflows/openOverview.ts
samikay101: fixed:     if (projectPath) {
      azureDetails = await getAzureConnectorDetailsForLocalProject(context, projectPath);
      const connectionJson = await getConnectionsJson(projectPath);
      connectionData = connectionJson ? JSON.parse(connectionJson) : {};
    }

apps/vs-code-designer/src/app/commands/workflows/openOverview.ts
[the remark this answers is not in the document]
samikay101: removed function and using an existing function that contains fallback and validation logic

apps/vs-code-react/src/app/overview/services/workflowService.ts
samikay101: fixed

apps/vs-code-react/src/app/overview/services/workflowService.ts
samikay101: no it does not, as long as it contains one dynamic connection minimum

apps/vs-code-react/src/app/overview/app.tsx
samikay101: lets do this in a future PR due to time constraints

apps/vs-code-react/src/app/overview/services/workflowService.ts
ccastrotrejo: Yes, thats right, its called resolveConnectionsReferences. You can use it @samikay101

PR #8648, opened 2025-12-08
fix(designer): Enable nesting Agent Loop nodes in FOREACH/UNTIL control flows in designer

hartra344:
## Commit Type

## Risk Level

## What & Why
Previously, Agent Loop nodes could only be added at the root level of agentic/A2A workflows. This change removes that restriction with a version gate:

**Changes:**
1. **Search panel filtering** (`searchView.tsx`): Agent operations now check `enableNestedAgentLoops` flag when not at root
2. **Context menu** (`EdgeContextualMenu.tsx`): "Add an agent" button visibility now checks the flag when not at root
3. **Version gating**: Added `enableNestedAgentLoops` hostOption that requires bundle version >= 1.115.0

**Files modified:**
- `libs/designer/src/lib/core/state/designerOptions/designerOptionsInterfaces.ts` - Added new hostOption
- `libs/designer/src/lib/core/state/designerOptions/designerOptionsSelectors.ts` - Added selector hook
- `libs/designer/src/lib/ui/panel/recommendation/searchView.tsx` - Check flag in filter
- `libs/designer/src/lib/ui/common/EdgeContextualMenu/EdgeContextualMenu.tsx` - Check flag for button visibility
- Same changes in `libs/designer-v2/`
- `apps/vs-code-react/src/app/designer/app.tsx` & `appV2.tsx` - Version check using `isVersionSupported`

## Impact of Change
- **Users**: Users with bundle version >= 1.115.0 can add Agent Loop nodes inside FOREACH/UNTIL loops in agentic workflows
- **Developers**: New `enableNestedAgentLoops` hostOption available
- **System**: No performance or architecture impact

## Test Plan
- [ ] E2E tests added/updated
- [x] Tested in: Standalone designer - verified Agent Loop appears in both operation search and context menu when inside a FOREACH loop (with flag enabled)

### DIFF
--- a/libs/logic-apps-shared/src/utils/src/lib/helpers/version.ts
+++ b/libs/logic-apps-shared/src/utils/src/lib/helpers/version.ts
@@ -6,8 +6,8 @@ import { ArgumentException } from '../exception';
 export const BundleVersionRequirements = {
   /** Minimum version for multi-variable support in Initialize Variable action */
   MULTI_VARIABLE: '1.114.22',
-  /** Minimum version for nested agent loops inside regular loops */
-  NESTED_AGENT_LOOPS: '1.115.0',
+  /** Minimum version for nested agent loops inside regular loops (TODO: update when actual version is known) */
+  NESTED_AGENT_LOOPS: '2.0.0',
 } as const;
 
 /**


apps/vs-code-react/src/app/designer/app.tsx
```diff
@@ -144,7 +152,12 @@ export const DesignerApp = () => {
   }, [connectionData]);
 
   const isMultiVariableSupportEnabled = useMemo(
-    () => isVersionSupported(panelMetaData?.extensionBundleVersion ?? '', '1.114.22'),
+    () => isVersionSupported(panelMetaData?.extensionBundleVersion ?? '', BundleVersionRequirements.MULTI_VARIABLE),
```
takyyon, line 155: Side question - Do we still need this check? I am guessing almost very logic app should already be on a higher version

Eric-B-Wu: I think the issue is users can pin backend versions which means that it could be that we'll always need to have this check in

hartra344: yeah I don't think it makes sense to remove this check in most cases since users can pin

PR #8673, opened 2025-12-17
fix(designer): Custom Connectors to be fetched by operation group instead of filtering from every custom operation

Eric-B-Wu:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

Previously we would fetch all the custom operations and filter the operations for a specific connector. Moving the logic so that we fetch based on the connector itself. Makes it so if users have many custom connector operations, they don't need to wait for the fetch of all custom operations to finish before seeing their connector's operations

## Impact of Change

- **Users**: Browse experience for custom connectors should now be improved given that operations load without having to finish the fetch of all operations
- **Developers**: Slight change in getOperationsByConnector to support custom connectors
- **System**: No System changes

## Test Plan

- [x] E2E tests added/updated
- [ ] Tested in: 

## Contributors

@Eric-B-Wu

### DIFF
--- a/e2e/designer/real-api/browse/browseCustomConnectorOperations.spec.ts
+++ b/e2e/designer/real-api/browse/browseCustomConnectorOperations.spec.ts
@@ -1,0 +1,129 @@
+import { test, expect } from '../../fixtures/real-api';
+import workflow from './workflow.json' assert { type: 'json' };
+
+test.describe(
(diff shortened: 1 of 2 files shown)


hartra344 (commented):
Can we please have some Unit tests added?

PR #8691, opened 2026-01-08
feat(vscode): add "Deploy to Azure" webview and deploy flow (create Logic App without wizard)

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

This moves the deploy command from using vscode wizards to a custom webview which allows more control and validation from the extension. We also display all relevant information so that users don't have to remember what they had input previously.

## Impact of Change

- **Users**: 
Users now see all content before deploying, and allow naming of resources when a new logic app is created.
- **Developers**: 
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in:

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -25,6 +25,24 @@ pnpm run build:lib  # or pnpm turbo run build:lib
 pnpm run build:extension
 ```
 
+### VS Code Extension E2E Tests (ExTester)
+```bash
+# Build + run all phases
+cd apps/vs-code-designer
+npx tsup --config tsup.e2e.test.config.ts
+node src/test/ui/run-e2e.js
+
+# Run specific phases
+$env:E2E_MODE="createonly"     # Phase 4.1: workspace creation
+$env:E2E_MODE="designeronly"   # Phase 4.2: designer lifecycle
+$env:E2E_MODE="newtestsonly"   # Phases 4.3-4.6: new tests
+node src/test/ui/run-e2e.js
+```
+
+**Key knowledge files for E2E tests:**
+- `apps/vs-code-designer/src/test/ui/SKILL.md` — Complete learning document (700+ lines)
+- `apps/vs-code-designer/CLAUDE.md` — Critical rules for writing new tests
+
 ### Testing
 ```bash
 # Run all unit tests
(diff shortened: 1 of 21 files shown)


apps/vs-code-designer/src/app/commands/createLogicApp/createLogicApp.ts
```diff
@@ -8,9 +8,49 @@ import type { SlotTreeItem } from '../../tree/slotsTree/SlotTreeItem';
...
 import { type MessageItem, window } from 'vscode';
 
+/**
+ * Creates a Logic App without showing wizard prompts - all required information must be provided in the context.
+ * @param context - Context with pre-filled values (newSiteName, location, resourceGroup or newResourceGroupName)
+ * @param subscription - Subscription ID or tree item
+ * @param skipNotification - If true, skips the completion notification
+ * @returns The created Logic App tree item
+ */
+export async function createLogicAppWithoutWizard(
```
andrew-eldridge, line 21: Shouldn't SubscriptionTreeItem.createChild already skip prompts if the values are provided in context? And if not, would it make sense to modify that method to skip prompts when not needed instead of adding this one?

apps/vs-code-designer/src/app/commands/deploy/deployWebview.ts
```diff
@@ -0,0 +1,117 @@
...
+          plan: data.isCreatingNewAppServicePlan ? undefined : { id: data.appServicePlan },
+          appServicePlanSku: data.appServicePlanSku || 'WS1',
+          newStorageAccountName: data.isCreatingNewStorageAccount ? data.storageAccount : undefined,
+          storageAccount: data.isCreatingNewStorageAccount ? undefined : { id: data.storageAccount },
+          createAppInsights: data.createAppInsights,
+          newAppInsightsName: data.appInsightsName,
+        };
+
+        // Create the Logic App using the wizard-free method
+        const node: SlotTreeItem = await createLogicAppWithoutWizard(
```
andrew-eldridge, line 79: Isn't createLogicApp called from deploy if the logic app doesn't exist? Can this be refactored so that the logic app is either always created prior to deploy or always created within deploy function?

PR #8828, opened 2026-02-19
feat(vscode) - Adding additional csharp lsp server functionality

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Adding ability to open overview page for a workflow, adding a new workflow, reducing the connection pane load time, and updating the package.

## Impact of Change

- **Users**: 
Workflows runs can now be opened and new workflows can be added
- **Developers**: 
- **System**: 

## Test Plan

- [x] E2E tests added/updated
- [ ] Tested in:


apps/vs-code-designer/src/app/commands/deploy/deploy.ts
```diff
@@ -86,21 +89,35 @@ async function deploy(
...
+      context.telemetry.properties.isCodefulProject = 'true';
+      ext.outputChannel.appendLog(localize('buildingCodefulProject', 'Building and publishing codeful Logic App project...'));
+
+      // Build the codeful project
       await publishCodefulProject(actionContext, logicAppNode);
+      ext.outputChannel.appendLog(localize('codefulProjectPublished', 'Codeful project built and published successfully.'));
+    } else {
+      // For codeless projects, build custom code functions if they exist
+      const customFolderExists = await fse.pathExists(path.join(logicAppNode.fsPath, libDirectory, customDirectory));
+      if (customFolderExists) {
```
andrew-eldridge, line 118: Shouldn't this be if !customFolderExists? Since we would build in the case where lib/custom/* doesn't already exist for codeless, i.e. not already built

apps/vs-code-designer/src/app/commands/deploy/deploy.ts
```diff
@@ -86,21 +89,35 @@ async function deploy(
...
+        await buildCustomCodeFunctionsProject(actionContext, logicAppNode);
```
andrew-eldridge: buildCustomCodeFunctionsProject is supposed to check whether the .net project is specifically a custom code functions project before building, part of that is to check for Microsoft.Azure.Workflows.Webjobs.Sdk in the .csproj, which I don't think will exist for codeful projects. Should there be a separate function for handling building codeful?

apps/vs-code-designer/src/app/commands/createNewCodeProject/CodeProjectBase/CreateLogicAppWorkspace.ts
```diff
@@ -112,38 +113,84 @@ export async function createLogicAppAndWorkflow(webviewProjectContext: IWebviewP
...
-const createAgentCodefulWorkflowFile = async (
+/**
+ * Adds a workflow Build() call to an existing Program.cs file.
+ */
+export const addWorkflowToProgram = async (programFilePath: string, workflowName: string): Promise<void> => {
+  const programContent = await fse.readFile(programFilePath, 'utf-8');
+
+  // Find the location to insert the new workflow builder
+  // Look for the "Build all workflows" comment or insert before host.Run()
+  const buildCallRegex = /(\s*)(\/\/ Build all workflows\s*\n(?:\s*\/\/.*\n)*\s*)(.*?)(\s*host\.Run\(\);)/s;
```
andrew-eldridge, line 124: Is it possible to avoid using a regex here to update Program.cs with a new workflow, and instead just overwrite Program.cs with one containing all workflows? We would need to track the workflows in a codeful project elsewhere, but we should probably avoid using regex as much as possible since it's bound to break in some cases

andrew-eldridge (approved):

PR #8921, opened 2026-03-15
fix(designers): Improving action name validation

Santiagosala2:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

When you select an action or node to change its name in the details panel, you type a name and then realize that you added an extra space at the end. When you try to remove it (without hitting Enter), it says "The name already exists..." (even though there are no other actions with that name). Finally, your given name gets removed.

Root cause: The selected node's previous name is being included in the name uniqueness validation. The validation should only check against the names of other nodes.

This improves the user experience, as it's common to rename actions, mistype or add extra spaces, and want to correct them mid-edit — especially when copying names from elsewhere.

## Impact of Change

- **Users**: better user experience
- **Developers**: 
- **System**: 

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Standalone designer 

## Contributors
@Santiagosala2


Santiagosala2:
@microsoft-github-policy-service agree

PR #8965, opened 2026-03-26
feat(designer): Add Foundry Models (Preview) agent model source

Elaina-Lee:
## Commit Type

## Risk Level

## What & Why
Add **"Foundry Models (Preview)"** (`MicrosoftFoundry`) as a new option in the Agent Model Source dropdown for the Agent Loop action. This enables users to select Foundry-hosted models in addition to Azure OpenAI.

Also fixes incorrect model versions in `AGENT_MODEL_CONFIG` (e.g. `gpt-4o` from `2025-04-14` → `2024-11-20`, `gpt-4` → `turbo-2024-04-09`) and adds 22 new models (`gpt-5.4` series, `o3`, `o4-mini`, `o1`, `model-router`, etc.).

Key changes:
- Rename `SUPPORTED_AGENT_MODELS` → `SUPPORTED_AGENT_OPENAI_MODELS` (curated 10-model list for Azure OpenAI)
- Add `SUPPORTED_FOUNDRY_AGENT_MODELS` as a full superset of all 31 models
- Deployment filtering uses the correct model list based on selected `agentModelType`
- For `MicrosoftFoundry`, `deploymentModelProperties.name` is set to the raw model ID (not an Azure deployment instance), with format/version looked up from `AGENT_MODEL_CONFIG`
- "Create new deployment" modal always shows the full model list
- All changes mirrored in both designer v1 and designer v2

## Impact of Change
- **Users**: New "Foundry Models (Preview)" option in Agent model source dropdown; corrected model versions for deployments
- **Developers**: `SUPPORTED_AGENT_MODELS` renamed to `SUPPORTED_AGENT_OPENAI_MODELS`; new `SUPPORTED_FOUNDRY_AGENT_MODELS` constant; `getEditorAndOptions` now accepts optional `agentModelType` parameter
- **System**: No architectural changes; additive manifest update in `agentloop.ts`

## Test Plan
- Tested in: Standalone designer

### Unit tests added
- `deploymentModelResource.spec.tsx` (v1 & v2) — 7 tests each: rendering, submit, cancel, error handling, model list verification
- `helpers.spec.ts` (v2, new) — 19 tests for agent connector helpers and `getConnectionToAssign`
- `helpers.spec.ts` (v1, expanded) — added `MicrosoftFoundry` cases and edge cases
- `Utils.test.ts` (v1 & v2) — added `MicrosoftFoundry` model type assertion

### Manual verification
- [ ] Open standalone designer with an agent workflow
- [ ] Verify "Foundry Models (Preview)" appears in the Agent model source dropdown
- [ ] Select "Azure OpenAI" — verify AI model combobox filters by curated OpenAI list
- [ ] Select "Foundry Models (Preview)" — verify AI model combobox filters by full Foundry list
- [ ] Click "Create new" on deployment model — verify dropdown shows all models
- [ ] Create a new deployment — verify correct version in API request (Network tab)

## Screenshots/Videos
https://github.com/user-attachments/assets/fa782f40-c7e8-4e9f-8c97-78f4dcee0a16

image

### DIFF
--- a/libs/designer-v2/src/lib/ui/common/resourcepicker/resourcefield.tsx
+++ b/libs/designer-v2/src/lib/ui/common/resourcepicker/resourcefield.tsx
@@ -36,7 +36,7 @@ export const ResourceField = ({
     }),
   };
 
-  const sortedResources = useMemo(() => resources.sort((a, b) => a.displayName.localeCompare(b.displayName)), [resources]);
+  const sortedResources = useMemo(() => resources.sort((a, b) => (a.displayName ?? '').localeCompare(b.displayName ?? '')), [resources]);
 
   const [selectedResource, setSelectedResource] = useState<string | undefined>('');
   useEffect(() => {
--- a/libs/designer-v2/src/lib/ui/panel/connectionsPanel/selectConnection/selectConnection.helpers.ts
+++ b/libs/designer-v2/src/lib/ui/panel/connectionsPanel/selectConnection/selectConnection.helpers.ts
@@ -21,7 +21,7 @@ export const compareFlattenedConnections = (a: ConnectionWithFlattenedProperties
   if (!a.invalid && b.invalid) {
     return -1;
   }
-  return a.displayName.localeCompare(b.displayName);
+  return (a.displayName ?? '').localeCompare(b.displayName ?? '');
 };
 
 export const getLabelForConnection = (item: ConnectionProperties): string => {
--- a/libs/designer/src/lib/ui/common/resourcepicker/resourcefield.tsx
+++ b/libs/designer/src/lib/ui/common/resourcepicker/resourcefield.tsx
@@ -76,7 +76,7 @@ const ResourceFieldDropdown = ({
   };
   const isDropdownSizeSmall = renderType === 'default';
 
-  const sortedResources = useMemo(() => resources.sort((a, b) => a.displayName.localeCompare(b.displayName)), [resources]);
+  const sortedResources = useMemo(() => resources.sort((a, b) => (a.displayName ?? '').localeCompare(b.displayName ?? '')), [resources]);
   const [selectedResource, setSelectedResource] = useState<string | undefined>('');
   useEffect(() => {
     if (!isLoading) {
--- a/libs/designer/src/lib/ui/panel/connectionsPanel/selectConnection/selectConnection.helpers.ts
+++ b/libs/designer/src/lib/ui/panel/connectionsPanel/selectConnection/selectConnection.helpers.ts
@@ -21,7 +21,7 @@ export const compareFlattenedConnections = (a: ConnectionWithFlattenedProperties
   if (!a.invalid && b.invalid) {
     return -1;
   }
-  return a.displayName.localeCompare(b.displayName);
+  return (a.displayName ?? '').localeCompare(b.displayName ?? '');
 };
 
 export const getLabelForConnection = (item: ConnectionProperties): string => {
--- a/libs/logic-apps-shared/src/designer-client-services/lib/standard/foundryAgentService.ts
+++ b/libs/logic-apps-shared/src/designer-client-services/lib/standard/foundryAgentService.ts
@@ -273,7 +273,7 @@ function extractVersionsData(response: unknown): FoundryAgentVersion[] {
 
   // Wrapped shape: { result: { data: [...] } }
   const result = resp['result'] as Record<string, unknown> | undefined;
-  if (Array.isArray(result?.['data'])) {
+  if (result && Array.isArray(result['data'])) {
     return result['data'] as FoundryAgentVersion[];
   }
 


libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1894,6 +1890,9 @@ export const updateParameterAndDependencies = createAsyncThunk(
           });
         }
 
+        // for debugging
+        inputDependencies['inputs.$.agentModelSettings.deploymentModelProperties.name'].definition.default = 'foo';
```
takyyon: **CRITICAL: Debug code left in — will ship to production.**

This hardcodes `'foo'` as the default deployment model name in `updateParameterAndDependencies`. Every agent action that runs through the parameter dependency pipeline will have its deployment model name corrupted.

This is a **merge blocker** — please remove these two lines.

libs/designer-v2/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/custom/__test__/deploymentModelResource.spec.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    renderComponent({ metadata: { cognitiveServiceAccountId: 'test-account-id' }, onClose });
+
+    const submitButton = screen.getByText('Submit');
+    await user.click(submitButton);
+
+    await waitFor(() => {
+      expect(mockCreateNewDeployment).toHaveBeenCalledWith(
+        expect.stringContaining('model-'),
+        constants.SUPPORTED_FOUNDRY_AGENT_MODELS[0],
```
takyyon: **Bug: Test assertion uses wrong model list.**

The test renders the component with `metadata: { cognitiveServiceAccountId: 'test-account-id' }` — `agentModelType` is **not** set. So the component defaults to `SUPPORTED_AGENT_OPENAI_MODELS`, meaning `modelKey` starts at `SUPPORTED_AGENT_OPENAI_MODELS[0]` which is `'gpt-5'`.

But this assertion expects `SUPPORTED_FOUNDRY_AGENT_MODELS[0]` which is `'gpt-5.4-nano'`. These are different values.

Either:
- Pass `metadata: { cognitiveServiceAccountId: 'test-account-id', agentModelType: 'MicrosoftFoundry' }` to match the Foundry expectation, or
- Change the expectation to `constants.SUPPORTED_AGENT_OPENAI_MODELS[0]`

(Same issue exists in the identical v1 test file.)

libs/designer-v2/src/lib/common/constants.ts
```diff
@@ -998,6 +998,39 @@ export default {
     'gpt-4.1-nano',
     'gpt-4o-mini',
   ],
+  SUPPORTED_FOUNDRY_AGENT_MODELS: [
```
takyyon: **Model lists duplicated in 3 places — will drift.**

This identical 31-model list now exists in:
1. `libs/designer/src/lib/common/constants.ts`
2. `libs/designer-v2/src/lib/common/constants.ts` (here)
3. `libs/logic-apps-shared/src/utils/src/lib/models/agent.ts` (as `AGENT_MODEL_CONFIG` keys)

Same for the 10-model OpenAI list (duplicated between v1 and v2).

Consider defining these once in `logic-apps-shared` (e.g., `Object.keys(AGENT_MODEL_CONFIG)` for the full list) and importing into both designers. That way model additions only need one edit.

libs/designer-v2/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -141,6 +142,19 @@ const updateAgentParametersForConnection = (
...
+  if (agentModelTypeValue === 'AzureOpenAI') {
+    const paramGroups = state.operations.inputParameters[nodeId]?.parameterGroups;
+    const defaultGrp = paramGroups?.[ParameterGroupKeys.DEFAULT];
+    const existingParam = defaultGrp?.parameters?.find((p) => p.parameterKey === 'inputs.$.agentModelType');
+    const currentValue = existingParam?.value?.[0]?.value;
+    const validManifestValues = Object.values(displayNameToManifestValue);
+    if (currentValue && validManifestValues.includes(currentValue) && currentValue !== 'AzureOpenAI') {
+      return;
+    }
+  }
```
takyyon, line 149: **Early return is too aggressive — skips all agent parameter updates.**

When `agentModelTypeValue === 'AzureOpenAI'` and a valid non-AzureOpenAI value exists in state, this `return` exits the **entire** `updateAgentParametersForConnection` function — not just the `agentModelType` update.

This means if a user changes the connection while `MicrosoftFoundry` is selected, other agent parameters (like `deploymentId` clearing, cognitive service account updates) won't be updated.

Suggestion: Instead of `return`, set `agentModelTypeValue = currentValue` and let the function continue with the rest of the parameter updates:
```ts
if (currentValue && validManifestValues.includes(currentValue) && currentValue !== 'AzureOpenAI') {
  agentModelTypeValue = currentValue; // preserve existing, but continue updates
}
```

(Same issue in v1 `connections.ts`.)

libs/designer-v2/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/index.tsx
```diff
@@ -991,32 +992,68 @@ export const ParameterSection = ({
...
+            directDeploymentModelUpdates.push({
+              groupId: group.id,
+              parameterId: targetParam.id,
+              propertiesToUpdate: { value: [createLiteralValueSegment(resolvedValue)] },
+            });
+          }
+        }
+
+        if (directDeploymentModelUpdates.length > 0) {
+          dispatch(updateNodeParameters({ nodeId, parameters: directDeploymentModelUpdates }));
```
takyyon: **Direct dispatch alongside async thunk — potential race condition.**

This synchronous `dispatch(updateNodeParameters(...))` writes deployment model properties directly to Redux. But the `AgentSchema` dependency pipeline (a few lines below at the `inputDependencies[key].dependencyType === 'AgentSchema'` check) also writes to the same parameters asynchronously via thunks.

Both paths write to the same Redux state keys (`deploymentModelProperties.name/format/version`), so the async thunk could overwrite these direct updates or vice versa.

If the intent is to ensure reliable writes, consider either:
- Removing the dependency-based path for these keys, or
- Removing this direct dispatch and fixing the thunk pipeline instead

Having both creates a dual-write race.

libs/designer-v2/src/lib/ui/panel/nodeDetailsPanel/tabs/parametersTab/connectionInline.tsx
```diff
@@ -98,13 +100,31 @@ export const ConnectionInline: React.FC<ConnectionInlineProps> = ({ showSubCompo
...
 
+  const agentModelTypeFromState = useSelector((state: RootState) => {
+    const paramGroups = state.operations.inputParameters[nodeId]?.parameterGroups;
+    const defaultGroup = paramGroups?.[ParameterGroupKeys.DEFAULT];
+    const param = defaultGroup?.parameters?.find((p: any) => p.parameterKey === 'inputs.$.agentModelType');
+    return param?.value?.[0]?.value ?? '';
+  });
+
+  const agentModelTypeDisplayName = useMemo(() => {
+    const manifestToDisplayName: Record<string, string> = {
```
takyyon: **Duplicated mapping — extract to shared constant.**

This `manifestToDisplayName` mapping is duplicated in both v1 and v2 `connectionInline.tsx`, and a reverse mapping exists in `connections.ts:updateAgentParametersForConnection`.

Consider adding a static mapping to `AgentUtils` (e.g., `AgentUtils.ManifestToDisplayName`) so all three callsites share one source of truth. When a new model type is added, only one place needs updating.

libs/designer-v2/src/lib/ui/panel/nodeDetailsPanel/tabs/codeViewTab.tsx
```diff
@@ -15,7 +15,7 @@ export const CodeViewTab: React.FC<PanelTabProps> = (props) => {
   const queryData = useQuery(['serialization', { nodeId }], () => serializeOperation(rootState, nodeId), {
     retry: false,
     cacheTime: 0,
-    refetchOnMount: false,
+    refetchOnMount: 'always',
```
takyyon: **Nit: `refetchOnMount` changed from `false` to `'always'` — intentional?**

This means every tab switch to code view triggers a full re-serialization of the operation. If this is fixing a staleness bug, great — but it's unrelated to the Foundry feature and could be a perf regression for users who toggle tabs frequently. Worth calling out in the PR description if intentional.

libs/logic-apps-shared/src/designer-client-services/lib/standard/manifest/agentloop.ts
```diff
@@ -491,7 +495,7 @@ export default {
                     parameters: [
                       {
                         name: 'agentModelType',
-                        values: ['AzureOpenAI'],
+                        values: ['MicrosoftFoundry'],
```
takyyon: **Note: `deploymentModelProperties` visibility narrowed from `AzureOpenAI` to `MicrosoftFoundry` only.**

Previously `.name`, `.format`, `.version` were visible when `agentModelType === 'AzureOpenAI'`. Now they're only visible for `MicrosoftFoundry`. If any consumer relied on these being visible for AzureOpenAI, this is a breaking change.

I assume this is intentional (AzureOpenAI gets these from the deployment ARM response programmatically) — if so, worth a brief note in the PR description for reviewers.

libs/designer/src/lib/core/utils/parameters/helper.ts
```diff
@@ -1894,6 +1890,9 @@ export const updateParameterAndDependencies = createAsyncThunk(
           });
         }
```
takyyon, line 1900: **Minor: Unrelated eslint-disable removals and test additions.**

Several `eslint-disable no-param-reassign` and `no-case-declarations` comments were removed across this file, and new tests for `titleCase`, `isOpenApiSchemaVersion`, `getSKUDefaultHostOptions` were added in `Utils.test.ts`. These are good cleanups but unrelated to the Foundry feature — consider splitting into a separate PR for cleaner git history and easier reverts.

libs/designer-v2/src/lib/core/actions/bjsworkflow/connections.ts
Elaina-Lee: Before (the bug): The inner if block had return; which exited the entire function, skipping deploymentId clearing, modelId updates, and the dispatch
  call.

  After (current code): Changed return; to agentModelTypeValue = currentValue;, which preserves the existing valid value but lets the function continue
  to update all other agent parameters (lines 151-213).

  This was fixed in both files (v1 and v2) in commit 6f7df9011. No further changes needed.

PR #9052, opened 2026-04-13
 Fix MCP connection authentication schema in consumption workflow.json

Bhavd13:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why
Fixes the `workflow.json` schema for both BuiltIn and Managed MCP tools in consumption Logic Apps to correctly include authentication details.

For BuiltIn MCP, the Authentication field is now serialized as a structured object instead of a flat string, and all auth types (MSI, Key, Basic, OAuth, etc.) are properly round-tripped through serialization/deserialization. For Managed MCP (e.g., GitHub MCP), `connectionProperties.authentication` with MSI details is now included in the connections parameter when the Logic App has a managed identity configured. Also fixes the identity picker resetting user-assigned identity selection due to an unstable `useEffect` dependency.

## Impact of Change

- **Users**:  Consumption Logic App users adding MCP servers with managed identity authentication will now see correct Authentication schema in `workflow.json` — identity, audience, and other auth properties are properly included for both BuiltIn and Managed MCP tools.
- **Developers**: No API changes.  Authentication field on BuiltIn MCP is now an object instead of a flat string for non-None auth types. `getConnectionPropertiesIfRequired` now recognizes managed MCP connectors for MSI auth.
- **System**: No performance or dependency changes. Fixes serialization/deserialization round-trip for all MCP auth types.

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: Manually verified in standalone designer against real consumption Logic App with both system-assigned and user-assigned managed identity

## Contributors
@Bhavd13 

## Screenshots/Videos

https://github.com/user-attachments/assets/d1fc73cb-0e24-4011-beb6-c15ceb83d196


libs/designer-v2/src/lib/common/constants.ts
```diff
@@ -53,6 +53,34 @@ const VARIABLE_TYPE = {
...
+  'clientId',
+  'secret',
+  'tenant',
+  'authority',
+  'pfx',
+] as const;
+
+/**
+ * Checks whether a connector is a managed MCP connector (API Hub MCP, not built-in).
+ * Uses the managedApis path segment + 'mcp' suffix convention (e.g., a365mcp, githubmcp, stripemcp).
```
rllyy97: Is there a better way to confirm that they are MCP connectors?
The ID may just have mcp at the end but not be a proper MCP connector

libs/designer-v2/src/lib/core/actions/bjsworkflow/connections.ts
```diff
@@ -310,11 +310,21 @@ const updateNodeConnectionAndProperties = async (
...
+  if (
+    !isManagedMcpConnector(connector) &&
+    !isConnectionMultiAuthManagedIdentityType(connection, connector) &&
+    !isConnectionSingleAuthManagedIdentityType(connection)
+  ) {
     return undefined;
   }
 
   const identity = WorkflowService().getAppIdentity?.();
+  if (!identity) {
```
rllyy97, line 324: Why was this early return added?

libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnectionInternal.tsx
```diff
@@ -122,6 +127,9 @@ export const CreateConnectionInternal = (props: {
         const userAssignedIdentity = selectedIdentity !== constants.SYSTEM_ASSIGNED_MANAGED_IDENTITY ? selectedIdentity : undefined;
         payload.connectionProperties = getConnectionProperties(connector as Connector, userAssignedIdentity);
```
rllyy97, line 128: It looks like we're not using `additionalConnectionProperties` here, is that intended?

rllyy97 (approved):

libs/designer-v2/src/lib/core/actions/bjsworkflow/connections.ts
Bhavd13: Actually previously this function only ran for multi-auth or single-auth MSI connections, where the app always has a managed identity. But now we want Managed Mcp Connector to also pass this guard. Kusto managed connector uses MI. Without this, the managed MCP connectors using non-MSI auth (e.g., API key) can reach this point, identity would be undefined and the code below would fail.

libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnectionInternal.tsx
Bhavd13: Yes, intentional. L126-128 -- Auth is built from the selected identity via `getConnectionProperties()` This comes from the identity picker.

L130-132 -- MSI auth properties come from the connection creation form values, not from the identity picker. These are passed in as `additionalConnectionProperties.`

rllyy97 (approved):
Reapproved

PR #9080, opened 2026-04-20
fix(vscode): Improve dev-container onboarding and local debug startup

lambrianmsft:
## Commit Type

- [x] fix - Bug fix

## Risk Level

- [x] High - Major changes, significant user/system impact

## What & Why

- Skip dependency onboarding in dev-container workspaces while auto-starting design-time APIs.
- Centralize extension asset resolution so templates and scripts work consistently in source, tests, and packaged builds.
- Harden design-time startup with tracked startup state, retry limits, safer process cleanup, and clearer startup failure handling.
- Improve workflow/custom-code debug attach selection and output logging, especially for Windows child-process resolution.
- Update templates, recommendations, tests, and config to support remote containers and cover the new startup/debug behavior.
- Hide devcontainer experience for now

## Impact of Change

- **Users**: 
- Users should see more stable behavior with startup for both types of workspaces.
- **Developers**: 
- N/A
- **System**: 
- N/A
## Test Plan

- [x] E2E tests added/updated
- [ ] Tested in:


apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -295,13 +310,83 @@ async function startFuncTask(
   }
 }
 
+async function getWorkflowDebugProcessCandidates(taskInfo: IRunningFuncTask): Promise<Array<string | undefined>> {
+  const firstChildProcessId = taskInfo.childProcessId?.[0] ?? (await pickChildProcess(taskInfo));
+  const hostChildProcessId = taskInfo.childProcessId?.[1] ?? (await pickFuncHostChildProcess(taskInfo));
+
+  taskInfo.childProcessId = [firstChildProcessId, hostChildProcessId];
```
andrew-eldridge: taskInfo.childProcessId is updated here and is also updated using the return value, can we leave taskInfo unchanged in this function?

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
```diff
@@ -116,10 +116,22 @@ export default class OpenDesignerForLocalProject extends OpenDesignerBase {
 
     await startDesignTimeApi(this.projectPath);
 
-    if (!ext.designTimeInstances.has(this.projectPath)) {
```
andrew-eldridge, line 119: Does this throw an error if projectPath is not in designTimeInstances? Since we don't check if it's present before getting now

andrew-eldridge:
this does seem like a high risk

andrew-eldridge (approved):

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
lambrianmsft: Updated this so getWorkflowDebugProcessCandidates() is now pure and no longer mutates taskInfo. The childProcessId cache assignment now happens explicitly at the call sites, so we are no longer mutating shared state and returning the same cached value from the helper.

apps/vs-code-designer/src/app/commands/workflows/openDesigner/openDesignerForLocalProject.ts
lambrianmsft: Map.get() itself does not throw here, and the missing-entry case is already handled immediately below. If projectPath is not present in designTimeInstances, we hit the !designTimeInstance branch and throw the
localized designTimeNotRunning error before reading any properties from the instance.

apps/vs-code-designer/src/app/utils/findChildProcess/findChildProcess.ts
```diff
@@ -49,7 +49,7 @@ async function runPowerShellScript(scriptPath: string, ...args: string[]): Promi
 
 export async function getChildProcessesWithScript(parentProcessId: number): Promise<ProcessInfo[]> {
   try {
-    const scriptPath = path.join(__dirname, 'assets', 'scripts', 'get-child-processes.ps1');
+    const scriptPath = getExtensionAssetPath('scripts', 'get-child-processes.ps1');
     const output = await runPowerShellScript(scriptPath, parentProcessId.toString());
```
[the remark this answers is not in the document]
lambrianmsft: The catch now formats the thrown value safely with error instanceof Error ? error.message : String(error) before building the new error, so non-Error rejections from PowerShell or JSON parsing are handled correctly.

andrew-eldridge (approved):

PR #9199, opened 2026-05-20
(fix): Show managed identity picker for custom MCP connector connection creation

Bhavd13:
## Commit Type

- [x] fix - Bug fix

## Risk Level

- [ ] Low - Very less impact
- [x] Medium - Moderate changes, some user impact. The change only affects UI rendering for custom MCP connector connection creation (adding a managed identity picker that was previously hidden); no backend logic, serialization, or existing connector flows are altered.

## What & Why
Custom MCP connectors with managed identity auth show no authentication options when creating a connection. Users only see a connection name field with no identity picker.

Root cause: `CreateConnection` component explicitly hides parameters of type `managedIdentity`. When the connector's only parameter set (e.g., oauthMI) uses this type, all parameters are hidden — leaving an empty form.

Fix:
Detect when the selected multi-auth parameter set contains a `managedIdentity` type parameter 
Render the managed identity picker for that case
Add fallback MCP auth parameter sets (None, Basic, Key, MI) for custom connectors missing `connectionParameterSets`

## Impact of Change

- **Users**: Custom MCP connector connection creation now shows the managed identity picker when the connector defines MI auth
- **Developers**: New optional `connectionParameterSetsOverride` prop on `CreateConnectionInternal` (designer-v2)
- **System**: No performance or architecture changes

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Consumption Logic App (Brazil South) with custom MCP connector using `oauthMI` managed identity parameter set. 
The fix adds conditional rendering of an existing component based on a new boolean flag; the component itself is already tested, and the behavior is validated through manual E2E testing with a real custom MCP connector.

## Contributors
@Bhavd13 

## Screenshots/Videos

This is the default behaviour which will appear when user will select a custom mcp server. They will have to select the auth type and fill in the details 

image

This is the behaviour when a custom deployed MCP with auth defined as MI is used. This is the use case for ICM team
image

### DIFF
--- a/libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnection.tsx
+++ b/libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnection.tsx
@@ -232,6 +232,15 @@ export const CreateConnection = (props: CreateConnectionProps) => {
   const isMultiAuth = useMemo(() => (connectionParameterSets?.values?.length ?? 0) > 0, [connectionParameterSets?.values]);
   const showMultiAuthDropdown = useMemo(() => (connectionParameterSets?.values?.length ?? 0) > 1, [connectionParameterSets?.values]);
 
+  // Check if the currently selected multi-auth parameter set is a managed identity type.
+  // When true, we show the identity picker instead of the hidden managedIdentity parameter.
+  const isMultiAuthManagedIdentitySet = useMemo(() => {
+    if (!isMultiAuth) {
+      return false;
+    }
+    return Object.values(multiAuthParams).some((param) => param.type === ConnectionParameterTypes.managedIdentity);
+  }, [isMultiAuth, multiAuthParams]);
+
   const hasOnlyOnPremGateway = useMemo(
     () => (connectorCapabilities?.includes(Capabilities.gateway) && !connectorCapabilities?.includes(Capabilities.cloud)) ?? false,
     [connectorCapabilities]
@@ -439,6 +448,9 @@ export const CreateConnection = (props: CreateConnectionProps) => {
     if (legacyManagedIdentitySelected && !selectedManagedIdentity) {
       return false;
     }
+    if (isMultiAuthManagedIdentitySet && !selectedManagedIdentity) {
+      return false;
+    }
     if (Object.keys(capabilityEnabledParameters ?? {}).length === 0) {
       return true;
     }
@@ -497,7 +509,7 @@ export const CreateConnection = (props: CreateConnectionProps) => {
     }
 
     const alternativeParameterValues = legacyManagedIdentitySelected ? {} : undefined;
-    const identitySelected = legacyManagedIdentitySelected ? selectedManagedIdentity : undefined;
+    const identitySelected = legacyManagedIdentitySelected || isMultiAuthManagedIdentitySet ? selectedManagedIdentity : undefined;
 
     return createConnectionCallback?.(
       showNameInput ? connectionDisplayName : undefined,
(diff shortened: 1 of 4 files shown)


takyyon:
@preetriti1 Mind reviewing this when you get a chance please?

takyyon (approved):

preetriti1:
The Authentication Type input is getting removed from UI after Managed Identity is selected, why is that? Shouldn't we allow user to change the auth type?

preetriti1 (changes requested):
Can you please check why the auth dropdown is missing after selecting MI

Bhavd13:
> The Authentication Type input is getting removed from UI after Managed Identity is selected, why is that? Shouldn't we allow user to change the auth type?

@preetriti1 
The second image where you see the auth type missing when MI is selected uses a custom template deployed as a MCP connector. The ICM team will be using a deployed custom template, this template will already have auth type defined as MI, hence no dropdown for auth type will appear. They will also populate the audience property in the template itself so it wont be added here again since the custom connector will create a connection on its own using those params. Here is a snippet from the template which defines auth - 

`"connectionParameterSets": {
            "uiDefinition": {
                "displayName": "Authentication type",
                "description": "Type of authentication to be used."
            },
            "values": [
                {
                    "name": "oauthMI",
                    "uiDefinition": {
                        "displayName": "Logic Apps Managed Identity",
                        "description": "Create a connection using a Logic Apps Managed Identity"
                    },
                    "parameters": {
                        "token": {
                            "type": "managedIdentity",
                            "managedIdentitySettings": {
                                "resourceUri": "api://icmmcpapi-ppe/mcp.tools"
                            },
                            "uiDefinition": {
                                "displayName": "Logic Apps Managed Identity",
                                "tooltip": "Managed Identity",
                                "description": "Sign in with a Logic Apps Managed Identity",
                                "constraints": {
                                    "location": "logicapp",
                                    "required": "true"
                                }
                            }
                        }
                    },`

Bhavd13:
> Can you please check why the auth dropdown is missing after selecting MI

So the way I have made changes is that, if a custom template is not deployed in the mcp custom connector, then it will ask users to connect and will provide all the auth options as shown. This will be a fallback method. Incase they have a template defined and auth selected (which is the use case for ICM team) then they dont need to select auth type here again and fill in the params. The schema is shared by Rama so if you have any questions related to that, I can add you to the group too. 
I did have a question though. Do we need to allow auth selection in the custom mcp if schema doesnt define auth type? Because previously we were not doing it that way, so I wonder if we should bring that change as it might affect others using custom mcp

libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnection.tsx
```diff
@@ -232,6 +232,19 @@ export const CreateConnection = (props: CreateConnectionProps) => {
...
+    if (!isMultiAuth) {
+      return false;
+    }
+    return Object.values(multiAuthParams).some(
+      (param) =>
+        param.type === ConnectionParameterTypes.managedIdentity ||
+        equals(param.uiDefinition?.constraints?.default, 'managedserviceidentity')
+    );
+  }, [isMultiAuth, multiAuthParams]);
+
```
preetriti1: See, you are no where checking anything specific to mcp custom template scenario but a generic check if the parameter managedIdentity is set...
Are you sure this is not true in other cases like legacyMultiAuth and also other connectors like serviceProviders/managed apis where user has selected managed identity.? because then you might be breaking those scnearios to show the extra input in connection creation.

preetriti1 (approved):
I do have some concerns but dont want to block this PR

libs/designer-v2/src/lib/ui/panel/connectionsPanel/createConnection/createConnection.tsx
Bhavd13: `isMultiAuthManagedIdentitySet` is now scoped. It requires both `enableManagedIdentityPicker === true` (only set by MCP wizard) AND empty visible params. SQL, Blob, and all other connectors never set this flag so no change to their behavior.

PR #9253, opened 2026-06-05
feat(designer) -  Show operation default setting values and host settings in the designer

JoaquimMalcampo:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Adds UI visibility for operation setting defaults returned by the settingDefaults API. Previously, settings like retry policy showed "Default" with no detail fields when no explicit value was set in the workflow definition. Now the designer fetches defaults from the host and displays actual values (e.g., "Exponential interval", count, intervals) while preserving the existing serialization behavior — no changes are written to the workflow JSON.

Additionally, unknown host-level settings returned by the API (e.g., maxMessageBatchSize) are now surfaced in a new read-only "Host settings" section in the settings panel.

## Changes
- settingDefaults.ts - React Query cached fetch for the settingDefaults API endpoint. Applies merge logic for returned setting values and routes additional settings as hostSettings.
- added defaultHint to SettingData<T> and hostSettings map to Settings interface
- serializer.ts - if defaultHint is present, does not serialize on retry
- hostsettings.tsx - dynamic read only section that renders host settings as text fields and is included with the rest of the settings sections in index.tsx
- operationmanifest.ts (standard/consumption/base) - add getSettingDefaults to operation manifest service interface and implementations.

## Impact of Change

- **Users**:  
Settings panel now shows actual default values for retry policy and other settings instead of an opaque "Default" label. Host-level read-only settings (e.g., maxMessageBatchSize) are visible in a new collapsible section. No changes to saved workflow definitions.
- **Developers**: 
New defaultHint field on SettingData<T> — indicates the value came from API defaults and should not be serialized. hostSettings on Settings is a dynamic Record<string, SettingData<unknown>> for arbitrary host-returned keys. KNOWN_SETTING_KEYS set in settingDefaults.ts controls which API keys are treated as known vs host-level.
- **System**: 
One additional HTTP call per operation during deserialization (cached by React Query). No changes to workflow JSON serialization output.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: VS Code Extension Development Host with Standard Logic App (Service Bus completeQueueMessageV2 operation)

## Screenshots/Videos
setting_enhancements_part2


libs/designer/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -349,6 +368,15 @@ export const initializeOperationDetails = async (
     );
 
     settings = addDefaultSecureSettings(settings, connector?.properties?.isSecureByDefault ?? false);
+    const swaggerDefaults = await fetchSettingDefaults(
```
andrew-eldridge: mcpDefaults, manifestDefaults, and swaggerDefaults are fetched the same. Can these be combined earlier in the method?

libs/designer/src/lib/core/queries/settingDefaults.ts
```diff
@@ -0,0 +1,125 @@
+import { OperationManifestService } from '@microsoft/logic-apps-shared';
+import type { Settings, SettingData } from '../actions/bjsworkflow/settings';
+import { getReactQueryClient } from '../ReactQueryProvider';
+import Constants from '../../common/constants';
+
+const KNOWN_SETTING_KEYS = new Set<string>([
```
andrew-eldridge: how is the settings view usually updated with new settings? Is it possible that this list will fall out of sync if another setting is added elsewhere? If possible, we should be pulling these values from some source of truth rather than hardcoded here

libs/designer/src/lib/core/queries/settingDefaults.ts
```diff
@@ -0,0 +1,125 @@
...
+ * during deserialization result in a single POST to the backend.
+ * Returns undefined if the host does not implement getSettingDefaults or the call fails.
+ */
+export const fetchSettingDefaults = async (
+  connectorId: string,
+  operationId: string,
+  supportedSettings: string[],
+  workflowKind?: string
+): Promise<Record<string, any> | undefined> => {
+  const service = OperationManifestService();
```
andrew-eldridge: nit: 'service' is vague here

libs/logic-apps-shared/src/designer-client-services/lib/standard/operationmanifest.ts
```diff
@@ -214,6 +214,38 @@ export class StandardOperationManifestService extends BaseOperationManifestServi
...
+          content: { settings: supportedSettings, workflowKind },
+        });
+      }
+
+      return await httpClient.post<Record<string, any>, { settings: string[]; workflowKind?: string }>({
+        uri: `${baseUrl}/operationGroups/${connectorName}/operations/${operationName}/settingDefaults`,
+        queryParameters: { 'api-version': apiVersion },
+        content: { settings: supportedSettings, workflowKind },
+      });
+    } catch {
```
andrew-eldridge: are there any exceptions which should be surfaced here rather than silently failing

libs/designer/src/lib/core/queries/settingDefaults.ts
```diff
@@ -0,0 +1,125 @@
...
+    const settingKey = key as keyof Settings;
+    const existing = merged[settingKey] as SettingData<unknown> | undefined;
+    if (!existing?.isSupported) {
+      continue;
+    }
+
+    if (isReadOnly) {
+      (merged as Record<string, unknown>)[settingKey] = { ...existing, value: defaultValue, readOnly: true };
+    } else if (settingKey === 'retryPolicy' && isDefaultRetryPolicy(existing.value)) {
+      (merged as Record<string, unknown>)[settingKey] = { ...existing, value: defaultValue, defaultHint: defaultValue };
```
andrew-eldridge, line 121: Is retryPolicy the only setting that uses defaultHint? I see it's included on the SettingData interface but only see usages for this setting

andrew-eldridge:
some of the copilot comments seem valid

libs/designer/src/lib/core/queries/settingDefaults.ts
JoaquimMalcampo: Removed the hardcoded set. Now derives "known" keys directly from the settings object produced by getOperationSettings ([Object.prototype.hasOwnProperty.call(settings, key)]), so the supported surface can't drift as settings are added/removed — the settings object is the single source of truth.

libs/designer/src/lib/core/queries/settingDefaults.ts
JoaquimMalcampo: Currently yes but will be extended if needed later.

libs/designer/src/lib/core/queries/settingDefaults.ts
JoaquimMalcampo: renamed

Elaina-Lee (approved):

PR #9265, opened 2026-06-09
Fix MCP UAMI persistence and dynamic-values identity threading

AbodeSaafan:
## Commit Type
- [x] fix - Bug fix

## Risk Level

## What & Why
Two MCP connection bugs in the Consumption designer (v1 and v2):

**Bug 1: MCP server node reverts to system-assigned after saving a UAMI connection.** The connection thunk re-derived the identity from the Logic App via `WorkflowService().getAppIdentity()` with a strict `equals(type, ''UserAssigned'')` check, dropping the user''s UAMI selection (and failing entirely for hybrid `SystemAssigned, UserAssigned` identities), so the canvas node fell back to system-assigned.

**Bug 2: `listMcpTools` was called without the UAMI identity.** `getListDynamicValues` hardcoded the identity to `undefined`, so both the cache key and the service call omitted the user''s selected identity, causing tool listings to use the wrong auth.

**Bug 3 (v2-only): Custom MCP connectors using non-standard parameter-set names (e.g. `oauthMI`) bypassed the identity gate.** `createConnectionInternal.tsx` only built MSI connection properties when the parameter set was literally named `ManagedServiceIdentity`, so custom connectors lost the UAMI selection at creation time and the wizard had nothing to thread into `listMcpTools` or the canvas node.

## Fix
**Connection persistence (v1 + v2):**
- New `getManagedIdentityFromConnection` helper reads the stored identity from both connection-storage shapes (multi-auth `parameterValueSet.values.identity.value` and single-auth `parameterValues.identity`).
- `getConnectionPropertiesIfRequired` prefers the stored connection identity, falling back to the Logic App identity only when none is stored.

**Dynamic values threading (v1 + v2):**
- `getListDynamicValues` accepts an optional `identity` arg, includes it (lowercased) in the cache key, and forwards it to the service.
- `dynamicdata.ts` extracts the identity from `connectionReference.connectionProperties.authentication.identity` and threads it through.

**v2 MCP wizard:**
- `mcpToolWizard.tsx` captures `createdIdentity` from the create-connection payload and threads `selectedIdentity` through `listMcpTools` and `addOperation`, so the canvas node binds to the picked UAMI.
- `add.ts` accepts a `connectionIdentity` and writes it to the operation''s connection reference.
- `createConnectionInternal.tsx` detects the identity parameter dynamically (by `ConnectionParameterTypes.managedIdentity` type or `identitypicker` editor) instead of matching only on the legacy `ManagedServiceIdentity` parameter-set name, so custom connectors like `oauthMI` work.

**Service-layer defense-in-depth:**
- Consumption connector `_buildMcpAuthentication` accepts an `identityOverride` and the caller computes an `effectiveIdentity` priority chain that supports both flat (`parameters.identity`) and nested (`connectionProperties.authentication.identity`) payload shapes.

## Impact of Change
- **Users**: MCP server connections in the Consumption designer (v1 and v2) correctly persist the selected User-Assigned Managed Identity on the canvas node, and `listMcpTools` calls use the right identity. Custom MCP connectors with non-standard parameter-set names now work end-to-end.
- **Developers**: New `getManagedIdentityFromConnection` helper. `getListDynamicValues`, `_buildMcpAuthentication`, and `addOperation` gained optional identity args. Identity parameter detection in `createConnectionInternal` is now editor/type-driven.
- **System**: No perf or dependency impact.

## Test Plan
- [ ] E2E tests added/updated

13 new unit tests across both v1 and v2 (plus existing wizard and createConnection suites still green):
- `connections.spec.ts` (v1 + v2) - 5 tests each for the helper across both storage shapes
- `connector.spec.ts` (designer + designer-v2) - 3 tests for identity threading, cache-key separation, and case-insensitive cache reuse
- `connector.spec.ts` (logic-apps-shared) - 5 tests for `_buildMcpAuthentication` priority and caller identity selection
- `mcpToolWizard.spec.tsx` - existing tests updated for new `selectedIdentity` plumbing
- createConnection suite (59 tests) green after dynamic-detection change

Full targeted runs green locally.


libs/designer/src/lib/core/queries/connector.ts
```diff
@@ -119,8 +120,9 @@ export const getListDynamicValues = async (
       operationId.toLowerCase(),
       dynamicState.operationId?.toLowerCase(),
       getParametersKey({ ...dynamicState.parameters, ...parameters }),
+      identity ?? '',
     ],
```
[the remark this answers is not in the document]
AbodeSaafan: Good catch  normalized to lowercase in the cache key (a85b8e5) so case-variant ARM IDs collapse to the same entry. The service call still passes the raw `identity` to preserve the canonical ID. Added a test in `connector.spec.ts` that asserts a case-variant second call is a cache hit.

AbodeSaafan:
Found another root cause for the v2 issue you hit at https://localhost:4200/v2  pushed as `09a799c12`.

**Root cause:** Built-in MCP connections that use UAMI are returned by Azure in the `parameterValueSet.values.<key>.value` shape (with `parameterValueSet.name === 'managedServiceIdentity'`), not the flat `parameterValues` shape. The shared `consumption/connector.ts::getListDynamicValues` branch detection only inspected `parameterValues?.mcpServerUrl`, so UAMI connections fell through into the managed-MCP path and `listMcpTools` POSTed without identity. This explains why intercepting the request only worked *after* publishing  pre-publish design-time was occasionally serving the flat shape, but the persisted Azure state comes back in the `parameterValueSet` shape.

**Fix:** Added `_flattenConnectionParameters` helper that merges both shapes (`parameterValueSet` wins, fallback to `parameterValues` for unset keys) and use it for branch detection + `_buildMcpAuthentication`. Existing UAMI tests had a redundant `parameterValues` block alongside `parameterValueSet` that masked the bug  stripped those so the tests exercise the real Azure response shape. All 25 `connector.spec.ts` tests pass.

No changes needed to the v2 `queries/connector.ts` wrapper or `dynamicdata.ts`  both already thread the identity correctly. The issue was entirely in the shared service consumed by v1 and v2.

AbodeSaafan:
Pushed 313a51dcc with a fix for the v2 issue you reported (calling `listMcpTools` with the wrong identity even though the saved connection had a UAMI).

**Root cause**: `createManagedMcpConnection` writes the UAMI to `properties.parameterValues.authentication.identity`, but `getManagedIdentityFromConnection` and the consumption connector's `listMcpTools` resolver only checked the flat `parameterValueSet.values.identity.value` / `parameterValues.identity` paths. So selecting an existing UAMI-backed connection produced a request body with the connection ID but no `identity`.

**Fix**: added a third fallback that reads `parameterValues.authentication.identity`, gated on `authentication.type === ''ManagedServiceIdentity''` so we don''t misread other auth shapes. Applied in 3 sites for v1/v2 parity:
- `libs/designer-v2/src/lib/core/utils/connectors/connections.ts`
- `libs/designer/src/lib/core/utils/connectors/connections.ts`
- `libs/logic-apps-shared/src/designer-client-services/lib/consumption/connector.ts`

Added 5 tests (positive nested-shape + negative type-guard in v1/v2 helpers, plus a connector test mirroring the existing managed-MCP UAMI test with the nested shape). All 40 tests in the 3 specs pass.

**Out of scope**: the pre-publish 401 you saw when manually injecting the identity is a backend requirement (workflow must be published before `listMcpTools` will accept the call)  not something we can fix from the designer.

AbodeSaafan:
Pushed c5f3243 with the actual fix after reproducing locally.

**Root cause (corrected from earlier guess in this PR):** `Microsoft.Web/connections` resources do not persist the UAMI ARM id for managed-identity parameter sets. For a managed MCP connector with `parameterValueSet.name = "oauthMI"` the resource has `parameterValueSet.values = {}` and no `identity` field anywhere, so reading the connection back can never recover which UAMI was chosen. The only authoritative source is the create-connection payload, and the wizard's `updateConnectionInState` was a no-op that dropped it.

**Fix:**
- Capture identity from `CreatedConnectionPayload` in `mcpToolWizard`. Both shapes are read: `payload.authentication.identity` (legacy MI flow) and `payload.connectionProperties.authentication.identity` (multi-auth `ManagedServiceIdentity` parameter set used by managed MCP connectors). The blocking-finding from the prior rubber-duck pass was that only the first shape was captured ΓÇö managed MCP connectors land identity in the second.
- Fall back to `getManagedIdentityFromConnection` on the fetched full connection / list entry for shapes that do expose identity on the resource.
- Reset captured identity when the user switches to a different existing connection.
- Thread `connectionIdentity` through `addOperation` -> `initializeOperationDetails` -> `trySetDefaultConnectionForNode` so the dispatched `updateNodeConnection` carries the explicit `connectionProperties.authentication.identity` (matching the existing identity-aware path).

**Known limitation:** picking an existing UAMI connection from the table for a managed MCP connector still cannot recover the UAMI from the resource alone. A wizard-level identity dropdown would be needed for full coverage ΓÇö calling that out as a follow-up rather than expanding scope here.

**Out of scope here:** the "401 unless the workflow is published first" behavior is a save/binding concern on the run side, not designer state ΓÇö tracking separately.

All 243 designer-v2 tests in the related suites pass. Comments from previous review are addressed in-scope.

libs/designer-v2/src/lib/core/parsers/ParseReduxAction.ts
```diff
@@ -204,6 +197,30 @@ export function flattenWorkflowNodes(nodes: WorkflowNode[]): WorkflowNode[] {
...
+      continue;
+    }
+    references[key] = {
+      api: { id: conn.id },
+      connection: { id: conn.connectionId ?? '' },
+      connectionName: conn.connectionName ?? key,
+      ...(conn.connectionProperties ? { connectionProperties: conn.connectionProperties } : {}),
+      ...(conn.connectionRuntimeUrl ? { connectionRuntimeUrl: conn.connectionRuntimeUrl } : {}),
+      ...(conn.authentication ? { authentication: conn.authentication } : {}),
+      ...(conn.impersonation ? { impersonation: conn.impersonation } : {}),
```
takyyon: 🔴 **HIGH RISK — silent behavior change for non-MCP Consumption workflows**

This helper now preserves four optional fields (`connectionProperties`, `connectionRuntimeUrl`, `authentication`, `impersonation`) that the inline predecessor in `initializeGraphState` didn't preserve — it only set `api`, `connection`, and `connectionName`.

That's the correct shape for the MCP UAMI flow (and the reason this PR exists), **but it's a real behavior change for every Consumption workflow whose `$connections.value[*]` already includes any of these four fields**. Downstream code that previously got `undefined` and synthesized defaults will now see live values, which could:
- override or shadow values computed elsewhere in deserialization
- expose fields to selectors / dialogs that previously couldn't see them
- cause MSI auth to be wired on non-MSI connectors that happen to carry `authentication` in `$connections`

The new unit test (`buildConnectionReferencesFromConnectionsParameter.spec.ts`) exercises this helper in isolation only — no integration coverage confirming downstream deserialization still produces the expected output for non-MSI Consumption flows carrying these fields. CI bot also flagged `add.ts` and `connections.ts` for coverage drops in this same blast radius.

**Suggested mitigation (pick one):**
1. Add an integration test loading a non-MSI Consumption workflow whose `$connections.value` includes `authentication`/`impersonation`, asserting deserialization output matches pre-change behavior.
2. Scope the new field preservation specifically to the MCP case — only spread `connectionProperties` (the one actually needed), drop the other three until a real consumer surfaces.
3. At minimum, call this expansion out in the PR description so a reviewer focused on the MCP bug doesn't merge it without thinking through the broader Consumption surface.

This is the highest-risk untested change in the PR.

libs/designer-v2/src/lib/core/actions/bjsworkflow/add.ts
```diff
@@ -571,8 +585,24 @@ export const trySetDefaultConnectionForNode = async (
...
     const connection = preferredConnection ?? (await tryGetMostRecentlyUsedConnectionId(connectorId, connections)) ?? connections[0];
-    await ConnectionService().setupConnectionIfNeeded(connection);
-    dispatch(updateNodeConnection({ nodeId, connection, connector }));
+    const userAssignedIdentity =
+      preferredConnectionIdentity && preferredConnectionIdentity !== constants.SYSTEM_ASSIGNED_MANAGED_IDENTITY
+        ? preferredConnectionIdentity
+        : undefined;
+    await ConnectionService().setupConnectionIfNeeded(connection, userAssignedIdentity);
+    if (preferredConnectionIdentity) {
+      dispatch(
```
takyyon, line 597: 🟡 **MEDIUM — behavior fork on `preferredConnectionIdentity` should also gate on "connector accepts MSI"**

This branch wires `connectionProperties.authentication` (MSI shape) onto the dispatched node whenever `preferredConnectionIdentity` is truthy. Today only the MCP wizard threads `connectionIdentity` through, and it's always a managed-identity connector, so this works.

If a future caller passes `preferredConnectionIdentity` for a non-MSI connector (or passes a UAMI ARM id for a connector that doesn't accept MSI), the dispatched `updateNodeConnection` will carry spurious `authentication: { type: 'ManagedServiceIdentity', identity: ... }` on a node whose connector doesn't support it. The reducer won't crash, but the canvas node will show MSI auth wiring on a connector that can't use it.

**Suggested guard:**
```ts
const canUseManagedIdentity =
  isConnectionMultiAuthManagedIdentityType(connection, connector) ||
  isConnectionSingleAuthManagedIdentityType(connection);

if (preferredConnectionIdentity && canUseManagedIdentity) {
  // dispatch with connectionProperties + authentication
} else {
  dispatch(updateNodeConnection({ nodeId, connection, connector }));
}
```

Contains the change to the MSI flow it was designed for and protects against future misuse.

libs/designer-v2/src/lib/ui/panel/recommendation/browse/mcpToolWizard.tsx
```diff
@@ -175,6 +179,27 @@ export const McpToolWizard = () => {
...
 
+  // Identity captured from the create-connection form. Microsoft.Web/connections resources
+  // do not persist the UAMI ARM id, so for newly created UAMI connections the only
+  // authoritative source is the payload the create flow emits.
+  const [createdIdentity, setCreatedIdentity] = useState<string | undefined>(undefined);
+
+  // The connection list response strips parameterValues, so the cached entry has no identity info.
+  // Fetch the full connection resource so we can detect UAMI for connections whose resource does
+  // expose it (some shapes do — e.g. via parameterValues.identity).
+  const { data: fullConnection } = useConnectionResource(localConnectionId ?? '');
```
takyyon: 🟡 **LOW — loading-window race on `useConnectionResource`**

While this resolves (`isLoading: true`), `fullConnection` is `undefined` and `selectedIdentity` falls back to `getManagedIdentityFromConnection(listEntry)`. Per the comment above, the list entry has stripped parameter values, so `selectedIdentity` is briefly `undefined` even for connections where the full resource would have produced one.

The `useQuery` for `mcpTools` has `selectedIdentity` in its queryKey, so when it resolves the tools query refetches — that path is covered.

The hole is `addOperation` in `handleNext`: if the user clicks Next during the loading window, `selectedIdentity` is `undefined` and the wrong identity (or none) gets threaded to the canvas node. Race only — wouldn't repro on fast connections — but worth a one-line guard either:
- Disable the Next button while `isLoading`, or
- In `handleNext`, await the resource if not yet loaded before calling `addOperation`.

libs/designer-v2/src/lib/ui/panel/recommendation/browse/mcpToolWizard.tsx
```diff
@@ -175,6 +179,27 @@ export const McpToolWizard = () => {
...
+
+  // Identity (UAMI) for the selected connection — threaded to listMcpTools and to addOperation so
+  // the workflow node and the tools query both bind to the user-chosen identity. Prefer the
+  // create-form payload; fall back to whatever the connection resource exposes.
+  const selectedIdentity = useMemo(() => {
+    if (createdIdentity) {
+      return createdIdentity;
+    }
+    const listEntry = connectionsData?.find((c) => c.id === localConnectionId);
+    return getManagedIdentityFromConnection(fullConnection) ?? getManagedIdentityFromConnection(listEntry);
```
takyyon, line 199: 🟡 **MEDIUM — known UX gap silently returns `undefined`; user lands in broken state with no feedback**

This fallback chain returns `undefined` when:
- the user picked an existing UAMI-backed connection (so `createdIdentity` is empty), **and**
- the connection resource itself doesn't expose `identity` (per your comment above — `Microsoft.Web/connections` does not persist the UAMI ARM id for some parameter sets like `oauthMI`).

When that happens, `selectedIdentity` is `undefined`, `listMcpTools` is called without identity, the canvas node binds without identity, and the user has no visible signal the wizard is in a broken state. They click Next, get either a 401 from `listMcpTools` or a silently-misbinding node, and have no recovery path short of deleting and re-creating the connection.

You called this out in the PR thread:
> picking an existing UAMI connection from the table for a managed MCP connector still cannot recover the UAMI from the resource alone

Fair scope call for this PR, but the user-facing failure mode deserves a near-term mitigation:
- When `localConnectionId` is set, `fullConnection` has loaded (`!isLoading`), and `selectedIdentity` is still `undefined` for a connector that requires MSI, show a non-blocking warning row in the wizard: *"Identity for this connection cannot be recovered. Re-create the connection or pick a UAMI explicitly to continue."*
- Log a `LogEntryLevel.Warning` so this state is observable in telemetry.
- File a follow-up issue for the dropdown-level fix and link it from the wizard warning.

Without one of these, this regression is invisible to the user.

libs/logic-apps-shared/src/designer-client-services/lib/consumption/connector.ts
```diff
@@ -255,7 +257,29 @@ export class ConsumptionConnectorService extends BaseConnectorService {
     return undefined;
   }
 
-  private _buildMcpAuthentication(connectionProperties: Record<string, any>): Record<string, any> | undefined {
+  private _flattenConnectionParameters(connection: any): Record<string, any> {
```
takyyon: 🟡 **MEDIUM — generic name, MCP-specific precedence rule; document or rename**

`_flattenConnectionParameters` is a generic-sounding name with an MCP-specific precedence rule: `parameterValueSet.values` wins, `parameterValues` only fills unset keys.

That's correct for the MCP MSI flow, but:
- The precedence isn't called out in JSDoc — a future caller seeing "flatten" could reasonably expect the other direction (single-auth `parameterValues` is the user's most recent edit in some flows).
- Today it's consumed only by the MCP path in this file. If it gets adopted by another flow with overlapping keys, the silent precedence will surprise someone.

Either:
- Rename to `_flattenMcpConnectionParameters` (intent-encoding, signals scope), or
- Add a JSDoc spelling out: *"Multi-auth `parameterValueSet.values` takes precedence; `parameterValues` keys only fill gaps. Designed for MCP where these two shapes never carry conflicting user state."*

Cheap, prevents the next bug.

takyyon (commented):
## Principal Architect review — generalized observations

Inline comments above flag the per-file concerns. Below are lower-priority items and observations that don't have a clean inline anchor.

### CI is red on 4 checks

- `validate-pr` failed in 3 seconds (configuration / approval gate, not a test)
- `vscode-e2e (p43-customcode)` failed after ~23 min
- `vscode-e2e (p43-inlinejavascript)` failed after ~22 min
- `vscode-e2e-summary` failed

These don't look like flakes given the run lengths. Worth investigating whether they're unrelated infra noise or actual regressions on the wiring path before merge — particularly given the touch on `ParseReduxAction.ts` (deserialization is in the e2e blast radius for Consumption workflows).

### Coverage drops flagged by CI bot

- `libs/designer-v2/src/lib/core/actions/bjsworkflow/add.ts` — 11%
- `libs/designer-v2/src/lib/core/actions/bjsworkflow/connections.ts` — 37%

The new `connectionIdentity` thread through `addOperation` → `initializeOperationDetails` → `trySetDefaultConnectionForNode` has no direct test — only the wizard-level mock in `mcpToolWizard.spec.tsx`. The identity-dispatch branch in `trySetDefaultConnectionForNode` is the most fragile new code in the PR and deserves a dedicated test in `add.spec.ts` exercising `preferredConnectionIdentity` set vs unset (would also cover the inline guard ask).

### No E2E test added

Test plan has `[ ] E2E tests added/updated` unchecked. **Four root causes across seven commits in a multi-layer wiring fix is exactly the shape that justifies one.** The next regression in this area will be the second time this bug is fixed here. Strongly recommend at least one e2e covering: open MCP wizard → create new UAMI connection → save → verify the canvas node binds to the picked UAMI.

### Stale state risk in `mcpToolWizard`

`createdIdentity` is reset on `handleConnectionSelect` (good), but **not** when the wizard re-opens with the same connection ID from a prior session. If a user creates a UAMI connection, saves, reopens the wizard pointed at the same connection, then proceeds — they could inherit the stale captured identity from the prior session. Safer to clear on a `useEffect` when the wizard step transitions back to the create step.

### Pre-existing rough edges (not blocking)

- **`_buildMcpAuthentication` multi-UAMI ambiguity:** when the Logic App has multiple UAMIs and no override/stored identity, the fallback picks the first key from `userAssignedIdentities` arbitrarily. Pre-existing behavior, but worth a follow-up — `SystemAssigned, UserAssigned` apps with 2+ UAMIs are a real customer config and they have no deterministic say in which one is chosen here.

### Documentation gap

The actual root cause of the v2 issue (`Microsoft.Web/connections` resources **do not persist the UAMI ARM id** for managed-identity parameter sets like `oauthMI`) is documented only in this PR's discussion thread. Add an inline comment at the consumer sites (`mcpToolWizard.tsx`, `createConnectionInternal.tsx`, or wherever a future contributor lands first) noting:
> Some `Microsoft.Web/connections` parameter sets (e.g. multi-auth `oauthMI`) do not persist the UAMI ARM id in the resource. The create-connection payload is the only authoritative source. Reading the connection back will produce `undefined` for these cases.

Without this, the next person debugging this will repeat the same 4-round-trip discovery process.

### What's working well (worth preserving)

- New `getManagedIdentityFromConnection` is well-scoped, well-tested across all 3 storage shapes, and includes a negative guard for non-MSI auth types.
- Editor/type-driven parameter detection in `createConnectionInternal.tsx` (instead of name-matching on `'ManagedServiceIdentity'`) is the right pattern and will pay forward for any future custom MSI connector.
- The `_flattenConnectionParameters` helper consolidates the dual-shape dance that was previously inlined — modulo the rename/JSDoc ask in the inline comment.
- v1/v2 parity is maintained explicitly; helper is mirrored in both `libs/designer` and `libs/designer-v2` consistent with the rest of the repo.
- Defense-in-depth (read helper → cache key → service call → wizard capture) means a single layer failing won't silently regress the whole path.

---

**Verdict: Y-with-fixes.** Strong fix for the named bugs; the 7-commit iteration shows real debugging rigor. Resolve CI before merge; address Finding #1 (silent behavior change in `ParseReduxAction.ts`) and Finding #4 (silent UX failure in `mcpToolWizard.tsx`) before merge or in a same-sprint follow-up; #2, #3, and #5 are good cleanups defensible to land separately.

libs/designer-v2/src/lib/core/parsers/ParseReduxAction.ts
AbodeSaafan: Addressed. `ParseReduxAction.ts::buildConnectionReferencesFromConnectionsParameter` now preserves **only** `connectionProperties` from `$connections.value` entries — the four legacy speculative fields (`connectionRuntimeUrl`, `authentication`, `impersonation`, plus pass-through of any extras) are dropped. Spec `buildConnectionReferencesFromConnectionsParameter.spec.ts` updated to assert this (positive: only `connectionProperties` preserved; negative: omits when absent; explicit non-pass-through of legacy fields). 5/5 passing.

libs/designer-v2/src/lib/core/actions/bjsworkflow/add.ts
AbodeSaafan: Addressed. `trySetDefaultConnectionForNode` in `add.ts` now gates `preferredConnectionIdentity` on connector MSI capability (`supportsManagedIdentity`) so we don't silently bind an identity to non-MSI connectors. `add.spec.ts` extended with positive/negative MSI-guard tests covering: `preferredConnectionIdentity` set + connector MSI-capable, set + non-MSI (no-op), and unset (no-op).

libs/designer-v2/src/lib/ui/panel/recommendation/browse/mcpToolWizard.tsx
AbodeSaafan: Addressed. `isFetchingFullConnection` (computed as `!!localConnectionId && !createdIdentity && isFullConnectionLoading`) disables the Next button while `useConnectionResource` is in-flight, so we can't thread an `undefined` identity into `addOperation` on a stale render. Covered by the new mcpToolWizard tests for the loading-state branch.

libs/designer-v2/src/lib/ui/panel/recommendation/browse/mcpToolWizard.tsx
AbodeSaafan: Partially addressed. We now surface the gap instead of failing silently: `needsManagedIdentityWarning` shows a non-blocking inline warning, disables Next, and emits a `LoggerService` Warning with `connectionId`/`connectorId`. A full inline-dropdown re-pick UX (so the user can resolve the gap in-wizard) is deferred — scope-limited per the PR's stated bounds, tracking as a follow-up so it doesn't expand this PR further.

libs/logic-apps-shared/src/designer-client-services/lib/consumption/connector.ts
AbodeSaafan: Addressed. Renamed to `_buildMcpAuthentication` and added a JSDoc block documenting the precedence: explicit override > stored `connectionProperties.identity` > Logic App UAMI fallback (with the new multi-UAMI warning). The MCP-specific rule is now scoped by name.

AbodeSaafan:
Thanks for the thorough review, @takyyon! Round-2 commit `7e6a7793d` addresses items **A, B, D, F**:

- **A — stale `createdIdentity` reset.** Added a `useEffect` in `mcpToolWizard.tsx` that resets `createdIdentity` on entry to the CREATE_CONNECTION step, so a back-nav from a stale create flow can't bleed identity into a subsequent pick.
- **B — UAMI doc coverage.** Tightened the identity comments in `mcpToolWizard.tsx` (kept them concise per session style).
- **D — E2E coverage.** Added `__mocks__/workflows/AgentWithMcpUami.json` (Consumption-shaped, UAMI in `$connections.value`) wired into `LogicAppSelector`, plus a round-trip serialization test in `e2e/designer/mcpSerialization.spec.ts` asserting the UAMI threads through `connectionReferences.mcp.connectionProperties.authentication.identity`.
- **F — multi-UAMI ambiguity.** `_buildMcpAuthentication` now emits a `LoggerService` Warning when more than one UAMI is configured on the Logic App and no override is set; SAMI and SystemAssigned+UserAssigned hybrids retain the existing fall-through (no `identity` → backend uses SAMI). Covered by 4 new test cases in `connector.spec.ts`.

**E — CI.** The four reds left on the PR are external/unrelated:
- `validate-pr` waits on the maintainer-applied `external-approved` label (external-contributor gate).
- The three `vscode-e2e p43-*` shards are ExTester runs unrelated to designer-v2/Consumption changes in this PR.

Inline threads on items 1–5 have individual replies confirming what was addressed (and what was deferred — the inline-dropdown re-pick UX from #4 is tracked as a follow-up). Ready for re-review when you have a moment.

takyyon (approved):

PR #9289, opened 2026-06-15
feat(vscode): Add workflow/webview validation and E2E launcher migration

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why
**Part 2 of 2** (depends on #9330). Adds webview validation features and E2E test infrastructure for the VS Code extension.

Key features:
- **Workflow name collision detection**: Block workflow names that already exist in the selected logic app project
- **Reserved folder name validation**: Prevent workflow names like `Artifacts`, `lib`, `.vscode`, `workflow-designtime` etc. (case-insensitive)
- **Create Project folder gate**: Block project creation when folder name collides with existing workspace folders
- **resetState race condition fix**: Remove `resetState()` from `CreateLogicApp` component that was wiping `existingFolders` data
- **E2E behavior tests**: 5 new reserved workflow name tests in `createWorkspace.behavior.test.ts`
- **E2E infrastructure**: Runtime dependency hardening in `run-e2e.js`, func+gozip permission repair

## Impact of Change

- **Users**: Better validation prevents accidental overwrite of reserved/existing workflow folders during creation
- **Developers**: New E2E behavior test pattern for webview validation
- **System**: No new dependencies; validation logic is client-side only

## Test Plan

- [x] E2E tests added/updated
- [ ] Tested in: VS Code on Windows, all 18 vscode-e2e CI jobs passed including `p41b-createworkspace-behavior` (14m25s)

## Contributors

@lambrianmsft
@Copilot

## Screenshots/Videos

N/A


apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
```diff
@@ -615,35 +625,47 @@ export function startDesignTimeProcess(
...
+ * Runs taskkill and waits for it to complete so file locks are released
+ * before subsequent build steps.
+ */
+function execTaskkill(pid: number): Promise<void> {
+  return new Promise((resolve) => {
+    cp.exec(`taskkill /pid ${pid} /t /f`, () => {
+      resolve();
+    });
+  });
+}
```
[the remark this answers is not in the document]
lambrianmsft: Update execTaskkill to accept number | string (taskkill accepts numeric strings) and keep the implementation otherwise unchanged.

apps/vs-code-designer/src/test/ui/designerHelpers.ts
```diff
@@ -179,17 +180,24 @@ function getFuncCoreToolsCandidatePaths(): string[] {
 
 function getFuncCoreToolsPath(): string {
   const candidates = getFuncCoreToolsCandidatePaths();
-  return candidates.find((candidate) => fs.existsSync(candidate)) ?? candidates[0];
+  const funcCandidates = candidates.filter((candidate) => path.basename(candidate).toLowerCase().startsWith('func'));
```
andrew-eldridge: wouldn't all values from getFuncCoreToolsCandidatePaths start with 'func'?

apps/vs-code-designer/src/test/ui/designerHelpers.ts
```diff
@@ -179,17 +180,24 @@ function getFuncCoreToolsCandidatePaths(): string[] {
...
+  return funcCandidates.find((candidate) => fs.existsSync(candidate)) ?? funcCandidates[0];
 }
 
 function assertFuncCoreToolsExecutable(context: string): void {
-  const funcBinaryPath = getFuncCoreToolsPath();
   ensureRuntimeDependencyExecutablePermissions();
-  if (!fs.existsSync(funcBinaryPath)) {
+  const funcBinaryPath = getFuncCoreToolsPath();
+  const existingCandidates = getFuncCoreToolsCandidatePaths()
+    .filter((candidate) => path.basename(candidate).toLowerCase().startsWith('func'))
```
andrew-eldridge: same comment here

apps/vs-code-designer/src/test/ui/designerHelpers.ts
andrew-eldridge: why use fs.constants.F_OK instead of fs.constants.X_OK for win32 if we're checking that file is executable? Wouldn't X_OK be the execute permission (not changed in this pr but isExecuatableFile is used)

apps/vs-code-designer/src/test/ui/designerHelpers.ts
```diff
@@ -179,17 +180,24 @@ function getFuncCoreToolsCandidatePaths(): string[] {
...
+    .filter((candidate) => fs.existsSync(candidate));
+  if (existingCandidates.length === 0) {
     throw new Error(`[depValidation] func binary not found after ${context}: ${funcBinaryPath}`);
   }
-  if (!isExecutableFile(funcBinaryPath)) {
-    throw new Error(`[depValidation] func binary exists but is not executable after ${context}: ${funcBinaryPath}`);
+
+  for (const candidate of existingCandidates) {
```
andrew-eldridge, line 192: should we assert that existingCandidates length is 1 before getting here? Is it possible for there to be multiple matches?

apps/vs-code-designer/src/test/ui/run-e2e.js
```diff
@@ -1027,11 +1028,92 @@ async function main() {
     };
   };
 
+  const runtimeExecutableAccessMode = () => (process.platform === 'win32' ? fs.constants.F_OK : fs.constants.X_OK);
+
+  const isRuntimeExecutable = (binaryPath) => {
```
andrew-eldridge: this appears duplicated from designerHelpers.ts, is it possible to de-dupe?

apps/vs-code-designer/src/test/ui/run-e2e.js
```diff
@@ -1609,6 +1676,51 @@ async function main() {
...
+    const depsRoot = path.join(os.tmpdir(), 'la-e2e-test', 'lsp-eperm-dependencies');
+    fs.rmSync(depsRoot, { recursive: true, force: true, maxRetries: 3, retryDelay: 1000 });
+    fs.mkdirSync(depsRoot, { recursive: true });
+
+    for (const [name, sourceDir] of requiredSourceDirs) {
+      fs.cpSync(sourceDir, path.join(depsRoot, name), { recursive: true });
+    }
+
+    for (const stalePath of [
+      path.join(depsRoot, 'LSPServer'),
```
andrew-eldridge: is there a list of dependencies somewhere we can use as source of truth here instead of manually specifying each one- this seems like it can easily fall out of sync with true dependencies list

apps/vs-code-designer/src/test/ui/runHelpers.ts
```diff
@@ -25,10 +25,35 @@ import * as assert from 'assert';
 import * as fs from 'fs';
 import * as http from 'http';
 import * as https from 'https';
+import * as os from 'os';
 import * as path from 'path';
 import { type Workbench, WebView, By, type WebDriver, VSBrowser, Key, EditorView, BottomBarPanel } from 'vscode-extension-tester';
 import { sleep, captureScreenshot, dismissAllDialogs, clearBlockingUI, focusEditor } from './helpers';
 
+function getFuncCoreToolsDiagnosticPaths(): string[] {
+  const funcToolsRoot = path.join(os.homedir(), '.azurelogicapps', 'dependencies', 'FuncCoreTools');
```
andrew-eldridge, line 36: isn't the dependencies path configurable? this looks like it will fail if set to something other than default

lambrianmsft: This is only used for e2e testing, where the dependency path wouldn't be set to something else other than the default!

apps/vs-code-designer/src/test/ui/designerHelpers.ts
lambrianmsft: Good catch - you are right. getFuncCoreToolsCandidatePaths() only returns paths whose basename is in the executableNames set (func or func.exe), so the .startsWith('func') filter was completely redundant. Removed in d2e3c65.

apps/vs-code-designer/src/test/ui/designerHelpers.ts
lambrianmsft: Same fix applied - redundant filter removed in d2e3c65.

apps/vs-code-designer/src/test/ui/run-e2e.js
lambrianmsft: Good point. Extracted the stale path names into a top-level constant LSP_STALE_ARTIFACT_NAMES with a JSDoc comment mapping each name to its source-of-truth location (languageServerProtocol.ts lines 10-12 and constants.ts line 39). Since run-e2e.js is plain JS it cannot import from TS source, but the doc reference makes it easy to keep in sync. Done in d2e3c65.

apps/vs-code-designer/src/test/ui/designerHelpers.ts
lambrianmsft: Good question. On Windows, Node.js docs state: 'fs.access() does not fully support X_OK, which is treated as F_OK.' Windows has no execute-permission bit - executability is determined by file extension (.exe, .cmd). Using F_OK explicitly is more honest that we are only checking file existence (which combined with .exe extension = executable). Added a detailed comment in the new sharedDepUtils.js explaining this (d2e3c65).

apps/vs-code-designer/src/test/ui/designerHelpers.ts
lambrianmsft: Multiple matches are expected and an assertion would be wrong. FuncCoreTools installs copies under multiple subdirectories (FuncCoreTools/func, in-proc8/func, in-proc6/func) for different .NET in-process versions. The extension picks one at runtime based on project config. We validate ALL existing binaries have execute permissions so no matter which path gets selected, it works. Added a comment explaining this in d2e3c65.

apps/vs-code-designer/src/test/ui/runHelpers.ts
lambrianmsft: The path is configurable via the autoRuntimeDependenciesPath VS Code setting, but E2E tests always set it to the default (~/.azurelogicapps/dependencies) via run-e2e.js settings injection (line 1210). This diagnostic function is only called within E2E test runs, never in production. Added a comment clarifying this in d2e3c65.

apps/vs-code-designer/src/test/ui/run-e2e.js
lambrianmsft: De-duped. Extracted the shared logic into apps/vs-code-designer/src/test/ui/sharedDepUtils.js - a plain CommonJS module importable by both run-e2e.js (plain Node launcher) and designerHelpers.ts (TypeScript compiled by tsup). Both now import isExecutableFile from the same source. Done in d2e3c65.

lambrianmsft: Updated to be .ts modules now!

ccastrotrejo (approved):

PR #9296, opened 2026-06-19
fix(designer): Serialize non-managed MCP server connections to connections.json

Elaina-Lee:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

Non-managed MCP server connections (auth types: None, Basic, Client Certificate) were not included in the deployment payload's `connections.json`, causing the validate API to fail with `InvalidAgentWorkflow`. Managed MCP connections (e.g., Kusto Query MCP Server) worked fine because they use ARM resource IDs.

`getConnectionsDataToSerialize` in the MCP utils serializer only processed ARM resource IDs via `isArmResourceId()`. Non-managed MCP connections use non-ARM IDs (`/connectionProviders/mcpclient/connections/<name>`) and were skipped entirely. Additionally, `convertConnectionsDataToReferences` did not handle `agentMcpConnections`, so MCP connections loaded from existing `connections.json` were never converted to Redux connection references.

IcM: 21000001042369

## Impact of Change

- **Users**: Non-managed MCP server connections (None, Basic, Client Certificate auth) now serialize correctly to `connections.json`, fixing workflow validation failures.
- **Developers**: New exported `isMcpConnectionReference()` utility available in both designer v1 and v2 serializers.
- **System**: No performance impact. Adds one conditional branch and one async fetch per MCP connection during serialization.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Standalone designer with ARM mode, non-managed MCP server creation flow

## Screenshots/Videos

image
image
image


AbodeSaafan (approved):
from a non ux engineer, it looks fine. We should get css to also test and confirm but we had a clear repro. Only concern is if we should get the "apikey" vs "key" fix in with this as well to fully unblock the feature

Elaina-Lee:
> from a non ux engineer, it looks fine. We should get css to also test and confirm but we had a clear repro. Only concern is if we should get the "apikey" vs "key" fix in with this as well to fully unblock the feature

Communicated offline, this fix (apikey and key fix) has been verified here - that the fix had been merged.

PR #9376, opened 2026-07-08
fix(vscode): harden runtime dependency validation (integrity checks, SHA256 verification, metadata timeout)

lambrianmsft:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

**TL;DR:** Hardens the VS Code extension against runtime-dependency corruption and startup hangs, so a corrupt/partially-deleted NodeJs or FuncCoreTools install is detected and repaired -- without re-slowing activation or redownloading when the on-disk copy is actually healthy.

**Why:** Users reported the extension getting stuck on **"Validating Runtime Dependency: NodeJS"** (blocking the designer from loading), silent failures when Function Host `node`/DLL files went missing (surfacing only as a cryptic `func` host crash: `The system cannot find the file specified` / `Failed to start Worker Channel`), and endless redownloads of NodeJs/FuncCoreTools on every restart. Only the extension bundle was integrity-checked; the other runtime dependencies had no health check.

This PR adds four layers of bounded, verifiable integrity checking:

1. **Metadata request timeout** -- version/metadata resolution is now bounded so a stalled network call can no longer hang activation on "Validating Runtime Dependency: NodeJS".
2. **CDN SHA256 download verification** -- downloaded archives are checked against the publisher-provided SHA256 (nodejs.org `SHASUMS256.txt`, func-core-tools `.zip.sha2`) before extraction; a corrupt download is caught and retried instead of extracted.
3. **On-disk integrity manifest (size + existence)** -- after a successful install we persist a `.logicapps-integrity.json` manifest of every extracted file with its size. On startup each dependency is validated against its manifest; a missing file or size mismatch triggers a targeted reinstall, while a healthy install with a valid manifest is **not** redownloaded.
4. **Bounded per-file SHA256 content verification** -- the manifest also stores a SHA256 for only the **critical executables** (`node.exe`/`node`, `func.exe`/`func`/`func.dll`), captured at install right after the archive''s CDN SHA256 was verified (so they are transitively CDN-anchored). On startup those few files are re-hashed to catch silent same-size corruption that a size-only check misses.

Whole-tree hashing was deliberately avoided: publishers hash the compressed archive (not the extracted tree, and extraction isn''t reversible), the Functions Core Tools release publishes zero per-file hashes, and re-hashing ~150 MB Function Host + ~70 MB NodeJs on every startup would risk reintroducing the exact activation hang being fixed.

## Impact of Change

- **Users**: No change for healthy installs (they are no longer needlessly redownloaded). A corrupt/incomplete NodeJs or FuncCoreTools install is now automatically detected and repaired instead of hanging activation or crashing the Function Host at debug time.
- **Developers**: `writeDependencyIntegrityManifest` / `verifyDependencyIntegrity` are now `async` (awaited at 3 call sites). New telemetry signals for observability: `<dep>IntegrityResult` (`passed` / `manifest-missing` / `size-mismatch` / `sha256-mismatch`), `<dep>IntegrityHashedFiles`, `<dep>IntegrityMismatchFile`.
- **System**: Adds a bounded metadata timeout and SHA256 verification on download; startup integrity check is size/existence for the full tree plus SHA256 for a handful of critical binaries only, so activation cost stays low. No workflow/designer product behavior changes.

## Test Plan

- [ ] E2E tests added/updated 
- [ ] Tested in: 

Build: `pnpm run build:extension` succeeds (esbuild). Biome clean on all changed files.

## Contributors

- Author: @lambrianmsft
- Co-author: Copilot (`223556219+Copilot@users.noreply.github.com`)

## Screenshots/Videos

Not applicable -- CI/dependency-infra change with no visual UI surface. Verification evidence is unit-test output (116 passing) and local extension-host logs showing integrity-check pass/reinstall decisions.


apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -404,6 +440,343 @@ export function getFunctionCoreToolsBinariesReleaseUrl(version: string, osPlatfo
...
+ */
+export async function computeFileSha256(filePath: string): Promise<string> {
+  return await new Promise<string>((resolve, reject) => {
+    const hash = createHash('sha256');
+    const stream = fs.createReadStream(filePath);
+    stream.on('error', reject);
+    stream.on('data', (chunk: Buffer) => hash.update(Uint8Array.from(chunk)));
+    stream.on('end', () => resolve(hash.digest('hex')));
+  });
+}
```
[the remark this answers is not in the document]
lambrianmsft: Good catch — fixed in aaf0973f0. computeFileSha256 now passes the Buffer chunk straight to hash.update(chunk) instead of copying it into a new Uint8Array per chunk, dropping the extra allocation/CPU on large-archive hashing.

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -404,6 +440,343 @@ export function getFunctionCoreToolsBinariesReleaseUrl(version: string, osPlatfo
...
+        ext.outputChannel.appendLog(
+          localize(
+            'integrityFileMissing',
+            '{0} on-disk integrity check FAILED: missing file "{1}". Scheduling reinstall.',
+            dependencyName,
+            file.path
+          )
+        );
+        return false;
+      }
```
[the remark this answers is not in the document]
lambrianmsft: Fixed in aaf0973f0. The verifier now checks stat.isFile() and fails fast (new 
ot-a-file integrity result → reinstall) when a recorded entry has been replaced by a directory/symlink, instead of only comparing size. Added a regression test and updated the existing statSync stubs to model isFile().

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -102,6 +105,34 @@ export async function downloadAndExtractDependency(
...
+        }
+        if (fs.existsSync(targetFolder)) {
+          fs.rmSync(targetFolder, { recursive: true, force: true });
+        }
+      } catch {
+        // Best-effort cleanup; ignore secondary errors.
+      }
+      throw new Error(errorMessage);
+    }
+    executeCommand(ext.outputChannel, undefined, 'echo', `Verified ${dependencyName} download checksum.`);
```
andrew-eldridge: is this equivalent to ext.outputChannel.appendLog? if so can we use that

apps/vs-code-designer/src/app/utils/binaries.ts
andrew-eldridge: nit: 'downloadFileWithVerification' implies that the verification is done here, but the checksum check isn't done until later. Should it be moved into here, or have this renamed?

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -404,6 +440,356 @@ export function getFunctionCoreToolsBinariesReleaseUrl(version: string, osPlatfo
...
+ * @param {string} dependencyName - The dependency name (e.g. NodeJs, FuncCoreTools).
+ * @returns {Promise<string | undefined>} The expected SHA256 hex digest, or undefined to skip verification.
+ */
+export async function resolveExpectedDependencySha256(
+  context: IActionContext,
+  downloadUrl: string,
+  dependencyName: string
+): Promise<string | undefined> {
+  if (dependencyName === nodeJsDependencyName) {
+    const match = downloadUrl.match(/nodejs\.org\/dist\/v([^/]+)\/(.+)$/);
```
andrew-eldridge: regex shouldn't be needed here since we know which dependency is being downloaded, just need to get that info to this function

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -404,6 +440,356 @@ export function getFunctionCoreToolsBinariesReleaseUrl(version: string, osPlatfo
...
+/**
+ * Writes an on-disk integrity manifest into a freshly installed dependency folder, recording every
+ * file's size plus a SHA256 content hash for the primary executables. Best-effort: a failure to write
+ * the manifest (or to hash an individual file) is logged but never fails the install.
+ * @param {IActionContext} context - Command context, used for telemetry.
+ * @param {string} targetFolder - The dependency folder that was just extracted.
+ * @param {string} dependencyName - The dependency name (e.g. NodeJs, FuncCoreTools).
+ * @returns {Promise<void>} Resolves once the manifest has been written (or the failure logged).
+ */
+export async function writeDependencyIntegrityManifest(
```
andrew-eldridge: this may be a major change to the approach, but would it make sense to just verify integrity once on download by comparing checksum for the zip, and if that is correct then we assume from then on that the dependency is valid, until a new one is downloaded? It looks like we store hashes for each individual file to check on startup, which seems excessive to counter the scenario where users intentionally modify their dependency files

apps/vs-code-designer/src/constants.ts
```diff
@@ -277,6 +277,14 @@ export const e2eStrictDependencyValidationSettingKey = 'e2eStrictDependencyValid
 export const useExperimentalExtensionBundleSettingKey = 'useExperimentalExtensionBundle';
 export const experimentalExtensionBundleSourceUriSettingKey = 'experimentalExtensionBundleSourceUri';
 export const experimentalExtensionBundleVersionSettingKey = 'experimentalExtensionBundleVersion';
+// Max time (ms) to wait for a dependency metadata/HTTP response (e.g. latest version lookups)
```
andrew-eldridge: nit: comments can be shortened/removed in some places

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: Good call — switched the checksum-verified log (and the new download-retry log) to ext.outputChannel.appendLog with localized strings instead of spawning echo. Done in 9c676eb3e.

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: Agreed the name was misleading. Renamed it to downloadFileWithTransportVerification since it only does the transport-level (Content-Length/Content-MD5) check. I kept the SHA256 archive verification in downloadAndExtractDependency on purpose — it needs to clean up both the temp and target folders on mismatch, and folding it into the vscode-free transport wrapper would entangle those concerns. 9c676eb3e.

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: Removed the regex. getNodeJsSha256 now takes the download URL and derives the artifact name via path.basename and the SHASUMS256.txt URL by swapping the last path segment (same /dist/vX/ dir), so no version/filename is re-parsed out of the URL. 9c676eb3e.

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: Great question — this pushed us to simplify. I dropped the per-file SHA256 hashing entirely; the on-disk manifest now records path + size only, and verify checks existence/size. Reasoning: the bug this feature actually targets is files going *missing/truncated* from the install after download (a deleted Function Host DLL or node.exe), which crashes the host at startup. A download-time zip checksum alone can't catch that because it's only evaluated once at download and the extracted tree is never re-validated against it. So we keep a cheap startup existence+size check (catches the real scenario, no re-download needed) but remove the per-file content hashing you flagged as excessive. The zip/archive SHA256 is still verified once at download time before extraction. 9c676eb3e.

apps/vs-code-designer/src/constants.ts
lambrianmsft: Trimmed both to a single concise line each. 9c676eb3e.

andrew-eldridge (approved):

PR #9410, opened 2026-07-16
fix(vscode): Data Mapper Test Panel 400 From NetFxWorker not launching

lambrianmsft:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

The Data Mapper **Test map** panel returns `400 - BadRequest: The .NET framework worker could not be found.` (regressed after #7785; last known-good extension was 5.230.15). Test map runs its XSLT transform on the extension bundle's `Microsoft.Azure.Workflows.Functions.NetFxWorker`, which the Functions runtime only spawns when the host runs in-process .NET 8 (`FUNCTIONS_WORKER_RUNTIME=dotnet` **and** `FUNCTIONS_INPROC_NET8_ENABLED=1`). The Data Mapper reuses the designer's shared `workflow-designtime` host, which was started as `node`, so the NetFxWorker never spawned. This configures the single shared design-time host as dotnet + inproc-net8 so the designer and Data Mapper share one correctly-configured host.

## Impact of Change

- **Users**: Data Mapper Test map works again (returns transformed output instead of a 400).
- **Developers**: Design-time `local.settings.json` schema now sets `FUNCTIONS_WORKER_RUNTIME=dotnet` + `FUNCTIONS_INPROC_NET8_ENABLED=1`.
- **System**: The shared `workflow-designtime` host now runs in-process .NET 8 instead of node.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in:  VS Code extension — shared host starts dotnet in-process and spawns NetFxWorker; DM Test map returns 200; designer operation discovery / dynamic data / token picker still work

## Contributors

@lambrianmsft

## Screenshots/Videos

N/A

### DIFF
--- a/apps/vs-code-designer/src/app/commands/dataMapper/FxWorkflowRuntime.ts
+++ b/apps/vs-code-designer/src/app/commands/dataMapper/FxWorkflowRuntime.ts
@@ -8,6 +8,8 @@ import {
   appKindSetting,
   designTimeDirectoryName,
   designerStartApi,
+  functionsInprocNet8Enabled,
+  functionsInprocNet8EnabledTrue,
   hostFileContent,
   hostFileName,
   localSettingsFileName,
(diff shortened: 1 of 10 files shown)


andrew-eldridge (approved):
was there a reason we left worker runtime as node for design time before? I thought there was some issue with switching to .net

lambrianmsft:
> was there a reason we left worker runtime as node for design time before? I thought there was some issue with switching to .net

I believe we had debugging issues previously but now I have verified it's not an issue.

PR #9430, opened 2026-07-20
fix(vscode): Speed up VS Code startup dependency validation and reduce multi-project log noise

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix
- [x] perf - Performance improvement

## Risk Level

## What & Why

VS Code extension startup dependency validation could take 30s+ and produced redundant, interleaved output for multi–Logic App workspaces. This PR speeds up startup and reduces log noise. All changes are scoped to `apps/vs-code-designer`.

- **Throttle dependency update checks to once per 24h** — `dependencyUpdateCheck.ts` records the last successful validation timestamp and skips expensive network/binary re-validation on subsequent launches within the window, while still catching genuinely missing/corrupted binaries.
- **Parallelize startup dependency validation** — dotnet, func, and Node version/binary checks now run concurrently instead of serially, and each validator uses the shared `binariesExist` from `binaries.ts` rather than a bespoke check.
- **Serialize design-time / runtime port allocation** — `portReservation.ts` reserves ports atomically so concurrent projects no longer collide on ports each thought were free.
- **Cache extension bundle health per session** — `bundleFeed.ts` verifies on-disk bundle integrity once per session and short-circuits subsequent design-time host launches; the "already verified" skip is logged at most once.
- **Collapse multi-project startup log noise** — `validateProjectArtifacts.ts` low-level regenerate helpers are now silent and return status; each project emits exactly one consolidated status line (valid / regenerated-what / failed). Preserves the Node design-time worker fallback behavior.

## Impact of Change

- **Users**: Faster VS Code extension startup (repeat launches skip the ~30s revalidation); far less noisy, one-line-per-project startup output in multi–Logic App workspaces; no more port collisions across concurrent projects.
- **Developers**: `regenerateLocalSettings`/`regenerateDesignTimeDirectory` now return status objects; new exported `ensureProjectRootArtifacts`; validators share `binariesExist` from `binaries.ts`. No behavior change to what a valid project looks like.
- **System**: Parallel + throttled validation and per-session bundle-health caching reduce redundant network/disk work at startup. No new runtime dependencies.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: local `apps/vs-code-designer` unit suites

Validation run locally:
- `npx vitest run` for all affected suites — `validateProjectArtifacts` (52/52), `startDesignTimeApi` + `bundleFeed` (121/121), `portReservation` + `dependencyUpdateCheck` + both `getTasks` (35/35) — all green.
- Added/updated coverage for `getTasks`/`writeTasksJson`, `portReservation`, `dependencyUpdateCheck`, `bundleFeed`, `validateProjectArtifacts`, and the dotnet/func/Node version validators, including that async binary checks are properly awaited.
- `npx tsup` type-check — 0 errors.
- `npx biome check` — clean.

## Contributors

@lambrianmsft

## Screenshots/Videos

N/A — no visual/UI changes.


apps/vs-code-designer/src/app/utils/dependencyUpdateCheck.ts
```diff
@@ -0,0 +1,46 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { ext } from '../../extensionVariables';
+
+/**
+ * Key used to persist the timestamp (epoch ms) of the last time we checked whether the
+ * runtime dependencies (Node.js, Functions Core Tools, .NET SDK) had newer versions available.
+ */
+const lastDependencyUpdateCheckKey = 'azureLogicAppsStandard.lastDependencyUpdateCheck';
```
andrew-eldridge: nit: this and below const could be moved to constants.ts

apps/vs-code-designer/src/app/utils/codeless/validateProjectArtifacts.ts
```diff
@@ -443,27 +411,139 @@ export async function regenerateDesignTimeDirectory(context: IActionContext, pro
...
+function collectChangedArtifacts(summary: {
+  hostChanged: boolean;
+  localSettingsChanged: boolean;
+  addedSettings: string[];
+  designTimeHostChanged?: boolean;
+  designTimeSettingsChanged?: boolean;
+}): string[] {
+  const changed: string[] = [];
+  if (summary.hostChanged) {
+    changed.push('host.json');
```
andrew-eldridge: is it possible to avoid hard-coding the names here and instead aggregate the updated names/paths as the updates are done?

apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
```diff
@@ -742,16 +743,15 @@ export async function promptStartDesignTimeOption(context: IActionContext) {
...
-
         if (autoStartDesignTime) {
+          // The scheduled startDesignTimeApi() runs validateAndRegenerateProjectArtifacts() once and
+          // logs a single per-project artifact summary, so don't regenerate up-front here — doing so
+          // would repeat both the work and the log lines for every project.
           scheduleStartDesignTimeApi(projectPath);
+        } else {
+          // Auto-start is off: keep source-controlled clones valid by regenerating the git-ignored
+          // host.json / local.settings.json now. Emits one concise per-project summary line.
+          await ensureProjectRootArtifacts(context, projectPath);
```
andrew-eldridge, line 754: would it make sense to move the ensureProjectRootArtifacts call out of startDesignTimeApi entirely, and just into main extension startup before reaching startDesignTimeApi?

apps/vs-code-designer/src/app/commands/binaries/validateAndInstallBinaries.ts
```diff
@@ -45,6 +46,12 @@ export async function validateAndInstallBinaries(context: IActionContext) {
       context.telemetry.properties.dependencyTimeout = `${dependencyTimeout} milliseconds`;
       context.telemetry.properties.dependencyPath = await ensureRuntimeDependenciesPath();
 
+      // Decide once, up front, whether this pass should perform the network "is there a newer
+      // version?" checks. Individual validators read the same throttle flag, so they stay in sync
+      // for the duration of this run (the timestamp is only advanced after a successful pass).
+      const performedUpdateCheck = shouldCheckForDependencyUpdates();
```
andrew-eldridge, line 52: will this be the same ~30s check from before, just only performed once per day now? If so we may need to look into other optimizations after this pr, even once a day >30s startup is way more than the target. Can be left to another pr

apps/vs-code-designer/src/app/commands/dotnet/validateDotNetIsLatest.ts
```diff
@@ -27,11 +28,16 @@ export async function validateDotNetIsLatest(majorVersion?: string): Promise<voi
           await installDotNet(context, version);
         } else {
           context.telemetry.properties.localVersion = localVersion;
-          const newestVersion: string | undefined = await getLatestDotNetVersion(context, version);
+          // Throttle the network "is there a newer .NET SDK?" lookup: an already-installed,
```
andrew-eldridge: nit: some of the comments can be cleaned up if self-explanatory. this same comment is in at least 3 different places

apps/vs-code-designer/src/app/utils/bundleFeed.ts
```diff
@@ -829,6 +829,34 @@ export type BundleInstallResult = 'unknown' | 'ok' | 'failed';
...
+ * `resetCachedBundleVersion`.
+ */
+let healthyBundleVersion: string | null = null;
+
+/**
+ * Whether the "already verified this session" skip message has been logged. Gated so the message
+ * appears at most once per session instead of once per design-time / runtime host launch (which is
+ * once per project in a multi-project workspace). Reset whenever the health cache is invalidated.
+ */
+let loggedBundleAlreadyVerified = false;
```
andrew-eldridge: this looks like it's used to avoid logging the same details (bundle verified) across multiple startDesignTimeApi, but aren't there multiple logs for each design-time startup? It may be more confusing to omit just this log in that case

andrew-eldridge (approved):

apps/vs-code-designer/src/app/utils/startRuntimeApi.ts
```diff
@@ -214,8 +214,9 @@ async function resolveChildFuncPid(processId: number, retries = 5, delayMs = 500
 
 export function stopRuntimeApi(projectPath: string): void {
   ext.outputChannel.appendLog(`Stopping Runtime API for project: ${projectPath}`);
-  const { process, childFuncPid } = ext.runtimeInstances.get(projectPath);
+  const { process, childFuncPid, port } = ext.runtimeInstances.get(projectPath);
   ext.runtimeInstances.delete(projectPath);
+  releaseReservedPort(port);
```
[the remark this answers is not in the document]
lambrianmsft: Good catch — added the same guard stopDesignTimeApi uses: fetch the instance, early-return when undefined, then destructure/delete/release. Covered by a new startRuntimeApi.test.ts. Fixed in a6579b1ba.

apps/vs-code-designer/src/app/utils/dependencyUpdateCheck.ts
lambrianmsft: Moved both to constants.ts and updated imports/tests. Fixed in a6579b1ba.

apps/vs-code-designer/src/app/utils/codeless/validateProjectArtifacts.ts
lambrianmsft: Done — each regenerate helper now returns the label(s) it changed (from the hostFileName/localSettingsFileName constants) and the call sites aggregate them, so collectChangedArtifacts is gone. 25a40a7ad.

apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
lambrianmsft: The split is deliberate: on the auto-start path the scheduled design-time startup already regenerates root artifacts and logs one summary, so an up-front pass would repeat the work and the log line for every project (the multi-project noise this PR targets). Happy to decouple it into a follow-up if you'd prefer the cleaner separation.

apps/vs-code-designer/src/app/commands/dotnet/validateDotNetIsLatest.ts
lambrianmsft: Trimmed to a one-liner pointing at shouldCheckForDependencyUpdates in all three validators. a6579b1ba.

apps/vs-code-designer/src/app/utils/bundleFeed.ts
lambrianmsft: Dropped the once-per-session gate so the skip line logs per launch like the other startup lines. edfafb86f

apps/vs-code-designer/src/app/commands/binaries/validateAndInstallBinaries.ts
lambrianmsft: Will do this in a future pr

PR #9436, opened 2026-07-21
perf(vs-code-designer): fix extension-host startup — bundle integrity fast-path + telemetry masking freeze

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix
- [x] perf - Performance improvement

## Risk Level

## What & Why

> **This PR now contains BOTH startup fixes.** #9457 has been folded in here and closed, so this
> single PR carries two independent bugs found in the same investigation:
> 1. **Bundle integrity fast-path** (+ a 24h deep-verify throttle-persist fix) — removes a
>    ~2.4s-warm / ~16.6s-cold byte hash and several CDN round-trips from the activation path.
> 2. **Telemetry masking freeze** (was #9457) — quadratic `maskUserInfo` regex backtracking in
>    `vscode-azext-utils` burning **15.4s of synchronous CPU** on the extension-host thread.
>
> They are genuinely different failure modes and needed both fixes: (1) is elapsed time /
> off-critical-path work, while (2) is the one that blocks the event loop and trips VS Code's
> **Unresponsive** warning. A build carrying only one of them still exhibits the other.
>
> **Scope note (revised after direct measurement).** An earlier revision of this description
> attributed the ~40s cold start *and* the **Unresponsive** warning to the bundle byte hash. Both
> attributions were wrong. The Unresponsive warning is fix (2), not fix (1) — nothing in the
> bundle work addresses it, and the flag persisted after every bundle change until the masking fix
> landed. The ~40s cold start was **CDN-dominated**, not byte-hash-dominated.

Cold VS Code start could take **>40s** before activation reported success. Diagnosis from real
output-channel logs surfaced **five** separate costs sitting on the activation critical path.
This PR moves all five off it.

**Measured cost of each (real bundle: `Microsoft.Azure.Functions.ExtensionBundle.Workflows`
1.179.3 — 1496 files, 593.3 MB, sequential/uncontended, 3 passes):**

| Activation cost | Before | After | Notes |
|---|---|---|---|
| 1. Bundle on-disk byte hash | **~2.4s warm / ~16.6s cold** (warm 2.10 / 2.26 / 2.46s; CPU-bound) | **0.12s** lstat fingerprint | ~19x warm, ~135x cold |
| 2. Bundle CDN source-MD5 `HEAD` | up to **30s** timeout | **4s** timeout, off hot path | dominant contributor |
| 3. Dependency integrity stat walk | ~10.7k **blocking** `statSync` | async batched, yields | ~1-2s of non-yielding event loop |
| 4. Bundle version-feed `GET` | up to **~24-40s** on a slow link | **0** CDN calls steady-state | dominant contributor |
| 5. Telemetry `maskUserInfo` masking | **15.4s SYNCHRONOUS CPU** (~650 subs / ~85KB) | **~6ms** (linear) | the actual Unresponsive cause |

Items **2 and 4 are where the ~40s of elapsed time actually came from** — network timeouts on the
activation path, not the local hash. Item 1 contributes ~2.4s warm / ~16.6s cold. Item **5 is the
only one that blocks the event loop**, and is therefore the only one that could produce VS Code's
Unresponsive warning. All five are fixed.

> **Warm vs cold.** The ~2.4s figure is page-cache-**warm**. Re-measured on a genuinely cold
> cache the same hash takes **16,637ms** (36 MB/s vs 253 MB/s warm — the cold cost is per-file
> open overhead across 1,496 files, not bandwidth). Cold is the case that matters for the first
> launch after a reboot, which is exactly when users notice startup. Every number in this PR is
> now labelled warm or cold.

**1. Bundle on-disk integrity — fast lstat fingerprint.**
`assertExtensionBundleOnDiskHealthy` -> `computeBundleContentHash` streamed and SHA-256-hashed
**every byte of every file** in the bundle (PowerShell runtime + .NET assemblies + workflow
bundle), serially, every launch.
- New `computeBundleTreeFingerprint(bundleDir)` walks the tree with `lstat` only (no byte reads),
  hashing `relPath\0size\0mtimeMs\0` per sorted entry into a sha256 digest. **~2.4s -> ~0.12s.**
- Persisted in the bundle sidecar (`treeFingerprint?`, parsed leniently so legacy sidecars stay valid).
- Fast-paths on a fingerprint match; falls back to the byte hash only when the fingerprint is
  missing (legacy -> verify + backfill), mismatches (-> existing contentMismatch/repair), or a
  **24h deep-verify throttle** is due.

**2. Bundle download steady-state path — hot-path fast-gate + background deep verify.**
`main.ts` fires `downloadExtensionBundle` fire-and-forget, and `ensureExtensionBundleHealthy` ->
`waitForExtensionBundleReady()` awaited it before the gate. Its steady-state `verifyLocalBundle`
re-ran the full byte hash **plus a 30s CDN HEAD** every launch.
- Inject `evaluateBundleFastGate` before `verifyLocalBundle`: on a fingerprint match the hot path
  returns immediately and schedules a **background** deep verify (byte hash + off-hot-path CDN
  source-MD5 republish check).
- A benign CDN republish is re-downloaded in the background with a progress notification (no
  alarming corruption toast); genuine byte corruption still repairs + toasts via the
  `forceVerify` re-entry.
- CDN HEAD timeout cut 30s -> **4s** (`FETCH_EXPECTED_MD5_TIMEOUT_MS`).

**3. Runtime-dependency integrity — async, non-blocking stat.**
`verifyDependencyIntegrity` ran `fs.statSync` in a **synchronous** loop over every manifest file
(NodeJs ~1961 + FuncCoreTools ~8795 = ~10.7k blocking stats) with no yield points. Even though
the validators run "concurrently", a sync loop never yields, so the ext-host event loop could not
service anything else for its duration.
- Converted to **async, bounded-concurrency (64) `fs.promises.stat` batches** with an `await`
  between chunks: the main thread yields repeatedly and per-file I/O overlaps. Semantics,
  telemetry, and first-failure determinism unchanged; callers now `await` it.
- **Correction:** an earlier revision claimed this resolved the Unresponsive flag. It did not —
  the flag persisted afterwards and was fixed by the telemetry-masking change (fix 5 below, the
  former #9457). Post-change profiling measures this path at ~33ms. It is retained because
  removing a ~10.7k-iteration non-yielding loop from activation is correct on its own merits, not
  because it fixed the freeze.

**4. Bundle version feed — never block activation on the CDN feed lookup.**
Even after the byte hash was off the hot path, `downloadExtensionBundleCore` still fetched the
bundle **version feed** (`getWorkflowBundleFeed` -> CDN `index.json`) on every launch *before* the
fast gate. Dependency validation blocks on this via `waitForExtensionBundleReady`, and the
in-memory feed cache is wiped each process restart, so on a slow/flaky CDN link this added
~24-40s to every cold start whose daily update-check window was due.
- **Hoist the healthy-local-bundle fast gate ABOVE the feed GET.** When the cheap lstat gate
  passes, activation returns immediately on the local bundle (`localLatest`; `fastPath` when the
  window is due / `fastPathThrottled` when not).
- The "is there a newer version?" lookup — and any upgrade download it triggers — now runs in a
  **throttled background task** (`scheduleBackgroundBundleFeedRefresh`, gated by the shared
  `shouldCheckForDependencyUpdates()` window, deduped to one in-flight run). A newer bundle is
  fetched in the background and applied on the **next** launch. Steady-state activation makes
  **zero** bundle CDN calls.
- Only a missing / corrupt / legacy-sidecar bundle (fast gate not ok) still falls through to the
  synchronous feed-backed verify/repair path (bootstrap + repair preserved).

**5. Telemetry masking freeze — the actual Unresponsive cause (folded in from #9457).**
`logSubscriptions` serialised the user's Azure subscriptions with a compact
`JSON.stringify(azureSubscriptions)`. `vscode-azext-utils` then masks every telemetry property
with regexes built on `\S+` / `\S*`. Compact JSON contains **zero whitespace**, so the entire
payload is one enormous `\S+` token and the masking regex backtracks quadratically: ~650
subscriptions (~85 KB) burned **15.4s of synchronous CPU** on the extension-host thread, tripping
VS Code's Unresponsive detector and stalling every other extension.
- Fix is `JSON.stringify(azureSubscriptions, null, 1)`. Indenting breaks the payload into short
  whitespace-delimited runs, making masking **linear (~6ms)** with byte-identical telemetry data.
- The pretty-print is therefore **load-bearing, not cosmetic**, and is commented as such so it
  isn't "tidied" back to compact later.
- This is a **synchronous main-thread** cost, which is why it — and not the async bundle byte
  hash — is what trips Unresponsive. Root cause is upstream in `vscode-azext-utils`; this is the
  local mitigation.

**Threat model preserved (bundle):** a deleted subfile drops an entry (fingerprint changes); a
modified file changes size and/or mtime (fingerprint changes). Only an in-place edit keeping
identical size AND mtime evades the fast path, and that is caught by the throttled deep byte-hash,
which now runs **in the background** rather than on the activation path. Legacy sidecars
auto-upgrade on first launch.

This is the follow-up requested on **PR #9430 review comment #5** — the bundle gate (not the
version update-check) is the startup bottleneck. Independent change; does not depend on #9430.
**#9457 has been folded into this PR and closed** — both fixes now ship together.

## Impact of Change
- **Users**: Cold start no longer blocks on the bundle byte hash, the 30s CDN HEAD, or the CDN
  version-feed GET. A newer bundle version is applied on the next launch instead of synchronously
  mid-activation. Byte-level corruption detection is unchanged in kind but now runs in the
  background, and is throttled to once per 24h. **The Unresponsive warning / extension-host
  freeze is also fixed here**, by the telemetry-masking change (fix 5) — not by any of the bundle
  work.
- **Developers**: New `treeFingerprint` sidecar field + `computeBundleTreeFingerprint` /
  `computeBundleFingerprints`; 24h deep-verify throttle constants; `verifyDependencyIntegrity` is
  now async (`Promise<boolean>`); `DownloadExtensionBundleOptions.forceVerify`; background seams
  `awaitBackgroundBundleDeepVerification` + `awaitBackgroundBundleFeedRefresh` for tests.
- **System**: First launch after upgrade (legacy sidecar) and once per 24h run the full byte hash,
  off the activation path; all other launches use the cheap stat fingerprint. The version feed is
  consulted at most once per update-check window, off the hot path. Dependency integrity remains a
  full per-file existence + size check, just non-blocking.
- **Known trade-off:** the 24h deep-verify throttle leaves a <=24h window in which a size- and
  mtime-preserving in-place edit goes undetected. An earlier revision of this description
  suggested dropping the throttle and deep-verifying on **every** launch, on the basis that the
  byte hash costs ~2.4s of "free" off-critical-path CPU. **That recommendation is withdrawn.**
  The ~2.4s was page-cache-warm; cold it is ~16.6s, and the hash is CPU-bound (CPU ~= wall), so
  even genuinely off the activation path it burns a saturated core for seconds on every launch.
  Off-path is not free. The 24h throttle is justified on CPU cost alone, independently of whether
  it ever blocks.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: `apps/vs-code-designer` vitest — `bundleFeed.test.ts` (116), `telemetry.test.ts`
  (5), `integrity.test.ts`
  (3), `binaries.test.ts` (94), `validateNodeJsIsLatest.test.ts` (17),
  `validateFuncCoreToolsIsLatest.test.ts` (10), `validateAndInstallBinaries.test.ts` (3) all pass.
  Biome clean; tsup 0 new TS errors in touched files. Verified on a real launch: the ~23-40s gap
  between the last dependency line and "extension bundle ... on-disk integrity check passed"
  collapses to ~1s. Byte-hash-vs-fingerprint costs benchmarked directly against the on-disk
  bundle (table above).

**In-host verification of the throttle (added after the initial review).** Instrumented builds
were run in a live extension host. Two consecutive launches both reported `deepVerify=false` and
ran **no byte hash at all**, and the on-disk sidecar timestamp advanced instead of staying frozen.
Measured on the activation path: `waitForExtensionBundleReady` 0-1ms, fast-gate walk 154-228ms,
total activation ~3.7-5.0s (now dominated by dependency validation, not the bundle).

> **A throttle bug this PR introduced, and the coverage gap that let it ship.** The sidecar
> rewrite was skipped whenever the fingerprints matched and `lastDeepVerifiedMs` merely *existed*.
> On the common healthy path fingerprints always match, so the fresh timestamp was never written
> and the sidecar value froze after its first write. Since `getLastDeepVerification` reads the
> sidecar **first**, that frozen value permanently outvoted the `globalState` mirror and the
> throttle never engaged — the full byte hash ran on every launch. Fixed by comparing the
> timestamp by value rather than by presence.
>
> Worth calling out for reviewers: **the suite had zero coverage of the sidecar-first path.** The
> pre-existing throttle test asserted only the `globalState` mirror, never the sidecar that
> actually decides, so all 115 tests passed against the bug — a green suite is not coverage. The
> added regression test pins a stale `lastDeepVerifiedMs` alongside **both** matching fingerprints
> with `globalState` deliberately fresh, so only the sidecar can make a deep verify due. It was
> validated in both directions: it **fails** on the old guard and passes on the new one.

Coverage highlights:
1. Fingerprint match => fast path, `computeBundleContentHash` NOT called (no `createReadStream`).
2. Fingerprint mismatch => byte hash invoked => `contentMismatch`.
3. Legacy sidecar (no `treeFingerprint`) => byte hash runs and fingerprint is backfilled.
4. Deep-verify throttle due => byte hash runs even on fingerprint match and timestamp recorded.
   Plus a regression test that the fresh `lastDeepVerifiedMs` is **persisted into the sidecar**
   when the fingerprints are unchanged (validated to fail on the pre-fix guard).
5. Download persists `treeFingerprint`; modern sidecar hot path does no byte hash / HEAD /
   background / sidecar rewrite.
6. Steady-state fast-gate schedules background deep verify; benign republish re-downloads without
   a corruption toast; genuine corruption repairs + toasts via `forceVerify`.
7. `fetchExpectedMd5` uses the 4s HEAD timeout.
8. `verifyDependencyIntegrity`: async batched stat verifies large multi-batch manifests, detects a
   size mismatch in a later batch, and preserves manifest-missing / invalid / file-missing /
   not-a-file / error semantics.
9. Bundle version feed is deferred to a background refresh (never the hot path) when a healthy
   local bundle exists — including when the update-check window is due; a newer version found by
   the background refresh is downloaded off the hot path and the update-check window is recorded;
   still consulted synchronously only when no local bundle exists (bootstrap) or the fast gate
   fails.
10. `logSubscriptions` telemetry masking: the subscriptions payload is emitted whitespace-delimited
    so `maskUserInfo` stays linear, the telemetry data is unchanged apart from formatting, and the
    error path still records `logSubscriptionsError`.


apps/vs-code-designer/src/app/utils/bundleFeed.ts
```diff
@@ -384,6 +397,86 @@ export async function computeBundleContentHash(bundleDir: string): Promise<strin
...
+ * fingerprint path is preserved. The residual in-place-edit gap is only exposed
+ * for at most one interval.
+ */
+function isDeepBundleVerificationDue(): boolean {
+  const last = ext.context?.globalState?.get<number>(lastBundleDeepVerificationKey);
+  if (typeof last !== 'number' || !Number.isFinite(last)) {
+    return true;
+  }
+  return Date.now() - last >= bundleDeepVerificationIntervalMs;
+}
```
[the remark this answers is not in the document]
lambrianmsft: Good catch on the original commit — this was addressed in a later commit on the branch. `isDeepBundleVerificationDue()` no longer reads `globalState` directly; it now calls `getLastDeepVerification(sidecar)`, which reads the deep-verify timestamp **sidecar-first** (`sidecar.lastDeepVerifiedMs`) with a `globalState` fallback, returning `undefined` only when neither source has a value. The doc comment was updated to match: a missing timestamp everywhere is treated as due (legacy / first run), which runs the byte hash exactly once and then backfills the sidecar timestamp so every subsequent launch takes the fast path. So `globalState` being unavailable during early activation no longer forces the expensive hash — the authoritative timestamp lives in the sidecar. No change needed.

libs/vscode-extension/src/graphify-out/GRAPH_REPORT.md
```diff
@@ -1,113 +1,128 @@
...
-- [[_COMMUNITY_Community 3|Community 3]]
-- [[_COMMUNITY_Community 4|Community 4]]
-- [[_COMMUNITY_Community 5|Community 5]]
-- [[_COMMUNITY_Community 6|Community 6]]
-- [[_COMMUNITY_Community 7|Community 7]]
-- [[_COMMUNITY_Community 8|Community 8]]
+- project.ts
+- project.ts
+- workflow.ts
+- index.ts
```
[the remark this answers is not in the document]
lambrianmsft: Agreed — these generated Graphify reports do not belong in this perf PR. They have been removed entirely: the branch was rebuilt to drop the `chore: update knowledge graphs` commit, so the PR is now just the 10 `apps/vs-code-designer` code files with no `graphify-out` changes. Resolved.

apps/vs-code-designer/src/app/utils/bundleFeed.ts
```diff
@@ -384,6 +397,86 @@ export async function computeBundleContentHash(bundleDir: string): Promise<strin
   return hash.digest('base64');
 }
 
+/**
+ * Fast, stat-only fingerprint of the extracted bundle tree — the cheap
```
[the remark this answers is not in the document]
lambrianmsft: Valid observation — thanks. This is a genuine (if narrow) edge case: an empty / partial extraction that still yields a source MD5 could be blessed with a sidecar. We evaluated the literal fix (return `undefined` from `computeBundleContentHash` / `computeBundleFingerprints` when no files are found) and it is the right long-term shape, but it breaks **43 of the 109** `bundleFeed.test.ts` tests: the suite intentionally models a bundle via an empty directory walk (`readdir -> []` -> the empty-input sha256, aliased `EMPTY_TREE_HASH`) across five describe blocks, so making empty -> undefined requires migrating all of those fixtures to mock non-empty trees. To keep this startup-perf PR focused and reviewable, we split it into a follow-up: Azure/LogicAppsUX#9452.

In practice the window is small — `downloadAndExtractDependency` must succeed and return `actualMd5` while extracting zero files — and the backfill / repair paths additionally gate on `hasRequiredBundleContent`.

lambrianmsft:
### Correction to this PR's description (self-review after direct measurement)

I re-benchmarked the bundle byte hash against the real on-disk bundle and found two claims in my
original description were wrong. I have rewritten the body; posting here so reviewers who already
read it see the delta.

**1. The byte hash is ~2.4s, not ~40s.**

`Microsoft.Azure.Functions.ExtensionBundle.Workflows` 1.179.3 — 1496 files, 593.3 MB, sequential
and uncontended, 3 passes:

| | wall | cpu |
|---|---|---|
| Full byte SHA-256 (`computeBundleContentHash` equivalent) | 2461 / 2258 / 2097 ms | 2406 / 2251 / 2108 ms |
| lstat tree fingerprint | 124 / 124 / 120 ms | — |

~19x, saving ~2.3s per launch (page-cache-warm; a genuinely cold read is higher). My first pass at
this benchmark reported 28-45s — that script hashed six installed bundle versions **concurrently**
via un-awaited async IIFEs, so it was measuring disk contention, not the real cost. Corrected.

The ~40s cold start was therefore **CDN-dominated** — the 30s source-MD5 `HEAD` (item 2) and the
version-feed `GET` (item 4), both on the activation path. Those are still real and still fixed
here. The local hash was a minor contributor.

**2. This PR does not fix the "Unresponsive" warning.**

That is a separate bug: quadratic regex backtracking in `maskUserInfo`
(`@microsoft/vscode-azext-utils`) burning **15.4s of synchronous CPU**, 51% of a 30s extension-host
profile. Fixed in #9457. The flag persisted through every change in this PR and cleared only when
#9457 landed. In hindsight this was predictable — the byte hash is async stream I/O and cannot hold
the event loop, so it could never have produced an Unresponsive flag.

Item 3 (`verifyDependencyIntegrity` sync -> async stat) is kept on its own merits — removing a
~10.7k-iteration non-yielding loop from activation is correct regardless — but its original
"fixes the Unresponsive flag" rationale is withdrawn. Post-change profiling puts that path at ~33ms.

**3. Open question for reviewers: should the 24h deep-verify throttle stay?**

It was sized on the assumption that the byte hash cost ~40s. At ~2.4s, and now that the deep verify
runs **off** the activation path, running it on every launch would restore pre-PR byte-level
integrity coverage in full — closing the <=24h window in which a size- and mtime-preserving in-place
edit goes undetected — at zero activation cost. I lean toward dropping the throttle. Happy to make
that change if reviewers agree.

Companion PR: #9457.

lambrianmsft:
Follow-up commit `49aabe16c` — a throttle bug **this PR introduced**, found while investigating a
report of a slow start, plus the coverage gap that let it ship green.

### The bug

`refreshBundleSidecarFingerprints` skipped its sidecar rewrite whenever the fingerprints matched
and `lastDeepVerifiedMs` merely *existed*:

```ts
const unchanged =
  fingerprints.treeFingerprint === sidecar.treeFingerprint &&
  fingerprints.structuralFingerprint === sidecar.structuralFingerprint &&
  sidecar.lastDeepVerifiedMs !== undefined;   // <- presence, not value
```

On the common "healthy bundle, nothing changed" path the fingerprints *always* match, so this
returned before writing the fresh timestamp and the sidecar value froze after its very first
write. Because `getLastDeepVerification` reads the sidecar **first** and only falls back to the
`globalState` mirror, the frozen value permanently outvoted the mirror,
`isDeepBundleVerificationDue` stayed true forever, and the full byte hash ran on **every launch**
instead of once per 24h. Fixed by comparing the timestamp by value.

### The coverage gap (the part worth reviewing)

The suite had **zero coverage of the sidecar-first path**. The pre-existing throttle test asserted
only the `globalState` mirror — never the sidecar that actually decides — so all 115 tests passed
against the bug. A green suite is not coverage.

The added regression test pins a stale `lastDeepVerifiedMs` alongside **both** matching
fingerprints with `globalState` deliberately fresh, so only the sidecar can make a deep verify
due. Validated in both directions: it **fails** on the old guard, passes on the new one. Omitting
either fixture detail makes it pass against the buggy code.

### In-host verification

Instrumented builds run in a live extension host, two consecutive launches: both reported
`deepVerify=false` and ran **no byte hash at all**, and the on-disk sidecar advanced instead of
staying frozen. Activation-path measurements: `waitForExtensionBundleReady` 0-1ms, fast-gate walk
154-228ms, total activation ~3.7-5.0s — now dominated by dependency validation rather than the
bundle. All timing instrumentation was stripped before commit; the diff is the one-line guard plus
the test.

### Two corrections to the description

1. **Warm vs cold.** The ~2.4s byte-hash figure was page-cache-**warm**. Cold it is **16,637ms**
   (36 MB/s vs 253 MB/s — per-file open overhead across 1,496 files, not bandwidth). Cold is the
   first launch after a reboot, i.e. exactly when users notice. Numbers in the body are now
   labelled.
2. **The "drop the 24h throttle" suggestion is withdrawn** — it was mine, and the cold number
   kills it. The hash is CPU-bound (CPU ~= wall), so even genuinely off the activation path it
   burns a saturated core for ~2.4s warm / ~16.6s cold on every launch. Off-path is not free. The
   throttle is justified on CPU cost alone, independently of whether it blocks.

apps/vs-code-designer/src/app/utils/integrity.ts
```diff
@@ -184,13 +184,15 @@ function delay(ms: number): Promise<void> {
...
+ * Timeout (ms) for the HEAD request issued by `fetchExpectedMd5`. The CDN
+ * republish (Content-MD5) check now runs only on the throttled background
+ * deep-verify path — never on the activation hot path — but a stalled
+ * connection (flaky network, captive portal, mis-configured proxy) must still
+ * never hang for long. A short budget keeps the background check snappy; the
+ * caller falls back to the cached bundle when the CDN cannot be reached
+ * promptly.
  */
-const FETCH_EXPECTED_MD5_TIMEOUT_MS = 30_000;
+const FETCH_EXPECTED_MD5_TIMEOUT_MS = 4_000;
```
bjbennet, line 195: This is a significant decrease in timeout time- is there any concern with it being too low? Or do we have metrics that show this time is sufficient?

lambrianmsft: No there is no concern. This was added recently to detect bundle corruption. If the cdn fails to respond, we will use the cached bundle, and if it does respond, we will possibly grab and download a new bundle. If it fails to respond we have a really long unnecessary wait that will delay user startup.

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -650,51 +650,67 @@ export function verifyDependencyIntegrity(context: IActionContext, dependencyNam
...
-          )
-        );
-        return false;
+    // Deep-verify each recorded file with async, bounded-concurrency stat batches. A dependency
+    // manifest can list many thousands of files (Func Core Tools alone is ~8.8k), and a synchronous
+    // fs.statSync loop would block the extension-host event loop long enough for VS Code to flag the
+    // extension as "Unresponsive". Batching fs.promises.stat and awaiting between chunks keeps the
+    // main thread responsive while overlapping the per-file I/O (faster than serial statSync).
+    const integrityBatchSize = 64;
+    for (let batchStart = 0; batchStart < manifest.files.length; batchStart += integrityBatchSize) {
```
andrew-eldridge, line 659: depending on the speed of this check now with batching, may want to put deep validation behind a setting flag

apps/vs-code-designer/src/constants.ts
```diff
@@ -304,6 +304,14 @@ export const extensionBundleId = 'Microsoft.Azure.Functions.ExtensionBundle.Work
 export const targetBundleKey = 'FUNCTIONS_EXTENSIONBUNDLE_SOURCE_URI';
 export const bundleSourceMd5SidecarFile = '.bundle-source-md5';
 
+// globalState key + interval throttling the expensive full-byte extension-bundle
+// integrity hash. On most launches a fast lstat tree fingerprint is compared
+// instead; the deep byte-hash only runs when this throttle says a verification is
+// due (defense-in-depth against an in-place edit that keeps identical size+mtime).
+// Mirrors the dependency update-check throttle pattern.
+export const lastBundleDeepVerificationKey = 'azureLogicAppsStandard.lastBundleDeepVerification';
```
andrew-eldridge, line 312: there's not a consistent naming convention for these global state keys (e.g. some don't use 'azureLogicAppsStandard' prefix or have another prefix- we should address in another pr

andrew-eldridge (approved):

apps/vs-code-designer/src/constants.ts
lambrianmsft: agreed

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: This check is sub 4 seconds now. We can have a follow up PR if we get more customer reports.

PR #9443, opened 2026-07-21
fix(designer-v2): preserve MCP allowed-tools selection after connection create

ccastrotrejo:
## Commit Type
- [x] fix - Bug fix

## Risk Level

## What & Why
Creating a connection in the designer-v2 MCP tool wizard cleared every "allowed tools" checkbox. This was reported as a regression in both the Azure Portal and standalone.

Root cause: the wizard held the tool selection in component-local `useState` behind a one-shot `hasInitializedTools` guard, while the tools list is a React Query keyed on the connection identity (with `staleTime: 0`). The identity resolves a beat after a connection is created, which changes the query key and refetches. During that window `toolsData` went `undefined` so the grid rendered with nothing checked, and the one-shot guard could not recover the selection (and for managed servers whose tool `value`s change after identity resolves, the stale selection no longer matched any option).

This is a v2-specific regression: the v1 wizard keeps tool selection in Redux (`state.mcpSelection.selectedOperations`), which is durable across the refetch. The v2 rewrite did not carry that guarantee over.

The fix keeps the durable-selection behavior without moving state back into Redux:
- Add `keepPreviousData: true` to the tools query so the grid does not flash empty while the identity-keyed refetch is in flight.
- Replace the one-shot init guard with a signature-based reconcile that keeps still-valid selections when the tool set changes, defaults to selecting all only when nothing valid remains, skips stale (previous-key) data, and never re-selects after the user intentionally clears tools within the same set.
- Reset the reconcile signature when the user explicitly switches connections so the newly selected connection re-initializes to "all".

## Impact of Change
- **Users**: Allowed-tool checkboxes are retained after a connection is created/selected; the grid no longer flashes empty during the identity refetch. Users can still deselect tools (including clearing all) without the selection snapping back.
- **Developers**: MCP tool selection now reconciles against the loaded tool set instead of relying on a one-shot flag; behavior is covered by regression tests that exercise the real `useQuery` flow.
- **System**: No new dependencies. Confined to `mcpToolWizard.tsx`. Uses the existing React Query v4 `keepPreviousData` / `isPreviousData` API.

## Test Plan
- [ ] E2E tests added/updated
- [x] Tested in: `pnpm vitest` for designer-v2 browse suite (78/78 pass, incl. 4 new regression tests); `tsc --noEmit` clean; Biome clean

New regression tests in `mcpToolWizard.spec.tsx` (real query flow, only `getListDynamicValues` mocked): auto-select-all on first load; selection preserved when the connection identity resolves and refetches; only still-valid tools kept when the tool set changes; no re-select after the user clears the selection.

Note on E2E: the current Playwright MCP suite (`e2e/designer/mcpSerialization.spec.ts`) is serialization-only and has no mock support for `listMcpTools` / the connection-create wizard UI, so a meaningful E2E would require new mock infrastructure. The regression is deterministically covered at the component/integration level instead. Happy to build the E2E mock harness as a follow-up if desired.

## Contributors
Reported by Stephen O'Connor and Abode Saafan.

## Screenshots/Videos
N/A

<sub>Note: `Localize/lang/strings.json` picked up one unrelated line (`g0M0Ey`) from the repo's mandatory `extract` pre-commit hook syncing pre-existing drift; it is not from this fix's code.</sub>


libs/designer-v2/src/lib/ui/panel/recommendation/browse/mcpToolWizard.tsx
```diff
@@ -330,18 +335,40 @@ export const McpToolWizard = () => {
...
-      const allToolValues = toolsData.map((tool) => tool.value);
-      setLocalAllowedTools(allToolValues);
-      dispatch(setMcpWizardTools(allToolValues));
-      setHasInitializedTools(true);
+    if (isPreviousToolsData || !toolsData || toolsData.length === 0) {
+      return;
+    }
+
+    const availableValues = toolsData.map((tool) => tool.value);
+    const signature = availableValues.join('|');
```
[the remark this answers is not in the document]
ccastrotrejo: Good catch, accepted. Tool values are server-provided, so a value containing the `|` delimiter could collide two distinct tool sets into the same signature and skip reconciliation. Switched to `JSON.stringify(availableValues)` for a collision-safe signature in 9916254.

Localize/lang/strings.json
```diff
@@ -4193,6 +4193,7 @@
   "_fuBVBE.comment": "Logic app name field label",
   "_fvGvnA.comment": "Chatbot error message",
   "_g076bL.comment": "Placeholder title for a newly inserted Email parameter",
+  "_g0M0Ey.comment": "Tooltip text shown when the workflow runtime is not available",
```
AbodeSaafan, line 4196: are these changes intended?

AbodeSaafan (approved):
thanks for the quick fix! Will try to get a designer package from this soon to see if customer can use that to verify but we might also need a portal canary deployment with it

Localize/lang/strings.json
ccastrotrejo: Good catch, not intended. These lines were auto-added by the mandatory `extract` pre-commit hook, which regenerated `strings.json` and picked up pre-existing drift (`g0M0Ey` already exists in `apps/vs-code-react/src/intl/messages.ts` on main but was never synced into the compiled file). It is unrelated to this fix, so I reverted it in b3f9869 and the PR now touches only `mcpToolWizard.tsx` and its test. The underlying main drift can be handled separately.

ccastrotrejo: Update: I put this string back in 0b59ee7. On reflection, dropping it left `Localize/lang/strings.json` out of sync with source (`g0M0Ey` already exists in `apps/vs-code-react/src/intl/messages.ts` on main), and the mandatory `extract` pre-commit hook re-adds it on every touched commit anyway, so keeping the extract output in sync is the correct state.

To be clear on cause and effect: this string is unrelated to the flaky `logic-apps-shared` `intl.spec` CI failure. That test loads the compiled French messages (`compiled-lang/strings.fr.json`), a different file that was never touched here; it passes locally and the CI red was an async-import flake. So this change is purely loc-sync hygiene, not a pipeline fix.

PR #9164, opened 2026-05-13
perf(vscode): parallelize E2E matrix + harden runtime readiness + restore coverage (closes #9172)

lambrianmsft:
## Commit Type

- [ ] feature - New functionality
- [ ] fix - Bug fix
- [ ] refactor - Code restructuring without behavior change
- [x] perf - Performance improvement
- [ ] docs - Documentation update
- [ ] test - Test-related changes
- [ ] chore - Maintenance/tooling

## Risk Level

- [ ] Low - Minor changes, limited scope
- [ ] Medium - Moderate changes, some user impact
- [x] High - Major changes, significant user/system impact

## What & Why

This PR is the collapsed integration branch for the VS Code E2E CI optimization stack. It folds #9175, #9178, #9179, #9180, and #9181 into #9164 in order. #9175 was already represented on #9164 before the collapse; #9179 was merged separately because its decision/audit docs were not included in #9181.

The integrated change makes VS Code E2E feedback faster, stricter, and easier to triage:

- Splits VS Code E2E into strict scenario-level matrix jobs with a `vscode-e2e-summary` rollup.
- Removes workflow-level `continue-on-error` and strict-scenario internal masking.
- Keeps `workflow_dispatch` for manual reruns when PR triggers are coalesced.
- Adds shared extension-build/setup flow, workspace fixture split, and scenario selectors in `run-e2e.js`.
- Restores and expands coverage for keyboard navigation, rules engine/custom code shapes, stateless variables, run-after persistence, conversion paths, and multi-designer scenarios.
- Hardens runtime/design-time readiness, marketplace dependency installation, webview interactions, and action-level run verification.
- Includes D-001 and the E2E shape-coverage audit so the final PR documents the intended test-shape contract.

Latest collapsed head: `4b68281a122cae99c323430c588277f0e5cbefdd`.

## Impact of Change

- **Users**: No product behavior changes expected. This is CI, test infrastructure, E2E coverage, and documentation.
- **Developers**: Faster and more granular VS Code E2E feedback; failures now point at individual scenarios instead of a broad serial job.
- **System**: Higher GitHub Actions matrix parallelism and stricter VS Code E2E failure semantics. Because this affects CI gating for every VS Code-related PR, risk is high.

## Test Plan

- [x] Unit tests added/updated
- [x] E2E tests added/updated
- [x] Manual testing completed
- [x] Tested in: local collapsed-branch validation and GitHub Actions on #9164/#9181.

Validation evidence:

[...]

PR #9449, opened 2026-07-23
fix(vscode): exclude local-only settings from portal deploy

lambrianmsft:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

On deploy, `uploadAppSettings` pushes `local.settings.json` values to the portal. Some values are local-only and must not be deployed:

- **`languageWorkers__node__defaultExecutablePath`** — an absolute path to the developer's local Node.js binary, set during pre-debug (PR #9164) so the inline-code language worker resolves locally. In the cloud that path does not exist, so the Node worker never starts and **Execute JavaScript Code** actions hang in a **Running** state indefinitely (ADO #10109800). Added to the deploy `settingsToExclude` list.
- **Empty `WORKFLOWS_*` connector sentinels** (`WORKFLOWS_SUBSCRIPTION_ID`, `WORKFLOWS_TENANT_ID`, `WORKFLOWS_RESOURCE_GROUP_NAME`, `WORKFLOWS_LOCATION_NAME`) — empty locally when Azure connectors are disabled/cancelled. Azure sets correct values on provisioning, but uploading the empty sentinel overwrote them and broke managed-connection resolution. `uploadAppSettings` now skips these keys when empty; non-empty values still upload.

## Impact of Change

- **Users**: Deploying a project using the inline "Execute JavaScript Code" action (or with Azure connectors disabled locally) no longer breaks the workflow in the portal.
- **Developers**: `settingsToExclude` in `deploy.ts` gains the inline-code node path; `uploadAppSettings` gains value-aware skipping of empty `WORKFLOWS_*` settings.
- **System**: No architectural change; only narrows what is uploaded on deploy.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: VS Code Standard Logic App deploy; verified excluded settings do not appear in uploaded app settings

## Contributors

@lambrianmsft

## Screenshots/Videos

N/A

### DIFF
--- a/apps/vs-code-designer/src/app/commands/appSettings/uploadAppSettings.ts
+++ b/apps/vs-code-designer/src/app/commands/appSettings/uploadAppSettings.ts
@@ -2,7 +2,14 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import { localSettingsFileName, logicAppFilter } from '../../../constants';
+import {
+  localSettingsFileName,
+  logicAppFilter,
+  workflowLocationKey,
+  workflowResourceGroupNameKey,
+  workflowSubscriptionIdKey,
+  workflowTenantIdKey,
+} from '../../../constants';
 import { ext } from '../../../extensionVariables';
 import { localize } from '../../../localize';
 import { getLocalSettingsJson } from '../../utils/appSettings/localSettings';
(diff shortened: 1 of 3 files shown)


apps/vs-code-designer/src/app/commands/appSettings/uploadAppSettings.ts
```diff
@@ -52,18 +59,26 @@ export async function uploadAppSettings(
...
-          delete localSettings.Values?.[settingName];
-          excludedAppSettings.push(settingName);
-        }
-      });
-    }
+    // Azure connector settings hold an empty-string sentinel locally when Azure connectors are
+    // disabled (or the user cancels the sign-in wizard). Uploading the empty value would overwrite
+    // the correct platform-provided value in the portal and break managed-connection resolution,
+    // so skip these keys whenever their local value is empty.
+    const skipWhenEmptySettings = [workflowSubscriptionIdKey, workflowTenantIdKey, workflowResourceGroupNameKey, workflowLocationKey];
```
andrew-eldridge, line 66: as a future pr, we should avoid setting empty value in local settings when connectors setup is skipped and just track in globalState instead. Then this check can be removed

andrew-eldridge (approved):

apps/vs-code-designer/src/app/commands/appSettings/uploadAppSettings.ts
lambrianmsft: yep will do a cleanup later

apps/vs-code-designer/src/app/commands/appSettings/uploadAppSettings.ts
```diff
@@ -52,18 +59,26 @@ export async function uploadAppSettings(
...
+
+    Object.keys(localSettings.Values).forEach((settingName) => {
+      const isExcludedByName =
+        exclude?.some((exclusion) =>
+          isString(exclusion) ? settingName.toLowerCase() === exclusion.toLowerCase() : settingName.match(new RegExp(exclusion, 'i'))
+        ) ?? false;
```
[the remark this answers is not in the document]
lambrianmsft: When the exclusion is a `RegExp`, reuse it directly and only re-create it (`new RegExp(source, flags + 'i')`) to add case-insensitivity when `ignoreCase` is missing — never passing a RegExp into the flags overload.

PR #9453, opened 2026-07-23
fix(pipeline): route npm/pnpm package restores through CFS for SFI Network Isolation (SR21)

Elaina-Lee:
## Commit Type
- [ ] fix - Bug fix
- [x] chore - Maintenance/tooling

## Risk Level

## What & Why
Route npm/pnpm package restores through Microsoft's Central Feed Service (CFS) instead of the public npmjs.com registry. Required for SFI Network Isolation compliance (AzRel Red Flag MountainPass SR21, IcM [#839785749](https://portal.microsofticm.com/imp/v5/incidents/details/839785749/home)). Deadline: August 21, 2026.

## Impact of Change
- **Users**: None — build pipeline change only, no user-facing impact
- **Developers**: CI builds now pull npm packages from the internal CFS-backed feed. Local dev is unaffected unless you opt in to the .npmrc.
- **System**: Adds \.npmrc\ at repo root pointing to \one_PublicPackages\ CFS feed. Adds \NpmAuthenticate@0\ task and \
pm_config_registry\ env var to \.azure-pipelines/templates/setup.yml\.

## Test Plan
- [ ] E2E tests added/updated
- [x] Tested in: Pipeline will be validated on merge; change is additive (auth + registry config only)


takyyon:
Thanks for driving the SR21 network-isolation work, @Elaina-Lee — routing restores through CFS is the right goal and the deadline is real. I did a careful pass and wanted to share a few findings so this lands safely, especially since `Azure/LogicAppsUX` is a public repo. Happy to pair on any of it.

### 1) Committing the CFS feed as the repo-default `.npmrc` will break external contributors
A tracked root `.npmrc` is auto-loaded by **every** `pnpm install` — it isn't opt-in. With `registry=<msazure CFS feed>` + `always-auth=true` committed, anyone without `msazure` ADO access (external contributors, forks, downstream consumers) gets a **401** on `pnpm install`. The description notes "local dev unaffected unless you opt in," but because this is the root `.npmrc`, it becomes the default for everyone.

### 2) Security: keep the internal feed + auth out of the tracked tree
Committing an internal `registry=` + `always-auth=true` next to a pipeline that injects a live token is the pattern pnpm hardened against in [GHSA-3qhv-2rgh-x77r](https://github.com/pnpm/pnpm/security/advisories/GHSA-3qhv-2rgh-x77r) — a change that repoints the registry can cause the injected token to be sent elsewhere during install (pnpm now ignores registry/creds in a committed project `.npmrc`). The consumer here (`1esmain.yml`) is `trigger: none` / `pr: none`, so it isn't fork-triggered and this isn't directly exploitable today — but keeping the feed/auth CI-only is the safer, future-proof posture.

### 3) A couple of regressions worth preserving
- The autogenerated `.npmrc` drops the existing pnpm settings: `auto-install-peers=true`, `resolution-mode=highest`, `ignore-workspace-root-check=true`. These affect workspace peer resolution and root installs.
- `ignore-scripts=true` skips lifecycle scripts: husky (`prepare: "husky || true"`) won't install git hooks, and native/build scripts (esbuild, playwright) won't run. If supply-chain hardening is the intent, pnpm's `onlyBuiltDependencies` allowlist keeps the hardening without breaking those.

### 4) Registry override is on the pnpm *bootstrap* step, not the restore
`npm_config_registry` is set only on `npm install -g pnpm`, not on the actual `pnpm install --frozen-lockfile` step — so the restore relies solely on the committed `.npmrc`. Moving the override onto the restore step removes the need to commit the registry at all.

### 5) Cross-org auth — worth verifying this authenticates
`setup.yml` is consumed only by `1esmain.yml`, which runs on the internal 1ES pool `VSEngSS-MicroBuild2022-1ES` (MicroBuild/DevDiv) — a different ADO org than the `msazure` feed. `NpmAuthenticate@0` with only `workingFile` authenticates via the build's `System.AccessToken`, which is scoped to the pipeline's own org and generally can't reach a feed in another org. Cross-org typically needs either an npm **service connection** via `customEndpoint`, or the **org-local CFS feed URL** that CFS onboarding provisions for this pipeline. Worth confirming the current `msazure` URL actually authenticates from this pipeline.

### 6) Scope check
`setup.yml` → `build.yml` runs `pnpm run build:extension`, so this affects only the manual **VS Code extension release build**. The designer package, standalone, docs, and PR validation all run on GitHub Actions (`ubuntu-latest`, public npm) and aren't covered here. If SR21 expects those isolated too, that's a separate (and trickier for OSS) conversation — might be worth confirming with the SR21 owner whether internal-1ES-only satisfies the requirement.

### Suggested shape (CI-only)
Keep the tracked `.npmrc` public (restore the three settings), and inject CFS only in the pipeline:

```yaml
# setup.yml — write a CI-only npmrc OUTSIDE the repo, auth it, point pnpm at it
- powershell: |
    $ciNpmrc = Join-Path "$(Agent.TempDirectory)" "cfs.npmrc"
    Set-Content -Path $ciNpmrc -Encoding ascii -Value @(
      "registry=$(cfsNpmRegistry)",   # from a variable group / service connection, not hardcoded
      "always-auth=true"
    )
    Write-Host "##vso[task.setvariable variable=CI_NPMRC]$ciNpmrc"
  displayName: "Write CI-only CFS .npmrc"

- task: NpmAuthenticate@0
  inputs:
    workingFile: $(CI_NPMRC)
    # customEndpoint: <npm service connection>   # if the feed is in a different ADO org

- script: npm install -g pnpm
  env: { npm_config_userconfig: $(CI_NPMRC) }

- script: pnpm install --frozen-lockfile --strict-peer-dependencies --recursive
  workingDirectory: $(working_directory)
  env: { npm_config_userconfig: $(CI_NPMRC) }
```

This keeps `git clone && pnpm install` working for external contributors while CI restores flow through CFS. Worth a quick check that `--frozen-lockfile` resolves cleanly against the feed (lockfile tarball URLs/integrity).

Happy to help wire up the variable group / service connection or pair on the `setup.yml` change. Thanks again for tackling this!

takyyon:
The updated approach looks much better, @Elaina-Lee — root `.npmrc` stays public and the CFS config is scoped to the pipeline. Thanks for the quick turnaround! 🙏

Since the goal is specifically "route restores through CFS," could we scope to the truly minimal set? For each addition below that doesn't seem required for registry routing, it'd help to understand **why it's needed and whether it actually helps** — otherwise I'd suggest dropping it to keep the diff minimal and easy to reason about for SR21:

1. **`ignore-scripts=true`** (in `.azure-pipelines/.npmrc`) — this is orthogonal to registry routing (a separate supply-chain hardening the Gardener plugin bundles) and disables lifecycle scripts for the entire install. What concrete threat does it mitigate on this pipeline, given it's a manual, protected build (`trigger: none` / `pr: none`)? And does that benefit outweigh the risk of breaking `pnpm run build:extension` (some deps rely on postinstall/native install scripts)? If it's wanted, could it be its own PR with a validated build rather than bundled here?

2. **`npm_config_registry` env on the `npm install -g pnpm` step** — by the time this step runs, the append + `NpmAuthenticate` have already written the CFS registry (and token) into root `.npmrc`, which this step reads at repo-root cwd. What does the env add that the file doesn't already provide? If nothing, can it be removed? (It's also the only remaining hardcoded feed URL in the YAML.)

3. **`type .npmrc`** (3rd line of the append step) — is this doing anything beyond debug logging? If it's just for log visibility, does it need to stay in the committed template?

4. **Root `.npmrc` trailing-newline change** — the `echo.>>` already inserts a separating newline before the appended block, so this doesn't look necessary. Can root `.npmrc` be left untouched for a clean, minimal diff?

A minimal version, if useful: root `.npmrc` untouched; `.azure-pipelines/.npmrc` = `registry` + `always-auth` only; `setup.yml` = append + `NpmAuthenticate`. Same routing, smaller surface.

Separately (not a minimality point, but worth a runtime check): `NpmAuthenticate@0` has no `customEndpoint`, and this pipeline runs on the DevDiv 1ES pool while the feed is in the `msazure` org — worth confirming `System.AccessToken` authenticates cross-org, or whether a service connection is needed.

takyyon (approved):

PR #9450, opened 2026-07-23
fix(ci): auth agentic workflows via built-in GITHUB_TOKEN + harden pr-ai-validation gate

takyyon:
## Commit Type

- [ ] feature - New functionality
- [x] fix - Bug fix
- [ ] refactor - Code restructuring without behavior change
- [ ] perf - Performance improvement
- [ ] docs - Documentation update
- [ ] test - Test-related changes
- [ ] chore - Maintenance/tooling

## Risk Level

- [ ] Low - Minor changes, limited scope
- [x] Medium - Moderate changes, some user impact
- [ ] High - Major changes, significant user/system impact

## What & Why

Two CI fixes:

**1. Agentic (gh-aw) workflows were failing to authenticate.** `issue-triage-agent`, `issue-triage-feedback`, and `daily-repo-status` failed on every run at **Execute GitHub Copilot CLI** because they authenticated Copilot through an expired `COPILOT_GITHUB_TOKEN` PAT. Adding `copilot-requests: write` to each workflow's permissions and recompiling makes the gh-aw engine authenticate with the built-in `${{ github.token }}` (`S2STOKENS=true`), removing the PAT entirely — the same approach `pr-ai-validation.yml` already uses.

**2. The `pr-ai-validation` gate could not validate large-diff PRs (this PR included).** The gate passed the prompt (system prompt + PR metadata + `head -2000` of the diff) as a single `-p "$(...)"` command-line argument. On Linux a single argv entry is capped at `MAX_ARG_STRLEN` (128 KB); large generated diffs (like this PR's recompiled `.lock.yml` files — 142 KB prompt) push past that, so `execve` fails with **E2BIG**, surfacing as an instant **exit 126** with empty output — every run, deterministically. `copilot --version` still worked (tiny args), which masked the cause and made it look intermittent. The gate then fell back to the Logic App and, on a gateway error, either skipped to a vacuous green (502) or threw a red error (504) — so it never actually validated large PRs.

## Impact of Change
- **Users**: None — no product-code or UI changes.
- **Developers**: Agentic triage/status workflows authenticate via the built-in `GITHUB_TOKEN` (no PAT to rotate). The PR-description gate now validates PRs of any diff size instead of silently failing on large ones.
[...]

PR #9456, opened 2026-07-24
fix(ci): validate large-diff PRs via stdin (avoids E2BIG in pr-ai-validation)

takyyon:
## Commit Type
- [x] fix - Bug fix

## Risk Level

## What & Why

The `pr-ai-validation` gate silently fails to validate large-diff PRs.

It builds the validation prompt (system prompt + PR metadata + `head -2000` of the diff) and passes it to the Copilot CLI as a single `-p "$(...)"` command-line argument. On Linux a single argv entry is capped at `MAX_ARG_STRLEN` (128 KB) regardless of the much larger total `ARG_MAX`. PRs with large generated diffs push the prompt past 128 KB, so `execve` fails with **E2BIG**, which surfaces as an immediate **exit 126 with empty output** on every run — deterministically, not intermittently. `copilot --version` still works (tiny args), which masks the cause and makes it look flaky. The gate then falls back to the Logic App and, on a gateway error, either skips to a vacuous green (502) or throws a red error (504) — so large PRs are never actually validated.

This passes the prompt via **stdin** (`copilot --yolo --model X < /tmp/full-prompt.txt`), which has no per-argument size limit, so the CLI validates PRs of any diff size. It also hardens the surrounding step: a uniform non-blocking skip on any infra-unavailable outcome (removing the 502/504 asymmetry), a pinned + cached CLI install with verify-and-retry, and a `workflow_dispatch` `pr_number` input for manual validation.

## Impact of Change
- **Users**: None — CI-only change, no product code.
- **Developers**: The PR-description gate now validates PRs of any size instead of failing on large diffs; infra outages produce a non-blocking skip rather than a spurious red.
- **System**: `pr-ai-validation.yml` delivers the prompt via stdin, uses a uniform fallback, and pins/caches the Copilot CLI install.

## Test Plan
- [ ] E2E tests added/updated
- [x] Tested in: verified locally that a 150 KB prompt via stdin runs cleanly (exit 0) while the same content via `-p` argv triggers E2BIG on Linux; verified via `workflow_dispatch` that the CLI validates a full large diff end-to-end.

CI/workflow config only; no product code, so unit/E2E tests are not applicable.

## Screenshots/Videos
Not applicable — no UI changes.

### DIFF
--- a/.github/workflows/pr-ai-validation.yml
+++ b/.github/workflows/pr-ai-validation.yml
@@ -4,6 +4,11 @@ on:
   pull_request_target:
     types: [opened, edited, synchronize, labeled, unlabeled]
   workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to validate (manual runs only)'
+        required: false
+        type: string
 
 permissions:
   pull-requests: write
@@ -15,6 +20,9 @@ jobs:
   validate-pr:
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    env:
+      PR_NUMBER: ${{ github.event.pull_request.number || github.event.inputs.pr_number }}
+      COPILOT_CLI_VERSION: 1.0.72-0
     steps:
       - name: Check if contributor is trusted
         uses: actions/github-script@v8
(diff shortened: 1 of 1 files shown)


takyyon:
Consolidating into #9450 instead of a separate PR.

PR #7560, opened 2025-06-17
feat(Designer): RoleService + assignment for agent MSI connections

rllyy97:
## Commit Type
- [X] feature - New functionality

## Risk Level
- [ ] Low - Minor changes, limited scope
- [X] Medium - Moderate changes, some user impact
- [ ] High - Major changes, significant user/system impact

## What & Why
- Added new `RoleService` and base implementation
- 

## Impact of Change
- **Users**: Agent MSI connections will now require either certain roles on the app, or role write permissions
- **Developers**: New service to be used by other parts of the application
- **System**: No impact

## Test Plan
- [ ] Unit tests added/updated
- [X] E2E tests added/updated
- [X] Manual testing completed
- [X] Tested in: Local environment

## Contributors
@rllyy97

PR #9460, opened 2026-07-27
fix(designer): Updated agent loop role assignments

rllyy97:
## Commit Type

- [x] fix - Bug fix

## Risk Level

- [x] High - Major changes, significant user/system impact

> Touches RBAC granted to customer managed identities. It only ever *reduces* what we grant and never removes existing assignments, but see **Pre-merge validation gap** below.

## What & Why

Fixes two problems with the role assignments the designer auto-creates for agent connections that authenticate with managed identity. Reported via [ICM 820634968](https://portal.microsofticm.com/imp/v5/incidents/details/820634968/summary) (support case 2606100050000251), where the customer said:

> "PG clarified that, if the signed-in user has roleAssignments/write on the OpenAI resource then these 3 role assignments are auto created for the Logic App managed identity... **This is not a solution. We can't grant this permission to our users or the MI.**"

### 1. We were over-granting by 3 roles

Agent MSI connections auto-assigned four built-in roles to the workflow app identity. The BPM runtime only ever issues **data-plane** requests against the model endpoint — chat completions with the `https://cognitiveservices.azure.com/` audience, and the Foundry agents/responses APIs with the `https://ai.azure.com/` audience. It never calls the ARM control plane for the account, and specifically **never calls `listKeys`**.

| Role | `dataActions` | Needed? |
|---|---|---|
| Azure AI Administrator | none | No — control-plane only, grants nothing at inference time |
| Cognitive Services Contributor | none | No — control-plane only, and hands over key-listing rights |
| Azure AI Developer | `accounts/OpenAI/*` | No — strict subset of Foundry User |
| Azure AI User | `Microsoft.CognitiveServices/*` | **Yes — the only one that matters** |

Microsoft's [Foundry RBAC guidance](https://learn.microsoft.com/en-us/azure/ai-foundry/concepts/rbac-azure-ai-foundry) is explicit: *"Don't assign built-in roles that start with **Cognitive Services**... Similarly, don't use the **Azure AI Developer** role for Foundry work."*

### 2. Role lookup was matching on display name, which Azure renamed

`Azure AI User` was renamed to **Foundry User** (role ID unchanged: `53ca6127-db72-4b80-b1b0-d745d6d5456d`). Our lookup matched `properties.roleName` against a hard-coded string, and #9279 added a guard that **silently skips** names that don't resolve. Net effect once the rename reaches a tenant:

- the one role that actually matters is never assigned, and
- `useHasRoleDefinitionsByNameQuery` returns a **false "you already have the roles" pass**, so the connection UI gate opens without the permission being present.

There was even a test encoding the symptom without diagnosing it: *"skips requested names with no matching definition... e.g. 'Azure AI User' is not returned as a built-in role in some environments."*

The docs call this out directly: *"use the role definition ID (GUID) instead of the role name in your code to avoid issues during the rename rollout."* This PR switches to ID matching.

### Also consolidates 4 duplicated lists

The role list was hard-coded in four places (shared manifest, both `parametersTab` copies, both Standalone save paths). Now a single `AGENT_MSI_REQUIRED_ROLE_DEFINITION_IDS` constant in `logic-apps-shared`.

## Impact of Change

- **Users**: New agent MSI connections get **1 role instead of 4** on the Cognitive Services / Foundry resource. This substantially lowers the bar for locked-down tenants that can't grant `Microsoft.Authorization/roleAssignments/write` broadly, and stops handing the workflow identity account-management and key-listing rights it never exercises. **Existing role assignments are untouched** — nothing is revoked. Tenants already on the renamed Foundry roles start getting the data-plane role assigned correctly again.
- **Developers**: Two breaking API changes on `@microsoft/logic-apps-shared` / `@microsoft/logic-apps-designer`:
  - `ManagedIdentitySetting.requiredRoles: string[]` → `requiredRoleDefinitionIds: string[]` (display names → role definition GUIDs)
  - `getMissingRoleDefinitions(resourceId, definitionNames)` → `getMissingRoleDefinitions(resourceId, definitionIds)` — **same TypeScript signature (`string[]`), different semantics**, so this will not fail to compile downstream. Callers passing display names will silently match nothing and assign no roles.
  - `useRoleDefinitionsByNameQuery` → `useRoleDefinitionsByIdQuery`; `useHasRoleDefinitionsByNameQuery` → `useHasRequiredRoleDefinitionsQuery`
  - ⚠️ **`AzureUX-LogicAppsPortal` must be updated in lockstep.** Its `Views/Workflow/utils/roleActions.ts` calls `getMissingRoleDefinitions` with the four display names. Because the type is unchanged, this breaks *silently*, not at build time. Follow-up PR required before the portal picks up a new package version.
- **System**: No new network calls or dependencies. Slightly fewer ARM `roleAssignments` PUTs on workflow save (1 instead of up to 4). The built-in role definition fetch is unchanged and still React Query cached.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: local build + unit tests only — see gap below

**Automated:**
- `libs/designer` — 45 passed (`role.spec.ts` 17, `openAIConnector.spec.tsx` 28)
- `libs/designer-v2` — 45 passed (same two suites)
- Rewrote the role spec fixtures so `ArmResource.name` holds the role GUID (matching real ARM shape) rather than the display name, which the old fixtures had wrong
- Added coverage for case-insensitive GUID matching and for unknown/unresolvable role IDs
- `pnpm run build:lib` — 9/9 packages green
- `npx tsc -p libs/logic-apps-shared/tsconfig.json --noEmit` — clean
- `npx biome check --write` — clean

**⚠️ Pre-merge validation gap — please confirm before merging:**

`Foundry User` covers *"build and develop in a project (data actions)"* but **not** *"publish agents"* (that needs `Foundry Project Manager`). The runtime's `POST {endpoint}/agents/{agentName}/versions` reads as a data-plane call and should be covered, but this has **not** been validated end-to-end against a live Foundry Agent Service workflow. If that path turns out to need more, we should add the specific role rather than reinstating the control-plane ones.

Recommended manual pass before merge, on a tenant where the Foundry rename has landed:
1. Create an agent connection with MSI auth against an **Azure OpenAI** account → save → confirm the workflow runs.
2. Same against a **Foundry Agent Service** project (create + invoke an agent) → confirm agent creation and run both succeed.
3. Verify only **Foundry User** is assigned on the resource afterwards.

## Contributors

Investigation context from ICM 820634968: @kshahid (support), @swak, @ahalkela, @krmitta, @hyehwalee.
Prior art this builds on: #7560, #8024, #8916, #9279.

## Screenshots/Videos

N/A — no visual changes. The only UI-visible difference is that the "missing role write permissions" warning in the agent connection panel now evaluates against the correct role, so it should stop appearing spuriously (and stop being spuriously suppressed).


Eric-B-Wu (approved):

takyyon (approved):
Can we please test this with all kinds of Agent connections?

PR #9147, opened 2026-05-07
fix(vscode): stop debug startup on Azurite readiness timeout

lambrianmsft:
## Commit Type

- [ ] feature - New functionality
- [x] fix - Bug fix
- [ ] refactor - Code restructuring without behavior change
- [ ] perf - Performance improvement
- [ ] docs - Documentation update
- [ ] test - Test-related changes
- [ ] chore - Maintenance/tooling

## Risk Level

- [ ] Low - Minor changes, limited scope
- [x] Medium - Moderate changes, some user impact
- [ ] High - Major changes, significant user/system impact

Medium risk because this changes the VS Code debug/runtime startup path, Azurite extension activation handling, telemetry/error propagation, and the E2E runner coverage for the regression.

## What & Why

This change ensures the Azurite extension is activated and Azurite is reachable before Logic Apps debug/runtime startup continues. If automatic Azurite startup fails or times out, startup now stops with the Azurite readiness error instead of continuing into `AzureWebJobsStorage` validation and showing the misleading storage-emulator prompt. The fix preserves the original startup error through telemetry wrappers, distinguishes missing/inactive Azurite extension states, and adds unit plus VS Code E2E coverage for the regression.

The screenshot below shows the confusing failure users could hit when Azurite was still starting or unavailable but debug continued into storage validation:
image

## Impact of Change

- **Users**: Debug startup now fails fast when Azurite cannot be started automatically; users see the Azurite timeout/startup error instead of the confusing `AzureWebJobsStorage` warning or `Debug anyway` path.
- **Developers**: Reviewers should focus on `executeOnAzuriteExt`, `activateAzurite`, `pickFuncProcess`, `startRuntimeApi`, and `validatePreDebug` error propagation. The VS Code E2E runner now has a Phase 4.9 `azuriteonly` path with a fake Azurite extension for failure coverage.
- **System**: Automatic startup failure waits up to the readiness timeout before surfacing the Azurite error. The E2E harness binds Azurite ports and uses `AZURITE_E2E_STEP` / `AZURITE_E2E_WORKSPACE_PARENT` to split workspace creation from the debug assertion.

## Test Plan

- [x] Unit tests added/updated
- [x] E2E tests added/updated
- [x] Manual testing completed
- [x] Tested in: VS Code extension unit tests, extension build, E2E compile, `E2E_MODE=azuriteonly`, `E2E_MODE=nonlogicappstartup`, and `E2E_MODE=conversioncreateonly`

[...]

PR #9462, opened 2026-07-28
fix(vscode): unblock codeful F5 debug (design-time DLL lock, Azurite race, func self-heal) + add cross-platform codeful E2E coverage

lambrianmsft:
> **:warning: Split in progress — do not merge as-is.** The codeful design-time DLL-lock fix and the cross-platform codeful E2E coverage have been moved into a focused, self-contained PR: **#9463** (8 files, +1000/-54). This PR still contains that work **plus** the Azurite startup-race fix and the func Core Tools self-heal, so it now fully overlaps #9463.
>
> Plan: land #9463 first, then rebase this PR to carry **only** the Azurite readiness fix (superseding the closed #9147) and the func Core Tools self-heal. Its concrete deliverable is promoting both codeful E2E legs from advisory to **blocking** — they are advisory in #9463 precisely because the Azurite race is not fixed there.

> [!NOTE]
> **Supersedes #9147** (now closed). 13 of the 22 files here are the Azurite readiness fix from `lambrian/azurite_race_condition`, cherry-picked intact and re-validated on both OSes. That branch reported `CONFLICTING` against `main`; this one merges cleanly. See "Relationship to #9147" at the bottom for the exact file split.

## Commit Type

- [x] fix - Bug fix

## Risk Level

**Why Medium:** the codeful design-time change is scoped to codeful projects only, but the Azurite and pre-debug validation changes sit on the debug path for **all** Logic Apps projects, codeless included. Reviewers should focus there.

## What & Why

Reported against 5.981.0: open a codeful project, let it fully load, press F5 — debug fails. Killing the two lingering `func.exe` processes makes it work. The shipped 5.961.19 build is unaffected. Codeless is unaffected.

Investigating that turned up **three independent defects on the codeful F5 path**, plus the reason it wasn't caught: we had no codeful E2E coverage at all.

### 1. The design-time host held file locks on the codeful output DLLs (the reported bug)

The design-time host runs `func host start` with `ProjectDirectoryPath` pointed at the project root. When it runs the **in-process .NET 8 worker** it loads the project's compiled workflow assemblies out of `lib/codeful`, so its `dotnet`/`func` worker holds locks on them. F5 then runs the csproj `CopyToCodefulFolder` target (`AfterTargets="Build;Publish"`), which copies `bin/Debug/net8/**` back over `lib/codeful` and fails:

```
MSB3026: Could not copy "bin/Debug/net8/*.dll" to "lib/codeful/*.dll"
         ... the file is locked by ".NET Host (<pid>)"
```

That's exactly why killing `func.exe` unblocked it. It's a **latent ordering bug** — `pickFuncProcessInternal` only waits on the VS Code *task* map (`waitForPrevFuncTaskToStop`) and never stops the design-time host, so the design-time worker survives the build. `pickFuncProcess.ts` is byte-identical between 5.961.19 and 5.981.0, so the bug was always there; what changed was moving the design-time worker from Node to `dotnet` + `FUNCTIONS_INPROC_NET8_ENABLED` (needed for the Data Mapper "Test map" NetFxWorker). Under Node the design-time host never loaded those assemblies and so could never lock them.

**Fix:** codeful projects now always use the **Node** design-time worker, via a new `shouldUseNodeDesignTimeWorker(projectPath, logicAppType?)`. Applied in both `regenerateDesignTimeDirectory` (generates Node settings) and `validateDesignTimeDirectory` — the latter means an existing codeful `workflow-designtime/local.settings.json` still on `dotnet` is treated as invalid and regenerated, so **projects already broken by 5.98x self-heal on the next design-time start**. Non-codeful projects are untouched and still honor `azureLogicAppsStandard.useNodeDesignTimeWorker`.

Codeful doesn't need the in-process NetFxWorker — that exists for the Data Mapper XSLT transform, which isn't a codeful design-time feature. This also avoids killing/restarting the design-time host on every debug.

### 2. Azurite startup race → "Failed to verify AzureWebJobsStorage connection"

`activateAzurite` issued the start command and returned immediately, so `validateEmulatorIsRunning` could run before Azurite was listening. Worse, the failure was swallowed by the telemetry wrapper, so debug continued into a confusing downstream error.

- New `waitForAzuriteReady()` polls `validateEmulatorIsRunning` (10 × 500 ms) and throws a localized `azuriteFailedToStart` on timeout.
- `executeOnAzurite` now **activates** a deactivated Azurite extension instead of merely warning, and throws when it's missing or fails to activate. Adds `azuriteExtensionAvailable` / `azuriteExtensionActive` / `azuriteStartCommandIssued` telemetry.
- `pickFuncProcess` / `startRuntimeApi` set `rethrow = true` / `suppressDisplay = true` so Azurite failures propagate instead of being silently absorbed.
- A missing `AzureWebJobsStorage` now shows a specific modal (`showMissingAzureWebJobsStorageWarning`) instead of the generic emulator warning, and "Debug anyway" is suppressed when `autoStartAzurite` is on — we shouldn't offer to continue past a failure we own.
- Also fixes `autoStartAzurite` not being re-read after the user enables it mid-flow.

### 3. `func` present but not runnable dead-ended F5

Nothing on the F5 path distinguished "`func.exe` exists on disk" from "`func.exe` actually runs". A partially-hydrated install meant `preDebugValidate` → `isFuncToolsInstalled()` threw, producing a blocking "You must have the Azure Functions Core Tools installed" modal — unanswerable headlessly, and a dead end for users mid-provision.

**Fix:** `attemptManagedFuncCoreToolsRepair()` silently reinstalls the managed binaries and re-verifies before falling back to the interactive modal.

### 4. The reason none of this was caught: no codeful E2E

This is the part worth the most scrutiny. New coverage:

- **Phase 4.10A** — creates modern + legacy codeful workspaces through the Create Workspace webview.
- **Phase 4.10B** — drives a real F5 and asserts the actual task chain, with two cases that assert *opposite* correct behavior: the modern template (`AfterTargets="Build;Publish"`) must **skip** publish (`publishStart:0`), the legacy template (`AfterTargets="Publish"` only) must **still run** it (`publishStart:1, publishExit:0`). Both assert `buildExit === 0` — i.e. no MSB3026.
- Asserts the design-time worker is `node` with no `FUNCTIONS_INPROC_NET8_ENABLED` — a deterministic, cross-platform signal for defect #1 (the file-lock symptom itself only reproduces on Windows).
- Azurite autostart-failure E2E for defect #2.
- E2E dependency gate is now **executable-aware**: `funcVersionRuns()` requires `func --version` to exit 0 before a test proceeds, rather than checking for the file's existence.
- Unit coverage for `connections.json` being empty/missing/whitespace on the connection-view path (see "Known gaps").

**All of it runs on Windows as well as Ubuntu.** Windows previously had no runtime-dependency cache at all — `actions/cache@v5` is `post-if: success()`, and the only Windows job that downloaded dependencies always failed, so it could never save one. Self-perpetuating deadlock. Added a `setup-runtime-deps (windows)` bootstrap job plus `restore`/`save` pairs so a red job still persists what it downloaded, gated on a check that the cached tree is actually *runnable* (a half-hydrated cache is worse than none — it restores as a hit forever and never self-repairs).

## Impact of Change

- **Users**: codeful F5 works without manually killing `func.exe`; codeful projects already broken by 5.98x self-heal on next design-time start; Azurite/storage failures are actionable instead of a generic warning; a partially-installed `func` repairs itself instead of dead-ending on a modal.
- **Developers**: new exported `shouldUseNodeDesignTimeWorker(projectPath, logicAppType?)`. `validateEmulatorIsRunning`'s third parameter widens from `boolean` to `boolean | ValidateEmulatorOptions` — **backward compatible**, existing boolean call sites are unchanged. New `waitForAzuriteReady()` and `attemptManagedFuncCoreToolsRepair()`.
- **System**: the codeful design-time host process changes from `dotnet` + in-proc net8 to Node (**codeful only**; codeless and the Data Mapper NetFxWorker path are unchanged). Azurite failures now propagate rather than being swallowed. CI adds one advisory Windows bootstrap job and a working `la-runtime-deps-Windows-v1` cache; no change to `path`/`key` on the existing Linux cache entry.

## Test Plan

- [x] Unit tests added/updated — `validateProjectArtifacts` (codeful→Node, dotnet file treated invalid, non-codeful unchanged), `validatePreDebug`, `activateAzurite` (not-running-then-ready, throw-on-timeout), `executeOnAzuriteExt`, `validateFuncCoreToolsInstalled` (self-heal succeeds / repair fails → falls back to prompt), `pickFuncProcess.azurite`, `startRuntimeApi.azurite`, `connectionPanel` (empty/missing `connections.json`). Full `vs-code-designer` suite green (1551 passed / 128 files).
- [x] E2E tests added/updated — Phases 4.10A/4.10B (codeful create + debug task chain, modern & legacy) and Azurite autostart-failure, on **both** ubuntu-latest and windows-latest.
- [ ] Manual testing completed — the original repro is on @lambrianmsft's machine, not mine; defect #1 is covered by the automated E2E above. **Worth one manual codeful F5 pass before publishing.**
- [x] Tested in: GitHub Actions `ubuntu-latest` + `windows-latest` — **4 consecutive all-green runs** (`30298228591`, `30305533988`, `30307633190`, `30315114458`), all 17 jobs. `vscode-e2e-codeful (windows-latest)` is now a **blocking** check; it had never passed before this branch.

## Contributors

@lambrianmsft — reported the regression and the empty `connections.json`, authored the Azurite readiness fix (#9147) that is cherry-picked here, and set the direction on codeful E2E parity + Windows CI coverage.

## Screenshots/Videos

N/A — no UI changes. CI evidence is linked in the Test Plan.

---

## Known gaps (deliberately not claimed as fixed)

- **Empty `connections.json`** — not reproduced. `createConnectionsJson` writes `{}` when the file is absent, so `{}` on a fresh codeful project is expected, and `saveConnectionReferences` only writes when there's at least one key, so it doesn't clobber. No deterministic clobbering write path found. This PR only **hardens the consume path** with unit tests; the root cause is still open.
- **Connection-view E2E** — the connection view is not palette-invokable (`azureLogicAppsStandard.openLanguageServerConnectionView` is declared `"when": "never"`); it's dispatched by the language server against `.cs` files. An E2E has to drive the LSP CodeLens, not `executeCommand`. Not attempted here.
- **Windows parity beyond codeful** — the non-codeful scenario shards and the compat matrix are still ubuntu-only.

## Relationship to #9147

#9147 (`fix(vscode): stop debug startup on Azurite readiness timeout`) is **closed and superseded by this PR**. Its work ships here in full:

| | Files |
|---|---|
| **From #9147** (Azurite readiness) | `activateAzurite.ts`, `executeOnAzuriteExt.ts`, `validatePreDebug.ts`, `pickFuncProcess.ts`, `startRuntimeApi.ts`, `designerHelpers.ts`, `azuriteAutostartFailure*.test.ts` + their unit tests — 13 files |
| **New here** | `validateProjectArtifacts.ts` (the codeful fix), `validateFuncCoreToolsInstalled.ts`, `codefulDebugTasks.test.ts`, `run-e2e.ts`, `runtimeBinaryCheck.ts`, `connectionPanel.test.ts`, `.github/workflows/vscode-e2e.yml` — 9 files |

Cherry-picked `4834c527b` and `b6270be50` from `lambrian/azurite_race_condition`; both auto-merged cleanly and preserve #9429's `updateGlobalSetting` behavior. `740dc3bed` aligns the `activateAzurite` tests with `waitForAzuriteReady`. Skipped `f76f71346` (docs-only) and `f2c84e679` (E2E infra against the older `run-e2e.js`, incompatible with the TypeScript `run-e2e.ts` this branch uses — those Azurite E2E tests run here on the newer harness instead, across both OSes).

Consolidating was the right call on two counts: #9147 reported `CONFLICTING` against `main` while this branch merges cleanly, and the Azurite autostart-failure E2E only gets real cross-platform CI coverage once it's on the `run-e2e.ts` harness that ships here.


andrew-eldridge:
the copilot comments look valid

apps/vs-code-designer/src/app/azuriteExtension/executeOnAzuriteExt.ts
```diff
@@ -11,12 +11,35 @@ import * as vscode from 'vscode';
...
+    throw new Error(
+      localize(
+        'missingAzuriteExt',
+        'Azurite extension is not installed or is unavailable in the current VS Code extension host. Make sure the Azurite extension is installed and enabled, then try debugging again.'
+      )
+    );
   }
+
+  context.telemetry.properties.azuriteExtensionAvailable = 'true';
+  if (!azuriteExtension.isActive) {
```
andrew-eldridge, line 25: was the issue here that sometimes our extension activates before azurite, and this is the cause of the intermittent azurite errors?

apps/vs-code-designer/src/app/azuriteExtension/executeOnAzuriteExt.ts
```diff
@@ -11,12 +11,35 @@ import * as vscode from 'vscode';
...
+        )
+      );
+    }
+  }
+
+  context.telemetry.properties.azuriteExtensionActive = 'true';
+  context.telemetry.properties.azuriteStartCommandIssued = 'true';
+  await vscode.commands.executeCommand(command, {
+    ...args,
+  });
```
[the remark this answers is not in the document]
andrew-eldridge: +1

apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -40,7 +40,16 @@ export async function validateFuncCoreToolsInstalled(context: IActionContext, me
     } else if (await isFuncToolsInstalled()) {
       installed = true;
     } else if (await useBinariesDependencies()) {
-      installed = await validateFuncCoreToolsInstalledBinaries(innerContext, message, install, input, installed);
+      // The managed func binaries may exist on disk but fail to execute (partial extract,
+      // poisoned runtime-deps cache, or a reinstall that hasn't finished). Before dead-ending
+      // on the interactive "Install" modal — which cannot be answered headlessly and blocks
+      // debug — attempt a silent reinstall of the managed binaries and re-verify. This lets a
+      // provisioned-but-unrunnable func self-heal instead of aborting F5.
+      if (await attemptManagedFuncCoreToolsRepair(innerContext)) {
```
andrew-eldridge, line 48: nit: this function is called validateFuncCoreToolsInstalled but it is also installing if missing/broken. 'ensureFuncCoreTools'?

apps/vs-code-designer/src/app/commands/pickFuncProcess.ts
```diff
@@ -79,8 +79,14 @@ export async function pickFuncProcessInternal(
   projectPath: string
 ): Promise<string | undefined> {
   await callWithTelemetryAndErrorHandling(autoStartAzuriteSetting, async (actionContext: IActionContext) => {
+    actionContext.errorHandling.rethrow = true;
     await runWithDurationTelemetry(actionContext, autoStartAzuriteSetting, async () => {
-      await activateAzurite(context, projectPath);
+      try {
+        await activateAzurite(actionContext, projectPath);
```
andrew-eldridge, line 85: is the azurite error being suppressed now? does this mean it'll silently fail?

apps/vs-code-designer/src/app/utils/codeless/validateProjectArtifacts.ts
```diff
@@ -158,6 +158,33 @@ export async function detectLogicAppProjectType(projectPath: string): Promise<Pr
...
+ * so it never locks them. Codeful does not need the in-process .NET 8 NetFxWorker (that exists only
+ * for the Data Mapper Test map XSLT transform, which is not a codeful design-time feature).
+ *
+ * For non-codeful projects the behavior is unchanged: honor the
+ * `azureLogicAppsStandard.useNodeDesignTimeWorker` workspace setting (default dotnet + net8).
+ * @param {string} projectPath - The logic app project root.
+ * @param {ProjectType} [logicAppType] - Pre-detected project type; detected on demand when omitted.
+ * @returns {Promise<boolean>} True when the design-time host should use the Node worker.
+ */
+export async function shouldUseNodeDesignTimeWorker(projectPath: string, logicAppType?: ProjectType): Promise<boolean> {
```
andrew-eldridge, line 180: this is the main part we need for current release to fix codeful, can just this be split into another pr? The other changes here seem to have some risk and are worth waiting for vendor test pass

apps/vs-code-designer/src/app/azuriteExtension/executeOnAzuriteExt.ts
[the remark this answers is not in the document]
lambrianmsft: Good catch — fixed in 98aa0679d.

You're right about the mechanism: `{ ...['a', 'b'] }` yields `{ 0: 'a', 1: 'b' }`, which VS Code would receive as a *single* argument rather than two. Now:

```ts
if (args.length > 0) {
  await vscode.commands.executeCommand(command, ...args);
} else {
  await vscode.commands.executeCommand(command, {});
}
```

The `else` branch preserves the empty-options payload exactly as you suggested — that's pinned by an existing test, and I didn't want to change what the Azurite start command actually receives.

Two caveats worth stating plainly so nobody reads more into this than is there:

**It isn't reachable today.** `executeOnAzurite` has exactly one production call site — `activateAzurite.ts:100`, `executeOnAzurite(context, extensionCommand.azureAzuriteStart)` — with zero extra args. So `args` is always `[]` and `{ ...[] }` is `{}`. Every other reference is a test or a `vi.mock`.

**It's pre-existing, not introduced here.** `{ ...args }` appears on the removed (`-`) lines of this hunk too; this PR carried the prior behavior forward unchanged.

So: a latent-correctness fix, not a live bug. Proved it's a real guard rather than a tautological test by running the new case against the old implementation first:

```
× forwards caller arguments verbatim instead of collapsing them into an object
  → expected "spy" to be called with arguments: [ 'azurite.start', 'first', …(1) ]
```

and after the fix: `Test Files 1 passed (1)` (4 tests).

.github/workflows/vscode-e2e.yml
```diff
@@ -519,6 +519,603 @@ jobs:
...
+      - name: Cache Logic Apps runtime dependencies
+        uses: actions/cache@v5
+        with:
+          # %USERPROFILE%\.azurelogicapps on Windows; ~ resolves correctly under
+          # the job's bash shell.
+          path: ~/.azurelogicapps/dependencies
+          key: la-runtime-deps-${{ runner.os }}-v1
+          restore-keys: |
+            la-runtime-deps-${{ runner.os }}-
+
```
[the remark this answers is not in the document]
lambrianmsft: This is the better of the two catches — you found a real inconsistency in my own work. Fixed in 98aa0679d by converting this step to `actions/cache/restore@v5` (restore-only).

I added a poisoned-cache guard (`deps-check` + a conditional `cache/save`) to the three codeful jobs precisely because a half-hydrated entry is worse than none — then left this smoke shard, which I also added in this PR, on the naive `actions/cache@v5`. It runs `p40-nonlogicapp` with `LA_E2E_BUNDLE_PREFLIGHT=0` and hydrates nothing, so it's the one site that could only ever save junk.

**The mechanism, confirmed from `actions/cache` source rather than assumed:**

`saveImpl.ts` skips the post-save only on an *exact* key match:

```ts
if (utils.isExactKeyMatch(primaryKey, restoredKey)) {
  core.info(`Cache hit occurred on the primary key ${primaryKey}, not saving cache.`);
  return;
}
```

A `restore-keys` **prefix** hit doesn't take that branch — it restores older content and then re-saves it under the primary key. So `restore-keys: la-runtime-deps-${{ runner.os }}-` is what makes this self-inflicted: any stale sibling entry would be *promoted* onto `-v1`. That's not hypothetical — I had to manually delete an orphaned `la-runtime-deps-Linux-v2` (935 MiB) left over from an earlier key bump.

And it's unrecoverable once it fires: the codeful jobs restore the junk as an exact hit, so their `cache-hit != 'true'` save guard suppresses the repair, and the entry refreshes on access so it never LRU-evicts. Permanent until someone deletes it by hand.

**Where I'd refine your wording slightly:** it can't currently save an *empty* tree. `saveCacheV1` throws on a non-existent path and `saveImpl` catches it as a warning, so nothing is written. Verified against a real cold-cache run (`30291459936`), where this job hit all the other preconditions — cache miss, job green:

```
Cache not found for input keys: la-runtime-deps-Windows-v1, la-runtime-deps-Windows-
...
[warning]Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

So it's **latent, not live** — the only thing standing between us and a poisoned cache was `~/.azurelogicapps/dependencies` happening not to exist. That's far too thin a guarantee to rely on, especially since the same log shows the extension already resolving `autoRuntimeDependenciesPath: C:\Users\runneradmin\.azurelogicapps\dependencies`. One eager `mkdir` on activation and it goes live. Your call stands.

**Why restore-only rather than adding a `needs:` edge:** the ubuntu matrix has the identical naive pattern at line 468, but it's immune by *ordering* — `needs: setup-fixtures` guarantees the hydrating bootstrap saves first, so the matrix always sees an exact hit and skips its save. This Windows job deliberately runs in parallel with `setup-runtime-deps-windows` for speed, so serializing it behind a ~15-minute hydration to get the same guarantee would be a poor trade. Never saving achieves it for free. I left line 468 alone for that reason — happy to harden it too if you'd rather have belt and braces.

`path` / `key` / `restore-keys` are byte-identical, since `actions/cache` derives the cache entry *version* from the path set — editing it would silently invalidate the existing 1006 MiB Windows and 936 MiB Linux entries and cold-start the blocking ubuntu leg. Re-verified there's still exactly one distinct `path||key` pair across all six sites, and I'll confirm `vscode-e2e-codeful (windows-latest)` still reports `cache-hit: true` on this run.

lambrianmsft:
Closing as **fully superseded**. Every piece of this PR has now either merged or been re-opened as a focused PR:

| Work | Status |
|---|---|
| Codeful design-time DLL lock (`shouldUseNodeDesignTimeWorker` / Node design-time worker for codeful) | ✅ Merged in #9463 (now `projectConsistency/projectFilesConsistency.ts` on `main`) |
| Cross-platform codeful E2E coverage (Phase 4.10A/B, `codefulDebugTasksModern`/`Legacy`, Windows shards) | ✅ Merged in #9463 |
| Executable-aware E2E dependency gate (`funcVersionRuns`, `isFuncRunnable`, `assertFuncCoreToolsExecutable`) | ✅ Merged in #9463 |
| Azurite startup race (`waitForAzuriteReady`, `ValidateEmulatorOptions`, `executeOnAzurite` activation, missing-`AzureWebJobsStorage` modal) | ✅ On `main` |
| Azurite autostart-failure E2E (`azuriteAutostartFailure*.test.ts`) | ✅ On `main` |
| **func Core Tools self-heal** (`attemptManagedFuncCoreToolsRepair`) | ➡️ **#9478** |

The func self-heal was the only unmerged delta left, so it now ships on its own, rebased on current `main` (2 files, +84/−1). The runtime-deps cache-key `v1`→`v2` bump explored here was already reverted on this branch (`37bd6fa1f`) and is intentionally not carried forward, so `.github/workflows/vscode-e2e.yml` is untouched in #9478.

PR #9478, opened 2026-07-30
fix(vscode): self-heal a provisioned-but-unrunnable func Core Tools before blocking F5

lambrianmsft:
> Extracted from #9462. That PR is now fully superseded: the codeful design-time DLL-lock fix and the cross-platform codeful E2E coverage merged as **#9463**, and the Azurite startup-race fix is already on `main` (`waitForAzuriteReady`, `ValidateEmulatorOptions`, `warnMissingAzureWebJobsStorage`). The func Core Tools self-heal was the only piece left, so it ships here on its own, rebased on current `main`. #9462 is being closed as superseded.

## Commit Type

- [x] fix - Bug fix

## Risk Level

**Why Medium:** the code sits on `preDebugValidate`, which runs for **every** Logic Apps project (codeful and codeless), and it can trigger a managed-binaries reinstall on that path. It is tightly gated — see below — but reviewers should focus on the gating conditions.

## What & Why

Nothing on the F5 path distinguished "`func.exe` exists on disk" from "`func.exe` actually runs".

`preDebugValidate` -> `validateFuncCoreToolsInstalled` -> `isFuncToolsInstalled()` **spawns** `func --version`. When that fails — because the extension-managed binaries are half-extracted, mid-reinstall, restored from a poisoned runtime-deps cache, or hang on Windows — the old fallback was the interactive modal:

> You must have the Azure Functions Core Tools installed to debug your local functions.

That is a dead end for a user who is mid-provision (the binaries *are* being installed, they just aren't runnable yet), and it is unanswerable headlessly, so debug aborts or stalls.

### 1. Self-heal inside the managed-binaries validation path

`validateFuncCoreToolsInstalled` now keeps one branch per install path:

1. `azureLogicAppsStandard.validateFuncCoreTools` disabled -> installed,
2. existing runnable `func` -> installed,
3. extension-managed binaries -> `validateFuncCoreToolsInstalledBinaries(...)`,
4. system/`PATH` install -> `validateFuncCoreToolsInstalledSystem(...)`.

The managed-binaries validation branch first attempts a silent repair. It either waits for an existing in-flight install or reinstalls the managed binaries via `installFuncCoreToolsBinaries`, re-verifies `func --version`, and only falls back to the interactive modal when the repair still cannot produce a runnable `func`.

The self-heal is reached only when **all** of these hold, i.e. only in a state where debug was already going to dead-end:

1. `azureLogicAppsStandard.validateFuncCoreTools` is enabled, **and**
2. `isFuncToolsInstalled()` already returned false, **and**
3. `useBinariesDependencies()` is true — the extension owns these binaries.

System/`PATH` installs are deliberately untouched: we don't own those binaries and should not silently reinstall over them. That path still goes straight to the existing prompt.

### 2. Repair is silent through the existing telemetry/error-handling pattern

Andrew's review feedback pointed out that the first version added too much branch and option plumbing. The repair now runs in its own nested `callWithTelemetryAndErrorHandling('azureLogicAppsStandard.repairFuncCoreTools', ...)` scope with `context.errorHandling.rethrow = true` and `context.errorHandling.suppressDisplay = true`.

That removed the bespoke `{ suppressUi: true }` path from `installFuncCoreTools.ts` and `binaries.ts`. The installer is now install-only: it logs to the output channel and preserves in-flight coalescing/serialization, but it no longer reveals the Output panel by itself. Output reveal and manual-install follow-up prompts are owned by explicit user-selected install branches.

`downloadAndExtractDependency` also no longer owns download/checksum/extract toasts. It logs, records telemetry, cleans up partials, and throws; callers decide whether the surrounding action should surface the error. Activation/background dependency failures therefore remain log/telemetry-only under their existing `suppressDisplay=true` scopes, while user-initiated func installs still get the manual-install prompt if the install fails.

### 3. The repair must not race a concurrent install

`installFuncCoreToolsBinaries` had no in-flight guard, and `extractDependency` deletes and recreates the shared target directory. A failing `func --version` is *exactly* what you observe while another install is mid-extract, so the repair could start a second install on top of the first and re-corrupt the tree.

`installFuncCoreTools.ts` now tracks the active install (`isFuncCoreToolsInstallInFlight()` / `waitForFuncCoreToolsInstall()`, mirroring the existing `inFlightBundleWork` precedent in `bundleFeed.ts`), coalesces an identical concurrent request, and serializes otherwise. The repair awaits an install that is already running instead of starting another.

### 4. A hanging `func` must not stall F5 either (found by the Windows E2E leg)

`tryExecuteCommand` only settles on `'close'`/`'error'`. A managed `func` that **hangs instead of exiting** therefore stalled `validateFuncCoreToolsInstalled` indefinitely: the self-heal never ran, no modal appeared, nothing was logged, and F5 silently did nothing — strictly worse than the failure mode above. The spawned process also kept a handle on the binary, so the next attempt to replace it failed with `EBUSY`.

This is not hypothetical: it is what the new Windows E2E leg hit on its first run. The gate was reached, the deliberately corrupted `func` was never repaired, and the retry could not overwrite `func.exe` because it was still locked — which only something spawned from it could do, since the debug task never started.

New `executeCommandWithTimeout` in `cpUtils` terminates a command that outlives its budget. Commands are spawned with `shell: true`, so the tracked pid is the shell; on Windows the whole tree is taken down with `taskkill` (matching `startDesignTimeApi` / `funcHostTask`), because killing only the shell leaves the real command running and still holding its file handles. The `func --version` probe is bounded at 60s — generous enough that a slow first run (on-access AV scanning a freshly extracted binary) is not mistaken for a broken one — and a hang is treated exactly like a failure so the repair runs.

The probe failure reason is now logged. It was the only signal distinguishing "func is broken" from "the gate never ran" when debug appears to do nothing, and its absence is what made the first Windows failure undiagnosable from CI artifacts.

Telemetry records `funcRepairAttempted`, `funcRepairSucceeded`, `funcRepairAwaitedExistingInstall`, and `funcRepairError` on the nested repair telemetry scope.

## Impact of Change

- **Users**: a partially-installed, not-yet-runnable, or hanging extension-managed `func` repairs itself instead of dead-ending F5 on a modal or stalling silently. When the repair succeeds, debug proceeds with no prompt at all. When it fails, behavior falls back to the same install/learn-more modal; user-selected install failures still show the manual-install follow-up.
- **Developers**: no public API changes. The repair is module-private and now lives in the managed-binaries validation flow. `installFuncCoreToolsBinaries` and `downloadAndExtractDependency` no longer expose bespoke UI-suppression options; callers own UI surfacing through telemetry/error-handling scopes and branch-local prompts.
- **System**: one additional `installFuncCoreToolsBinaries` call, only on the managed-binaries path and only when `func --version` has already failed. No new dependencies. `.github/workflows/vscode-e2e.yml` gains the two self-heal shards described below.

## Test Plan

- [x] Unit tests added/updated — `validateFuncCoreToolsInstalled` (repair succeeds -> returns `true` without prompting; repair runs but `func` still won't run -> falls back to the prompt; the reinstall itself throws -> telemetry records it and falls back; an in-flight install is awaited rather than duplicated; a **timed-out** probe still triggers the repair; the `validateFuncCoreTools`-disabled opt-out; nested repair telemetry sets `rethrow=true` and `suppressDisplay=true`; silent repair does not reveal the Output panel; user-selected install does reveal it and still prompts for manual install on failure). `cpUtils` covers Windows `taskkill` tree-kill and non-Windows `SIGKILL`, clean exits, non-zero exits, and that `tryExecuteCommand` stays unbounded. `installRuntimeDependencies` and `binaries` cover install coalescing/serialization and the new caller-owned UI behavior.
- [x] E2E tests added/updated — new **Phase 4.14 `p414-funcrepair`** (`funcRepair.test.ts`): corrupts the managed func binary (containment-guarded, only ever writing under the resolved `LA_E2E_RUNTIME_DEPS_ROOT`), presses F5, and asserts the extension **silently repairs it instead of showing the install modal**, then restores. Runs on two CI shards — Linux (`func-selfheal`) and `vscode-e2e-funcselfheal (windows-latest)`. A CI manifest guard fails the job rather than letting the test `this.skip()` into a vacuous pass.
- [x] Tested in: `pnpm exec biome check --write <changed files>`; focused unit run (**4 files / 130 tests passed**); full `apps/vs-code-designer` unit suite (**127 files / 1563 tests passed**); E2E TypeScript compile (`npx tsup --config tsup.e2e.test.config.ts`). A repo-local `tsc --noEmit -p apps/vs-code-designer/tsconfig.json` run is not a clean validation signal in this worktree because it reports pre-existing unrelated extension typing errors outside the touched files.

### E2E evidence

| | Linux (`func-selfheal`) | Windows (`vscode-e2e-funcselfheal`) |
|---|---|---|
| Result | `repaired=true, pastConnectionKeys=true` | `repaired=true, pastConnectionKeys=true`, attempt 1/2 |
| Time from gate to repaired | 29s | 76s |

`pastConnectionKeys=true` is what proves the test is not passing vacuously: it confirms execution actually reached `preDebugValidate` (the gate under test) rather than succeeding somewhere upstream.

The Windows timing is itself the confirmation of part 4: 76s ~= the 60s bounded probe firing, plus the reinstall. Linux, where the corrupt binary fails fast rather than hanging, takes 29s. Before the bound existed, the same Windows scenario never repaired at all within a 300s budget.

## Contributors

@lambrianmsft — reported the codeful F5 regression that surfaced this defect and directed splitting it out of #9462 once the design-time, Azurite, and cross-platform E2E work had merged, asked for Windows coverage, and requested Andrew's review feedback be addressed by simplifying the flow.
@andrew-eldridge — review feedback that folded repair into managed-binaries validation and replaced bespoke UI-suppression options with the established telemetry/error-handling pattern.

## Screenshots/Videos

N/A — no UI changes. The only user-visible difference is the *absence* of a modal when the silent repair succeeds.


apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -79,6 +88,28 @@ async function isFuncToolsInstalled(): Promise<boolean> {
...
+ * re-verifies that `func --version` runs. Used to self-heal a func that exists on disk but
+ * fails to execute (partial extract / poisoned cache / interrupted install) so we don't
+ * dead-end on the interactive install modal that cannot be answered headlessly.
+ * @param {IActionContext} context - Command context.
+ * @returns {Promise<boolean>} True when the repair produced a runnable func, false otherwise.
+ */
+async function attemptManagedFuncCoreToolsRepair(context: IActionContext): Promise<boolean> {
+  context.telemetry.properties.funcRepairAttempted = 'true';
+  try {
+    await installFuncCoreToolsBinaries(context);
```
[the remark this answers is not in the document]
lambrianmsft: Good catch — confirmed and fixed in 57629f337.

The race is real and **in-process**, not just cross-window: activation runs `validateAndInstallBinaries` → `validateFuncCoreToolsIsLatest` → `installFuncCoreToolsBinaries(context, majorVersion)`. If you press F5 while that's still extracting, `func --version` fails *because of* that install, and the repair would start a second one on top of it. `downloadAndExtractDependency` stages into a shared `os.tmpdir()` folder and `extractDependency` deletes and recreates the target folder, so the two would clobber each other.

What changed:

- `installFuncCoreToolsBinaries` now tracks the in-flight install, following the `inFlightBundleWork` pattern already used in `bundleFeed.ts`. A concurrent request for the **same** major version joins the running promise; a request for a **different** major version waits for it to finish before starting. Two extracts can no longer overlap.
- New `isFuncCoreToolsInstallInFlight()` / `waitForFuncCoreToolsInstall()` helpers (the latter never rejects — the failure belongs to whoever started the install).
- `attemptManagedFuncCoreToolsRepair` checks the flag first: if an install is already running it awaits it and re-probes `func --version` instead of kicking off its own (recorded as `funcRepairAwaitedExistingInstall`). Worth noting the two call sites pass different `majorVersion` values — activation passes the feed-derived major, the repair passes `undefined` — so plain coalescing alone wouldn't have covered this pairing.

Tests: `installRuntimeDependencies.test.ts` covers joining, serializing, the in-flight flag, and that a failed install doesn't reject waiters; `validateFuncCoreToolsInstalled.test.ts` covers the repair awaiting an existing install rather than starting a second one.

**Deferred: cross-process (multi-window) locking.** This only serializes within a single extension host. I'm intentionally not adding a lockfile in this PR because:

1. It's pre-existing and broader than this change — every VS Code window already races into the same shared dependencies path at activation, with or without the self-heal.
2. It affects all four managed dependencies (node / dotnet / func / bundle), so it belongs in `downloadAndExtractDependency` / `ensureRuntimeDependenciesPath` rather than in the func-specific path.
3. A lockfile without a well-designed stale-lock recovery converts a transient crash into a permanent block, which is strictly worse than today's behavior.

Happy to file a follow-up issue for that if you'd like it tracked.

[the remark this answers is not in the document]
lambrianmsft: Confirmed — it wasn't actually silent. Two leaks, both fixed in 57629f337:

- `installFuncCoreTools.ts` called `ext.outputChannel.show()` unconditionally, so the automatic repair popped the Output panel open.
- `downloadAndExtractDependency` calls `vscode.window.showErrorMessage(...)` directly in three places (download failure, checksum mismatch, extract failure), which bypasses `innerContext.errorHandling.suppressDisplay`. A failed repair therefore showed an extra toast immediately before the existing "You must have the Azure Functions Core Tools installed" modal.

`installFuncCoreToolsBinaries` now takes an `{ suppressUi }` option that it forwards to `downloadAndExtractDependency`, and the repair path passes it. Logging and telemetry are unchanged — I also added an `appendLog` to the download-failure and checksum-mismatch branches, which previously only wrote telemetry, so a suppressed failure is still fully diagnosable in the output channel. User-initiated installs (the "Install" button and the activation-time check) keep their existing toasts.

Covered by new tests in `binaries.test.ts` (toast shown by default; suppressed with `suppressUi` while still logging and recording telemetry) and an assertion in `validateFuncCoreToolsInstalled.test.ts` that the repair passes `{ suppressUi: true }`.

lambrianmsft:
### Added E2E coverage for the self-heal

Up to now this PR was proven only by unit tests. There was no end-to-end coverage of the behavior at all — `funcVersionRuns()` in `runtimeBinaryCheck.ts` is a harness *precondition gate* that quietly waits for the self-heal, so it consumes the fix rather than proving it, and nothing deliberately broke `func` and asserted recovery.

`953fb1be4` adds scenario **`p414-funcrepair`** (Phase 4.14, `apps/vs-code-designer/src/test/ui/funcRepair.test.ts`).

**What it does**

1. Waits for the managed install and for two consecutive successful `func --version` probes. The second probe is the point — it proves activation-time validation has settled, so the repair observed later can't be attributed to it.
2. Backs up and overwrites every managed func executable **in place** with a short marker line. Each file still exists and keeps its `.exe` extension and execute bit, but no longer runs — the actual poisoned runtime-dependency cache state this PR exists to handle, not a mock.
3. Presses F5, which routes through `pickFuncProcessInternal` → `preDebugValidate` → `validateFuncCoreToolsInstalled` — the only caller of the gate.
4. Asserts recovery.

**Assertions**

| Kind | Assertion |
|---|---|
| Hard | The blocking `"You must have the Azure Functions Core Tools installed"` prompt never appears. The watch loop breaks early if it does, so a real regression fails in seconds rather than burning the timeout. |
| Hard (primary) | `func --version` runs again — disk-level, modeled on `waitForBundleRepaired`. |
| Hard | The repaired file no longer holds the marker bytes, catching a probe that resolved some *other* func. |
| Hard (liveness) | The workbench scrape returned text at least once, so the negative assertion above can't pass vacuously. |
| Soft | No `Error downloading the` / `Checksum verification failed` / `could not be installed at` toast — the user-visible half of `{ suppressUi: true }`. Non-fatal for the same reason `waitForRepairNotification` in `bundleRepair.test.ts` is: toast timing is racy and the disk signal is authoritative. |

**Two details worth calling out**

The corruption marker is deliberately non-empty and has no leading `#`. glibc's `execvp` falls back to `/bin/sh` on `ENOEXEC`, so an empty or comment-only file would *exit 0* on Linux and the whole scenario would pass while proving nothing. The test asserts `func --version` actually fails after the write, so a platform that ever disagrees fails loudly instead of silently going green.

Writes are guarded by a containment check against the managed `FuncCoreTools` directory, and `afterEach` restores the original bytes wherever the repair didn't already replace them — a mid-test failure can't poison the shared runtime-dependency cache for sibling scenarios. `run-e2e.ts` now publishes the resolved dependency root as `LA_E2E_RUNTIME_DEPS_ROOT` (the same mechanism `lspeperm` already uses) so the test targets what the harness actually wrote rather than guessing a path.

**What this does *not* cover:** the concurrent-install guard from [#discussion_r3680448234](https://github.com/Azure/LogicAppsUX/pull/9478#discussion_r3680448234). Two installs racing isn't deterministically reproducible through the UI — forcing the interleaving would mean driving the internals directly, at which point it's a unit test wearing an E2E costume. The 5 tests in `installRuntimeDependencies.test.ts` remain the proof there.

CI runs this on its own `func-selfheal` shard rather than folding it into `suite-and-bundle`: it's the only scenario that downloads the Func Core Tools package twice, so with `LA_E2E_SCENARIO_RETRIES=1` it risked pushing a retried shard past that job's 35-minute budget.

lambrianmsft:
### Windows E2E coverage — and the bug it found

Following up on the Linux-only coverage above: `p414-funcrepair` now also runs as its own blocking shard, `vscode-e2e-funcselfheal (windows-latest)`. That mattered more than expected — **it failed on its first run, and the failure was real, not harness flake.**

The gate was reached (`refreshConnectionKeys returned`), but the corrupted func was never repaired, no install modal appeared, and the extension logged nothing at all for the full 300s budget. The decisive clue was the *retry*: its corruption step failed `EBUSY` because `func.exe` was still locked. The debug task never started, so the only thing that could have created that lock was the `func --version` probe itself — i.e. the probe spawned a process that never exited.

Root cause: `tryExecuteCommand` only settles on `'close'`/`'error'`, so a `func` that **hangs instead of exiting** stalls `validateFuncCoreToolsInstalled` forever. The self-heal never gets a chance to run, and F5 silently does nothing — strictly worse than the failure mode this PR set out to fix, and a mode that skews Windows-heavy (AV quarantine / partial extraction leaving a non-PE `func.exe`).

Fixed in `5ca5fddc9` by bounding the probe (`executeCommandWithTimeout`, killing the process *tree* on Windows since `shell: true` means the tracked pid is only the shell) and treating a hang exactly like a failure. Both platforms are green now:

| | Linux (`func-selfheal`) | Windows (`vscode-e2e-funcselfheal`) |
|---|---|---|
| Result | `repaired=true, pastConnectionKeys=true` | `repaired=true, pastConnectionKeys=true`, attempt 1/2 |
| Gate → repaired | 29s | 76s |

The Windows timing is itself the confirmation: 76s ≈ the 60s bound firing plus the reinstall, versus 29s on Linux where the corrupt binary fails fast. `pastConnectionKeys=true` is what rules out a vacuous pass — it proves execution actually reached `preDebugValidate` rather than succeeding upstream.

Also added the log line that made this diagnosable at all. Its absence is why the first Windows failure was invisible in CI artifacts, and it is the same signal a user needs when debug "does nothing".

One scope note for reviewers: the concurrency guard from #discussion_r3680448234 is still proven by unit tests only — it is a same-process race that the E2E harness cannot deterministically induce.

apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -40,7 +46,16 @@ export async function validateFuncCoreToolsInstalled(context: IActionContext, me
     } else if (await isFuncToolsInstalled()) {
       installed = true;
     } else if (await useBinariesDependencies()) {
-      installed = await validateFuncCoreToolsInstalledBinaries(innerContext, message, install, input, installed);
+      // The managed func binaries may exist on disk but fail to execute (partial extract,
+      // poisoned runtime-deps cache, or a reinstall that hasn't finished). Before dead-ending
+      // on the interactive "Install" modal — which cannot be answered headlessly and blocks
+      // debug — attempt a silent reinstall of the managed binaries and re-verify. This lets a
+      // provisioned-but-unrunnable func self-heal instead of aborting F5.
+      if (await attemptManagedFuncCoreToolsRepair(innerContext)) {
```
andrew-eldridge: Could this just be part of the 'validateFuncCoreToolsInstalledBinaries' logic? If we can minimize the branching to keep the code simple would be optimal

apps/vs-code-designer/src/app/utils/binaries.ts
```diff
@@ -49,13 +49,24 @@ export { useBinariesDependencies } from './runtimeDependencies';
...
 export type { DownloadAttemptResult } from './integrity';
 
+export interface DownloadDependencyOptions {
+  /**
+   * Suppresses user-facing error toasts for installs the user did not explicitly request — for example
+   * the automatic Functions Core Tools repair attempted before a debug session, which already surfaces
+   * its own actionable prompt when it can't recover. Output-channel logging and telemetry are
+   * unaffected, so failures stay diagnosable.
+   */
+  suppressUi?: boolean;
```
andrew-eldridge: error suppression can be handled by wrapping sub-actions in a callWithTelemetryAndErrorHandling and setting the context.errorHandling.rethrow and context.errorHandling.suppressDisplay as needed to avoid using these options. This is the pattern I started using throughout the project in telemetry pr (e.g. here https://github.com/Azure/LogicAppsUX/pull/9480/changes#diff-ceae41388287784e922fb9c4e74a334c85c8e610a9c9d5352e5ddb40a70d83cbR25 )

andrew-eldridge:
One general note: I noticed in multiple places throughout vscode extension we have closely related/coupled functions, some of which could even be combined and simplified, spread across multiple different files which makes it harder to see the full picture. I think this applies to installFuncCoreTools.ts and validateFuncCoreToolsInstalled.ts here for example

apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsInstalled.ts
lambrianmsft: Addressed in `f7738b8`: the repair is now the first step inside `validateFuncCoreToolsInstalledBinaries(...)`, so the outer validator is back to one branch per path (managed binaries vs. system/PATH) instead of carrying a separate `attemptManagedFuncCoreToolsRepair` branch.

apps/vs-code-designer/src/app/utils/binaries.ts
lambrianmsft: Addressed in `f7738b8`: removed the bespoke `suppressUi` option plumbing from `installFuncCoreToolsBinaries` and `downloadAndExtractDependency`. The automatic repair now runs in a nested `callWithTelemetryAndErrorHandling('azureLogicAppsStandard.repairFuncCoreTools', ...)` scope with `rethrow = true` and `suppressDisplay = true`. Output-channel reveal is now owned only by the user-selected install branch, and the shared download helper logs/telemetry/throws without owning UI toasts.

lambrianmsft:
Addressed the simplification pass in `f7738b8`. I kept install mechanics in `installFuncCoreTools.ts` and validation policy in `validateFuncCoreToolsInstalled.ts`, but tightened the coupling points: repair now lives in the managed-binaries validation flow, install/download helpers no longer carry UI-suppression options, and user-facing prompts are branch-local where the selected action is known. That keeps the file ownership split while making the F5 path readable as one validation decision tree.

apps/vs-code-designer/src/app/commands/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -116,16 +186,31 @@ async function validateFuncCoreToolsInstalledSystem(
...
     await openUrl('https://aka.ms/Dqur4e');
   }
-  return installed;
+  return false;
+}
+
+async function showManualFuncCoreToolsInstallPrompt(context: IActionContext): Promise<void> {
+  if (
+    (await context.ui.showWarningMessage(
+      localize('failedInstallFuncTools', 'The Azure Functions Core Tools installion has failed and will have to be installed manually.'),
```
andrew-eldridge: nit: if this is a transient issue, 'will have to be installed manually' seem misleading. 'Please try again or manually install func core tools'?

apps/vs-code-designer/src/app/utils/funcCoreTools/cpUtils.ts
```diff
@@ -25,17 +25,40 @@ export async function executeCommandWithSanityLogging(
...
+ * exiting stalls its caller indefinitely and keeps its file handles open. Use this for probes that
+ * sit on an interactive path, where a hang is worse than a failure.
+ * @param {IAzExtOutputChannel | undefined} outputChannel - Output channel to stream the command output to.
+ * @param {string | undefined} workingDirectory - Directory to run the command from.
+ * @param {number} timeoutMs - Milliseconds to wait before killing the command.
+ * @param {string} command - Command to run.
+ * @param {string[]} args - Command arguments.
+ * @returns {Promise<string>} The command output, or a rejection when it fails or times out.
+ */
+export async function executeCommandWithTimeout(
```
andrew-eldridge: there is starting to be a lot of similar functions here, e.g. 'executeCommandWithTimeout', 'executeCommandInternal', 'tryExecuteCommandInternal'. I'm not familiar with all implementations, but can this be simplified to just use optional args for timeout for example?

andrew-eldridge (approved):

PR #9480, opened 2026-07-30
refactor(vscode): Refactor startup flow and 'nested' commands, improve telemetry for vscode extension

andrew-eldridge:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why
- Properly scope action context to fix issue where telemetry on single context is overwritten in sub-actions
- Add duration and result as minimal telemetry for each action
- Refactor some code to support implementation of the above two points
- Decouple the project files consistency check and startDesignTimeApi, now runs on startup and workspace folder change instead
- Add 'runProjectConsistencyCheck' command to run all checks (currently .vscode, project files) on each project in workspace
- Clean up some workspace utils

## Impact of Change

- **Users**: Project files consistency check (host.json and local.settings.json in root and workflow-designtime) now happen on startup and workspace folder change independent of whether auto start design-time enabled
- **Developers**: Improve granularity of telemetry for individual actions and ensure duration/result telemetry exists for each action, small refactors
- **System**: N/A

## Test Plan

- [ ] E2E tests added/updated
- [ ] Tested in: 

## Contributors
@andrew-eldridge


apps/vs-code-designer/src/app/utils/startRuntimeApi.ts
```diff
@@ -24,96 +24,66 @@ import { releaseReservedPort, reserveFreePort } from './portReservation';
...
-          actionContext.errorHandling.suppressDisplay = true;
-          throw error instanceof Error ? error : new Error(String(error));
-        }
-      });
-    });
-
-    if (azuriteSetupCancelled) {
-      throw new UserCancelledError(autoStartAzuriteSetting);
-    }
+export async function startRuntimeApi(context: IActionContext, projectPath: string): Promise<void> {
```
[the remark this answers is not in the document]
andrew-eldridge: this is fine for now, startRuntimeApi is unused

apps/vs-code-designer/src/app/commands/parameterizeConnections.ts
```diff
@@ -3,112 +3,111 @@
...
 import { getWorkspaceLogicAppFolders } from '../utils/workspace';
-import { DialogResponses, type IActionContext } from '@microsoft/vscode-azext-utils';
-import { window, workspace } from 'vscode';
+import { type IActionContext } from '@microsoft/vscode-azext-utils';
+import { workspace } from 'vscode';
 import type { ConnectionsData } from '@microsoft/vscode-extension-logic-apps';
 
 /**
- * Parameterizes the connections in each workspace project if needed.
+ * Parameterizes the connections in all Logic Apps projects within the workspace.
```
lambrianmsft, line 18: Users may have a mix of older projects that they don't want to parameterize. So I am not sure that this is a good idea.

apps/vs-code-designer/src/constants.ts
```diff
@@ -142,13 +142,15 @@ export const dependencyUpdateCheckIntervalMs = 24 * 60 * 60 * 1000; // 24 hours
 
 // Commands
 export const extensionCommand = {
+  ensureWorkspace: 'azureLogicAppsStandard.ensureWorkspace',
```
lambrianmsft: why do we need to add all these as extensionCommands? It doesn't look like all of these are used. I see some in registerCommands but not all.

apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
```diff
@@ -315,7 +315,6 @@ export async function startDesignTimeApi(projectPath: string): Promise<void> {
           designTimeInst.validationRetryCount = 0;
           actionContext.telemetry.properties.startDesignTimeApi = 'true';
           updateFuncIgnore(projectPath, [`${designTimeDirectoryName}/`]);
-          actionContext.telemetry.measurements.startDesignTimeApiDuration = (Date.now() - loadDesignTimeStart) / 1000;
```
lambrianmsft, line 318: Why was the design time startup duration removed?

lambrianmsft (approved):

apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
andrew-eldridge: since duration is logged automatically on context for each callWithTelemetryAndErrorHandling, I just updated to make sure each of the actions used a properly scoped actionContext, including with this one. Think I missed a few places where startDesignTimeApi is used, will go back and fix those

apps/vs-code-designer/src/app/commands/parameterizeConnections.ts
andrew-eldridge: this one is only used when the global setting 'parameterizeConnectionsInProjectLoad' is enabled, for individual projects parameterizeProjectConnections can still be used. I don't think there is a behavior change here- in case we do want to allow configuring this setting per project it'll require additional changes

apps/vs-code-designer/src/constants.ts
andrew-eldridge: yeah I think this is right- I was adding the new values I needed for callbackId in action scopes here, and they don't all correspond to an extension command. Will see if there's a better way to differentiate them

andrew-eldridge: removed all extensionCommand that don't correspond to an actual command, fixed telemetry callbackIds

apps/vs-code-designer/src/app/utils/codeless/startDesignTimeApi.ts
andrew-eldridge: updated startDesignTimeApi usages so they have telemetry scoped to just this function, so the built-in 'duration' will be equivalent to this now

Issue #9538, opened 2026-08-14
UX for httpStatusCodes misleading and feature itself does not work as expected.


asos-martinsmith:
### Severity

P3 - Medium (Minor functionality affected)

### Describe the Bug with repro steps

I was glad to see the Http Status codes are now selectable when creating a retry policy but have the following feedback on the feature.

**The UX is missing information and is misleading**

It is not stated in the UX whether the selected "HTTP status codes" are _instead of_ the standard set of 408, 429, and 5xx or are _in addition_ to it.
It is left up to the user to experiment to find out for themselves.
The list of HTTP Status codes that you can select *includes* 408, 429 and all of the 5xx ones so my guess would be that this is _instead of_. This guess would be wrong.

I suggest the info tooltip be expanded to state this explicitly and additionally  408, 429 and all of the 5xx ones removed from the list (as selecting them will have no effect). I would also propose adding 499 to the list (thrown on stateless workflow timeout) as this is a use case I often want retried

**When the custom status code is encountered the specified retry policy is ignored and it degrades to default exponential**

I would expect the action below to retry 7 times not 4

```
            "418_With_Custom_Status": {
                "type": "Http",
                "inputs": {
                    "uri": "https://httpbin.org/status/418",
                    "method": "GET",
                    "retryPolicy": {
                        "type": "exponential",
                        "count": 7,
                        "interval": "PT20S",
                        "minimumInterval": "PT5S",
                        "maximumInterval": "PT30S",
                        "httpStatusCodes": [
                            418
                        ]
                    }
                }
```

### What type of Logic App Is this happening in?

Standard (Portal)

### Which operating system are you using?

Windows

### Did you refer to the TSG before filing this issue? https://aka.ms/lauxtsg

Yes

### Workflow JSON

```json
{
    "definition": {
        "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
        "contentVersion": "1.0.0.0",
        "actions": {
            "418_No_Custom_Status": {
                "type": "Http",
                "inputs": {
                    "uri": "https://httpbin.org/status/418",
                    "method": "GET",
                    "retryPolicy": {
                        "type": "exponential",
                        "count": 7,
                        "interval": "PT20S",
                        "minimumInterval": "PT5S",
                        "maximumInterval": "PT30S"
                    }
                },
                "runAfter": {},
                "runtimeConfiguration": {
                    "contentTransfer": {
[...]

PR #9544, opened 2026-08-17
fix(designer): clarify retry policy HTTP status codes UX (#9538)

takyyon:
## Commit Type
- [x] fix - Bug fix

## Risk Level

## What & Why
Addresses the UX feedback in #9538 on the retry-policy **HTTP status codes** picker.

Two UX problems were reported:
1. The picker never states how a custom status-code list interacts with the default retryable set (408, 429, 5xx), leaving users to guess whether custom codes are used *instead of* or *in addition to* the defaults.
2. Codes the user commonly wants retried (e.g. 499 on stateless-workflow timeout) could not be selected — the control only accepts values from its predefined list.

This PR:
- **Rewords the tooltip** to state that selecting codes *replaces* the default retryable set (only the selected codes are retried) and that leaving it empty keeps the defaults.
- **Adds `499 - Client Closed Request`** to the selectable options.
- **Removes the dead `freeform` prop** from `SettingTagPicker`: the control never committed typed-but-unmatched values, so `freeform` was misleading. Selection is now explicitly from the curated list.
- **Keeps 408/429/5xx in the list on purpose** — under replace semantics users must be able to re-add them alongside custom codes. (The issue suggested removing them; that would break the re-add use case.)

> Note: the issue also reports that the runtime ignores `httpStatusCodes` (a 418 in the list retried the default 4x instead of the configured count). That behavior lives in the Logic Apps runtime engine, not this repo — the designer serializes the field correctly. It is tracked separately for the runtime/PM team; this PR only fixes the in-designer UX.

## Impact of Change
- **Users**: Clearer tooltip explaining custom-vs-default retry behavior; 499 is now selectable.
- **Developers**: `SettingTagPicker` no longer advertises freeform entry it did not support.
- **System**: No runtime/behavioral change to workflow serialization; text + option-list only.

## Test Plan
- [x] Unit tests added/updated (settingtagpicker snapshot updated for removed `freeform`)
- [ ] E2E tests added/updated
- [x] Tested in: Standalone designer (Standard), settings -> Networking -> Retry policy


asos-martinsmith:
Thanks for picking this up. However this implementation seems to assume that the `httpStatusCodes` do replace the standard set of status codes and therefore there is some value in including 408, 429 and all of the 5xx ones in the list?

The demo in the linked issue shows that they do not replace the standard set. They are in addition to it. Note that only 418 was specified in the array and 500 was retried as normal. It did not suppress the retry of 500 and there does not seem to be any need to explicitly include all the "normal" ones along with the additional ones.

This does prove my point that the documentation is lacking on the feature though.

PR #9560, opened 2026-08-21
fix(designer): warn before enabling trigger concurrency

Elaina-Lee:
## Commit Type

- [x] fix - Bug fix

## Risk Level

## What & Why

Enabling trigger concurrency control is **irreversible** for a trigger once turned on. Previously, the designer's General settings toggle enabled this option immediately on click, with no warning that the choice is permanent.

This change adds an explicit user consent step: clicking the toggle to enable trigger concurrency control (from off to on, trigger operations only) now opens a confirmation dialog that states the setting can't be turned off once it takes effect. The toggle only takes effect after the user explicitly confirms via the dialog's confirm button; canceling leaves the setting untouched. Enabling/disabling concurrency on non-trigger (action) operations, and turning concurrency control off, are unaffected and continue to apply immediately, since only the trigger-enable path is irreversible.

Implements the fix requested in serverless-paas-balam/logic-apps-central#610.

## Impact of Change

- **Users**: When enabling trigger concurrency control in the designer's General settings panel, users see a confirmation dialog explaining the setting can't be turned off once it takes effect, and must explicitly confirm before the toggle takes effect. Canceling leaves the setting untouched. No change for actions or for disabling concurrency.
- **Developers**: The shared `Confirm` dialog component (`libs/designer-ui/src/lib/dialogs/confirm.tsx`) has an optional `confirmText` prop to override the default "OK" primary button label. The v1 (`libs/designer/...`) and v2 (`libs/designer-v2/...`) General settings sections each wire a warning confirmation into the trigger-concurrency toggle when it is switched on.
- **System**: No performance, architecture, or dependency impact; purely additive UI confirmation logic and new localized strings.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Local unit test runs (targeted) - `confirm.spec.tsx` (shared `Confirm` dialog `confirmText` override) and updated `general.spec.tsx` for both v1 and v2 covering the warning-confirmation dialog before enabling trigger concurrency. All targeted tests passed.

## Screenshots/Videos

v1:
image
v2:
image


Elaina-Lee:
Superseded by the clean replacement PR #9576, which preserves the intended trigger-concurrency confirmation change while excluding unrelated snapshot, TimelineHeader, quickViewPanel, merge, and persisted-lock churn. Please continue review and CI on #9576. Leaving this PR open for history; it is not being closed by this handoff.

PR #9581, opened 2026-08-27
fix(designer): regenerate stale fluentui-icon snapshots missed by #9493

takyyon:
## Summary

`libs/designer` (v1) had **11 stale Vitest snapshots** that baked in real inline SVG markup from `@fluentui/react-icons`. They pass on stale local installs but fail in CI's fresh install, breaking the `build` and `coverage` jobs on every PR (and on `main`).

## Root cause

#9493 made the `@fluentui/react-icons` Vitest mock reliably intercept — it added the package as a `devDependency` of `shared-test-utils` so `vi.mock('@fluentui/react-icons')` resolves to the same physical module consumers import. It then regenerated the affected snapshots in `libs/designer-ui` and `libs/designer-v2`, and concluded:

> libs/designer (v1) also wires this setup file but has no snapshots containing real icon SVG output, so nothing to regenerate there.

That was incorrect. 11 v1 snapshots still contained real icon SVG output. With the mock now intercepting under CI's fresh pnpm layout, those icons render as `<span data-icon-name="…" />` / `<span data-testid="fluent-icon" />` stubs, so the committed snapshots mismatch.

**Why it hid locally:** on a stale `node_modules`, `libs/shared-test-utils/node_modules/@fluentui/react-icons` isn't linked, so the mock silently doesn't resolve and real icons render — matching the stale snapshots. The mismatch only reproduces after a clean `pnpm install`, which is exactly what CI does. That's why `build`/`coverage` are red on `main` and every open PR while local runs look green.

## Changes

- Regenerate the 11 stale `libs/designer` snapshots to the mock's documented stub output:
  - `MonitoringTimeline`: `TimelineButtons`, `TimelineContent`, `TimelineGroup`, `TimelineHeader`
  - `menuItems`: `collapseMenuItem`, `pinMenuItem`
  - `panel`: `agentChatHeader`, monitoring `inputsPanel`, `outputsPanel`
  - `settings`: `advancedSettingsMessage`, `settingsection`
- Port the host-element filter #9493 applied to **designer-v2**'s `TimelineHeader.test.tsx` to the **v1** copy it missed. Under the mock the icon stub spreads `className` onto its host `<span>`, so `findAllByProps({ className: 'timeline-icon' })` matched both the wrapper component instance and its host element, double-counting the icon. Filtering to host elements restores the intended count of 1.

## Out of scope

`libs/designer`'s `quickViewPanel.spec.tsx` has a pre-existing, unrelated failing assertion (the test reassigns its Redux store mid-test without re-rendering), documented as pre-existing in #9493. It is not caused by the mock and is left untouched.

## Testing

Clean `pnpm install`, then from `libs/designer`:
- All 11 previously-failing spec files pass (68 tests).
- `npx vitest run -u` produces no further snapshot changes.

## Commit Type

fix

## Risk Level

low — test-only change (regenerated snapshots + one test-assertion filter); no product code touched.

### DIFF
--- a/libs/designer/src/lib/ui/MonitoringTimeline/__tests__/TimelineHeader.test.tsx
+++ b/libs/designer/src/lib/ui/MonitoringTimeline/__tests__/TimelineHeader.test.tsx
@@ -83,9 +83,14 @@ describe('TimelineHeader', () => {
   });
 
   it('should render timeline icon in both expanded and collapsed states', () => {
+    // Only count rendered host elements (e.g. the icon's <span>/<svg>), not the wrapping
+    // component instance, which also carries the same className prop and would otherwise
+    // be double-counted by findAllByProps.
+    const isHostElement = (instance: { type: unknown }) => typeof instance.type === 'string';
+
     // Test expanded
     const expandedComponent = renderWithIntl(defaultProps);
-    const expandedIcons = expandedComponent.root.findAllByProps({ className: 'timeline-icon' });
+    const expandedIcons = expandedComponent.root.findAllByProps({ className: 'timeline-icon' }).filter(isHostElement);
     expect(expandedIcons).toHaveLength(1);
 
     // Test collapsed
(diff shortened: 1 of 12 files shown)


takyyon:
Closing in favor of #9580, which lands the same fix (identical 11 snapshot regenerations + the same TimelineHeader host-element filter ported from #9493's designer-v2 change, quickViewPanel likewise left out). Elaina's PR got there first and carries the fuller context from the #9544/#9560 chain. Opened this independently before spotting #9580 — no need for two.

PR #9598, opened 2026-09-03
feat(knowledge): persist Cosmos DB resource ID

bjbennet:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Knowledge hub connections currently persist the selected Cosmos DB endpoint but discard the Azure resource ID chosen in the connection wizard. This change serializes the existing Cosmos DB account selection as `cosmosDB.resourceId`, making the account identity available alongside the endpoint in `connections.json`.

## Impact of Change

- **Users**: Newly created and updated knowledge hub connections retain the selected Cosmos DB account resource ID.
- **Developers**: `KnowledgeHubConnectionModel.cosmosDB` now includes a required `resourceId` property.
- **System**: The Standard connection serializer writes the selected account ID under `knowledgeHubConnections.HubConnection.cosmosDB.resourceId` and restores it when editing.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: Focused Vitest suites for designer knowledge connection metadata and Standard connection persistence (43 tests)

## Contributors

@brbenn_microsoft

## Screenshots/Videos

N/A - no visual changes.

### DIFF
--- a/apps/vs-code-react/src/app/designer/servicesHelper.ts
+++ b/apps/vs-code-react/src/app/designer/servicesHelper.ts
@@ -434,7 +434,7 @@ const addConnectionInJson = (
   connectionAndSetting: ConnectionAndAppSetting<LocalConnectionModel>,
   connectionsJson: ConnectionsData
 ): ConnectionsData => {
-  const { connectionData, connectionKey, pathLocation } = connectionAndSetting;
+  const { connectionData, connectionKey, pathLocation, isUpdate } = connectionAndSetting;
   const pathToSetConnectionsData: any = clone(connectionsJson);
 
   for (const path of pathLocation) {
@@ -442,7 +442,7 @@ const addConnectionInJson = (
       pathToSetConnectionsData[path] = {};
     }
 
-    if (pathToSetConnectionsData && pathToSetConnectionsData[path][connectionKey]) {
+    if (pathToSetConnectionsData && pathToSetConnectionsData[path][connectionKey] && !isUpdate) {
       break;
     }
     pathToSetConnectionsData[path][connectionKey] = connectionData;
--- a/libs/designer/src/lib/core/knowledge/utils/connection.ts
+++ b/libs/designer/src/lib/core/knowledge/utils/connection.ts
@@ -39,7 +39,7 @@ const getAllConnectionParameters = (intl: IntlShape) => {
         constraints: {
           clearText: true,
           required: 'true',
-          serialize: false,
+          serializationPath: ['cosmosDB', 'resourceId'],
         },
       },
     } as ConnectionParameter,
--- a/libs/designer/src/lib/core/knowledge/utils/connection.ts
+++ b/libs/designer/src/lib/core/knowledge/utils/connection.ts
@@ -374,7 +374,7 @@ export const createOrUpdateConnection = async (parameterValues: Record<string, a
     const connection = await ConnectionService().createConnection(
       'HubConnection',
       { id: '/dummy/knowledgehub' } as unknown as Connector,
-      { displayName, connectionParameters: parameterValues },
+      { displayName, connectionParameters: parameterValues, isUpdate: !isCreate },
       { connectionParameters, connectionMetadata: { required: true, type: ConnectionType.KnowledgeHub } }
     );
 
--- a/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Models/Workflow.ts
+++ b/apps/Standalone/src/designer/app/AzureLogicAppsDesigner/Models/Workflow.ts
@@ -122,6 +122,7 @@ export interface ConnectionAndAppSetting {
   connectionData: ServiceProviderConnectionModel | FunctionConnectionModel;
   settings: Record<string, string>;
   pathLocation: string[];
+  isUpdate?: boolean;
 }
 
 export interface ConnectionsData {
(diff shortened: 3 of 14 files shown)


libs/designer/src/lib/core/knowledge/utils/connection.ts
```diff
@@ -39,7 +39,7 @@ const getAllConnectionParameters = (intl: IntlShape) => {
         constraints: {
           clearText: true,
           required: 'true',
-          serialize: false,
+          serializationPath: ['cosmosDB', 'resourceId'],
```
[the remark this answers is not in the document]
bjbennet: updated

apps/vs-code-react/src/app/designer/servicesHelper.ts
```diff
@@ -434,15 +434,15 @@ const addConnectionInJson = (
...
+  const { connectionData, connectionKey, pathLocation, isUpdate } = connectionAndSetting;
   const pathToSetConnectionsData: any = clone(connectionsJson);
 
   for (const path of pathLocation) {
     if (!pathToSetConnectionsData[path]) {
       pathToSetConnectionsData[path] = {};
     }
 
-    if (pathToSetConnectionsData && pathToSetConnectionsData[path][connectionKey]) {
+    if (pathToSetConnectionsData && pathToSetConnectionsData[path][connectionKey] && !isUpdate) {
```
preetriti1, line 445: I keep seeing isUpdate code being used in multiple places, then are we enabling the resource id change from UI as well when we do edit connection?

bjbennet: `isUpdate` allows V1 to overwrite an existing `HubConnection`, so an edit save can add `cosmosDB.resourceId` to a legacy connection that did not have it. The property is added only when the edit UI has a selected Cosmos DB account resource ID to serialize; `isUpdate` does not automatically discover or backfill the ID without that value.

preetriti1 (approved):

apps/vs-code-react/src/app/designer/servicesHelper.ts
preetriti1: But in UI only until you change the model this Save button will be enabled, so to fix the MI connections would user have to change their models?

preetriti1 (changes requested):

apps/vs-code-react/src/app/designer/servicesHelper.ts
bjbennet: Yes, this is correct. Backing filling connections would be an entirely different change.

bjbennet: pushed a fix to address this

libs/designer/src/lib/ui/knowledge/panel/connection/edit.tsx
```diff
@@ -41,6 +47,12 @@ export const EditConnectionPanel = ({ mountNode }: { mountNode: HTMLDivElement |
   const { data: completionModels = [] } = useCompletionModelsByEndpoint(openAIEndpoint, openAIKey);
   const { data: embeddingModels = [] } = useEmbeddingModelsByEndpoint(openAIEndpoint, openAIKey);
   const [isDirty, setIsDirty] = useState(false);
+  const { data: subscriptions, isLoading: areSubscriptionsLoading } = useSubscriptions();
+  const shouldResolveCosmosDbResourceId = !parameterValues?.cosmosDbServiceAccountId && !!parameterValues?.cosmosDBEndpoint;
```
preetriti1: You should put this in useMemo, so wehn parameterValues are empty it updates when it gets values otherwse shouldResolve will always resolve into first value which might be false.

libs/designer/src/lib/ui/knowledge/panel/connection/edit.tsx
```diff
@@ -53,6 +65,19 @@ export const EditConnectionPanel = ({ mountNode }: { mountNode: HTMLDivElement |
...
+    if (resolvedCosmosDbResourceId) {
+      setConnectionParameterValues((values) =>
+        values.cosmosDbServiceAccountId
+          ? values
+          : {
+              ...values,
+              cosmosDbServiceAccountId: resolvedCosmosDbResourceId,
+            }
+      );
+    }
```
preetriti1, line 81: Can we only do set when resourceId isnot present, just avoiding react update

preetriti1 (approved):

libs/designer/src/lib/ui/knowledge/panel/connection/edit.tsx
bjbennet: Updated `shouldResolveCosmosDbResourceId` to use `useMemo` with `parameterValues` as its dependency, so the predicate is recomputed when the asynchronously loaded edit values change.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

libs/designer/src/lib/ui/knowledge/panel/connection/edit.tsx
bjbennet: Updated the effect to call the setter only when the current state is missing `cosmosDbServiceAccountId`. I retained the functional-update guard as well so a delayed lookup cannot overwrite a newer resource ID.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

PR #9607, opened 2026-09-09
feat(vscode): add stateless codeful template

lambrianmsft:
## Commit Type

- [ ] fix - Bug fix

## Risk Level

## What & Why

Adds the missing Stateless workflow option and C# scaffold for codeful Logic Apps. Codeful workspace creation already supported Stateful, Autonomous agent, and Conversational agent workflows, while the shared workflow model defined a Stateless codeful type that was not exposed or generated. This change completes that end-to-end contract, discovers created Stateless workflows with the correct runtime kind, and makes them selectable in the existing Update debug mode command so `WithStatelessRunHistory` can be enabled and executed runs appear on Overview.

## Impact of Change

- **Users**: Can create Stateless codeful Logic App workflows and explicitly enable persisted run history for local Overview debugging.
- **Developers**: Adds explicit codeful workflow-type normalization, template selection, source/debug-mode discovery, and focused tests.
- **System**: Generated projects remain stateless by default and use the existing bundled `Microsoft.Azure.Workflows.Sdk`; run-history persistence stays opt-in through the existing debug-mode setting.

## Test Plan

- [ ] E2E tests added/updated
- [x] Tested in: 144 targeted Vitest tests across `vs-code-react` and `vs-code-designer`; generated Stateless C# template compiled against the bundled `Microsoft.Azure.Workflows.Sdk` with 0 warnings and 0 errors; VS Code extension bundle completed successfully. The repository type-check still reports pre-existing errors in unrelated files.

## Contributors

@Copilot

## Screenshots/Videos

N/A - this adds an option to the existing workflow type dropdown and review summary without changing styling or layout.


apps/vs-code-designer/src/app/commands/workflows/switchDebugMode/switchDebugModeSteps/__test__/StatelessWorkflowsListStep.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const normalizedPath = String(filePath);
+      if (normalizedPath.endsWith('workflow.json')) {
+        return JSON.stringify({ kind: 'Stateless' });
+      }
+      if (normalizedPath.endsWith('StatelessWorkflow.cs')) {
+        return 'WorkflowFactory.CreateStatelessWorkflow("codeful-stateless", workflow);';
+      }
+      if (normalizedPath.endsWith('MultipleWorkflows.cs')) {
+        return `
+          WorkflowFactory.CreateStatelessWorkflow("codeful-stateless-two", workflow);
```
andrew-eldridge, line 57: mocked workflow file contents are invalid (appears to just contain the lines we're looking for in code). One thing I noticed about our unit tests in general is they are sensitive to changes and will break from even small changes to implementation details. Should try to make more robust high-level tests going forward

apps/vs-code-designer/src/app/commands/workflows/overview/utils/codefulHelpers.ts
```diff
@@ -169,7 +168,8 @@ export async function getCodefulWorkflowDataList(
       ? [
           {
             workflowName: workflowInfo.workflowName,
-            workflowKind: workflowInfo.workflowType === 'agent' ? 'Agent' : 'Stateful',
+            workflowKind:
+              workflowInfo.workflowType === 'agent' ? 'Agent' : workflowInfo.workflowType === 'stateless' ? 'Stateless' : 'Stateful',
```
andrew-eldridge, line 172: should we just use the enum for workflowKind rather than trying to convert to string? It looks like this is missing agentic workflowType as well

apps/vs-code-designer/src/app/commands/workflows/overview/utils/codefulHelpers.ts
```diff
@@ -248,12 +248,20 @@ export function getCodefulWorkflowNames(filePath: string): string[] {
...
       let match: RegExpExecArray | null;
       while ((match = workflowRegex.exec(fileContent)) !== null) {
-        const workflowName = match[1];
-        if (workflowName && !workflowNames.includes(workflowName)) {
-          workflowNames.push(workflowName);
+        const factoryMethod = match[1];
+        const workflowName = match[2];
+        if (workflowName && !workflows.some((workflow) => workflow.workflowName === workflowName)) {
+          const workflowKind =
+            factoryMethod === 'CreateStatelessWorkflow'
```
andrew-eldridge, line 259: also seems to be missing agentic option

apps/vs-code-designer/src/app/commands/workflows/switchDebugMode/switchDebugModeSteps/__test__/StatelessWorkflowsListStep.test.ts
lambrianmsft: Updated the tests to exercise real temporary project directories and complete C# workflow-provider files instead of mocking filesystem calls with isolated detector fragments. This keeps the cases focused on observable discovery behavior while still covering codeless/codeful workflows, duplicates, multiple declarations, comments, non-literal names, malformed JSON, exclusions, and missing paths.

apps/vs-code-designer/src/app/commands/workflows/overview/utils/codefulHelpers.ts
lambrianmsft: Updated this to use the shared `WorkflowKind` constants and narrowed `CodefulWorkflowData.workflowKind` accordingly. The autonomous `agenticCodeful` option is intentionally represented as `WorkflowKind.stateful`: its template calls `CreateStatefulWorkflow`, and `WorkflowKind.agentic` is defined as `'Stateful'`. The separate conversational `agentCodeful` option calls `CreateAgentWorkflow` and maps to `WorkflowKind.agent` (`'Agent'`). Added regression coverage for both mappings.

apps/vs-code-designer/src/app/commands/workflows/overview/utils/codefulHelpers.ts
lambrianmsft: Confirmed the agent cases are covered, but under two different runtime kinds. Both conversational-agent APIs (`CreateConversationalAgent` and `CreateAgentWorkflow`) are discovered as `Agent`. The autonomous agentic template uses `CreateStatefulWorkflow`, so it is intentionally discovered as `Stateful` rather than requiring a separate source detector branch. Added explicit tests documenting this contract and converted the mapping to `WorkflowKind` constants.

